PR #7, opened 2015-09-17
Migration to CMS fronts stack in AWS

piuccio:
As a first step we've decide to move the application to a different AWS stack, owned by CMS fronts.

There's no change in functionality or architecture, the new app is still writing to the same S3 bucket and talking to the same SQS queue on frontend (consumed by `facia-press`).

The most important changes are

* `configuration.scala`uses Play configuration. The order is
   * /etc/gu/facia-tool.properties -> should contain at least the stage, this is used by EC2 instances mainly
   * application.conf -> default values
   * application.local.con -> to make it work locally
   * /etc/gu/facia-tool.application.secrets.conf -> secrets stored in a private S3 bucket
* `Aws.scala` defines two credentials
   * mandatoryCredentials -> used to access resources in CMS front stack
   * mandatoryCrossAccountCredentials -> used to access resources in frontend stack, through STS
* Build generates a `.deb` artefact instead of the usual `.zip`

@lindseydew @janua @sihil @philwills @steppenwells @gklopper @adamnfish 

Huge thanks to everyone for their advice on getting this done.


piuccio:
@adamnfish we've tried inlining cloudformation policies but it didn't work nicely, reverted.

Is this good enough for a +1 ?

adamnfish:
STS stuff looks great, nice one

piuccio:
I've gathered feedback from most people, no one feels confident enough to :+1: because of the wide scope of the PR, at the same time no one has reasons to :-1: 

I'm merging anyway, if anything comes up later I'll deal with it in another PR

rrees:
@piuccio @lindseydew could we have broken this up in some way to avoid terrifying people during the review? Any lessons we should learn?

piuccio:
@rrees code-wise, I don't think, we did our best to avoid any possible change. Maybe what was intimidating is the number of commits, which makes it look like a huge change. Because we were both working at the same time on a branch that was constantly pushed, I didn't want to rewrite the history.

janua:
I think this PR was fine to review for someone who knows the frontend codebase well.

I couldn't have a look last week because of our hack week.

PR #14, opened 2015-09-25
Copy to clipboard

Reettaphant:


uplne:
It is a nice practice to put some description what is this PR meant to do. I can get it from the code but it is helpful (sometimes including screenshots too).

facia-tool/public/css/style.css
```diff
@@ -1523,6 +1523,9 @@ button {
     text-align: center;
 }
 
+.tool--small--copy,
```
uplne: This is smelly. For me creating modifications of modifications is a bad practice.

Look at it this way:

You have 'tool' object which has some default tool styling.
You have modificator 'tool--small' which makes tool object smaller.

And now you have 'tool--small--paste' which adds some additional style to 'tool--small'. What if you will need add those styles to 'tool--large' in the future? Will you copy styles to a new class named 'tool--large--paste'?

It makes sense to create another modification 'tool--paste' which will add only 'paste' styles without knowing if the element is small, big or whatever size.

So you will end up with class="tool tool--small tool--paste" which is much more modular and easy to change or reuse in the future.

Hope it makes sense.

facia-tool/public/css/style.css
```diff
@@ -1533,9 +1536,9 @@ button {
     width: 120px;
 }
 
-.latest-articles .tool--small--copy,
-.latest-articles .tool--small--copy.large {
-    width: 83px;
+.latest-articles .tool--small--copy.large,
+.latest-articles .tool--small--copy-to-clipboard.large {
```
uplne: This is interesting. You have '.tool--small--copy.large' We are creating object which is called 'small' and making it large by adding class 'large' :-)

This would be probably fixed with the styles I've mentioned above. You would be able to do this class="tool tool--large tool--copy"

uplne:
:+1: It looks ok. CSS advice is more for the future.

piuccio:
Fully agree on the CSS, it's mostly my fault, it's now a gigantic mess, one day we should step back and rewrite it using at least some common sense.

PR #17, opened 2015-09-29
Build on travis

piuccio:
Yeah that, build on travis

@lindseydew @philwills @rich-nguyen


upload-riffraff.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
philwills: I'm a bit confused by this file. I don't understand why the upload is separate from the normal `continuous-integrations.sh` and I'd expect the ivy/lock cleanup stuff in the `.travis.yml`, like with https://github.com/guardian/status-app/blob/travis-riffraff-artifact/.travis.yml

piuccio: Didn't know about `before_cache`, I like what you do there, will change

I'm using a build matrix to speed up the build, `continuous-integrations.sh` gets called twice. If I upload from `continuous-integrations` it uploads twice, and if I upload only during scala test it uploads even when JS tests fail.

That's why I have two separate steps and `upload-riffraff.sh` is only called on `after_success`

philwills: Got it, I haven't used build matrices and hadn't thought through the consequences. That makes sense.

PR #18, opened 2015-09-29
Close without saving

Reettaphant:
When editing an article, there is now an option to close article editing without saving the changes made to it. 
close button

### DIFF
--- a/facia-tool/public/js/models/collections/collection.js
+++ b/facia-tool/public/js/models/collections/collection.js
@@ -319,6 +319,26 @@ define([
         return this.front.mode() !== 'treats' && this.history().length;
     };
 
+    Collection.prototype.replaceArticle = function(articleId) {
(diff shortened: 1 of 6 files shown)


facia-tool/public/js/models/collections/collection.js
```diff
@@ -319,6 +319,25 @@ define([
         return this.front.mode() !== 'treats' && this.history().length;
     };
 
+    Collection.prototype.replaceArticle = function(articleId) {
+        var self = this;
+        var collectionList = this.front.getCollectionList(this.raw);
+
+        _.each(collectionList, function(item, index) {
```
piuccio: We can avoid looping on all articles using `_.find()` instead of `_.each` to get `item`.

piuccio:
:+1:

PR #21, opened 2015-10-02
Add fronts permissions

lindseydew:
This adds a client implementation of the permission service. Intention is to abstract the reads of the cache into a client lib, but I will do this work separately as I wanted to release the functionality first.


piuccio:
:+1:

PR #24, opened 2015-10-06
move log initialisation to global settings

lindseydew:
Having the onStart method in global overrode the onStart methods defined in other areas of the code base. This moves the logging initialisation to a separate trait so that all the onStart methods defined here will run. 

https://fronts.code.dev-gutools.co.uk/switches 

cc @piuccio


piuccio:
:+1:

PR #25, opened 2015-10-06
Mark articles as read

Reettaphant:
Articles in the visited articles column that the user has visited will now be marked. 
screen shot 2015-10-06 at 12 05 49

### DIFF
--- a/facia-tool/public/js/models/collections/article.js
+++ b/facia-tool/public/js/models/collections/article.js
@@ -496,6 +496,7 @@ define([
 
         Article.prototype.setVisitedToTrue = function () {
             visitedArticleStorage.addArticleToStorage(this.id());
+            mediator.emit('set:article:to:visited', this.id());
             return true;
         };
 
--- a/facia-tool/public/js/models/collections/article.js
+++ b/facia-tool/public/js/models/collections/article.js
@@ -376,8 +376,6 @@ define([
 
             this.slimEditor = opts.slimEditor;
 
-            this.visited = ko.observable(opts.visited);
-
             this.state = asObservableProps([
                 'enableContentOverrides',
                 'underDrag',
@@ -386,6 +384,7 @@ define([
                 'isLiveBlog',
                 'isLoaded',
                 'isEmpty',
+                'visited',
                 'inDynamicCollection',
                 'tone',
                 'primaryTag',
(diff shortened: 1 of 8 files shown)


facia-tool/public/js/models/collections/article.js
```diff
@@ -373,6 +376,8 @@ define([
 
             this.slimEditor = opts.slimEditor;
 
+            this.visited = ko.observable(opts.visited);
```
piuccio: I know that `Article` is a mess, but I think that `visited` should be a `state`

facia-tool/public/js/models/collections/latest-articles.js
```diff
@@ -50,6 +51,10 @@ class Latest extends BaseClass {
             return title;
         }, this);
 
+        this.listenOn(mediator, 'set:article:to:visited', this.setArticleToVisited);
+
+        this.listenOn(mediator, 'set:article:to:visited', this.setArticleToVisited);
```
piuccio: You don't need two listeners

facia-tool/public/js/widgets/trail.html
```diff
@@ -6,7 +6,8 @@
         underDrag: state.underDrag() && !state.underControlDrag(),
         underControlDrag: state.underDrag() && state.underControlDrag(),
         imageHidden: meta.imageHide,
-        slimEditor: slimEditor
+        slimEditor: slimEditor,
+        visited_article: visited()
```
piuccio: for consistency this should be called `visitedArticle` and you don't need to call the function

`visitedArticle: visited` should do the job, knockout knows how to handle it

facia-tool/test/public/spec/visited.articles.spec.js
```diff
@@ -0,0 +1,72 @@
...
+
+    it('displays article as visited after redrawing data', function (done) {
+        this.testInstance.load()
+        .then(() => {
+            $('.tool--small--copy-to-clipboard', dom.latestArticle(1)).click();
+        })
+        .then(() => {
+            $('.tool--small--href', dom.latestArticle(1)).click();
+        })
+        .then(() => wait.ms(4000))
```
piuccio: do we need to wait `4s` here?

piuccio:
:+1:

PR #26, opened 2015-10-06
Viewer link

Reettaphant:
Articles that are pasted on the clipboard from viewer should be recognised as guardian content on the clipboard rather than snap links.

### DIFF
--- a/facia-tool/public/js/utils/article-path.js
+++ b/facia-tool/public/js/utils/article-path.js
@@ -1,0 +1,15 @@
+import urlHost from './url-host';
+import urlAbsPath from './url-abs-path';
+import CONST from 'constants/defaults';
+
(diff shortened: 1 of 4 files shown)


facia-tool/public/js/utils/article-path.js
```diff
@@ -0,0 +1,14 @@
+import urlHost from './url-host';
+import urlAbsPath from './url-abs-path';
+
+export default function(url) {
+
+    var host = urlHost(url);
+
+    if (host === 'viewer.gutools.co.uk') {
```
piuccio: can you move `viewer.gutools.co.uk` to `facia-tool/public/js/constants/defaults.js` please

piuccio:
good stuff :+1:

PR #28, opened 2015-10-07
Update babel

piuccio:
And fix some TODO


rrees:
This is a very large change. I think it would have been better to split it up into parts.

PR #31, opened 2015-10-12
make the user check case insensitive

lindseydew:
Does a case insensitive check on the userId to mitigate against the case where a user is in the permissions system with a different case check than the users' panda token.

Intend to port this logic to the https://github.com/guardian/editorial-permissions-client when this is up and running


piuccio:
:+1: hopefully there are no two different users whose emails differ only for the upper/lower case of their email

lindseydew:
Yeah that's a good point. My feeling is that it's an unlikely enough edge-case that we can deal with it if it arises.

PR #32, opened 2015-10-12
Get rid of DFP dependency

piuccio:
And delete a bunch of useless files


kelvin-chappell:
:+1: for the DFP dependencies.

PR #33, opened 2015-10-15
Search fronts

Reettaphant:
Adds a search column to config front. The search column allows for searching fronts by front and collection names and dragging collections to the fronts config column. 
screen shot 2015-10-15 at 11 50 02

### DIFF
--- a/facia-tool/public/js/widgets/columns/fronts-search.html
+++ b/facia-tool/public/js/widgets/columns/fronts-search.html
@@ -18,9 +18,8 @@
                     text: id
                 "></span>
             </div>
-            <div class="cnf-front__inner" data-bind="foreach: collections">
+            <div class="cnf-front__inner" data-bind="foreach: collections, makeDraggable: true">
                 <div class="cnf-collection open">
-                    <div data-bind="makeDroppable: true">
                         <span class="cnf-collection__index" data-bind="text: $index() + 1"></span>
                         <a class="cnf-collection__name" data-bind="
                             attr: {href: '/' + id, title: id},
--- a/facia-tool/public/js/widgets/columns/fronts-search.html
+++ b/facia-tool/public/js/widgets/columns/fronts-search.html
@@ -23,6 +23,16 @@
                         attr: {href: '/' + id, title: id},
                         click: function() { return false;},
                         text: meta.displayName() || '(no title)'"></a>
+                    <span class="placements" data-bind="if: parents().length > 1">
+                        also on
+                        <span data-bind="foreach: parents">
+                            <!-- ko if: $parents[1].id !== id -->
+                                <span class="cnf-collection__also-on" data-bind="
+                                    text: id"
+                                ></span>
+                            <!-- /ko -->
+                        </span>
+                    </span>
                 </li>
             </ol>
         </div>
--- a/facia-tool/public/js/widgets/columns/fronts-search.html
+++ b/facia-tool/public/js/widgets/columns/fronts-search.html
@@ -6,10 +6,9 @@
         <input
           type="text"
           placeholder="search for collections"
-          data-bind='
-            value: searchTerm,
-            valueUpdate: ["afterkeydown", "afterpaste"]
-        '/>
+          data-bind="
+            textInput: searchTerm
+        "/>
     </div>
     <div class="cnf-fronts open" data-bind="foreach: searchedFronts">
         <div class="cnf-front open">
--- a/facia-tool/public/js/widgets/columns/fronts-search.html
+++ b/facia-tool/public/js/widgets/columns/fronts-search.html
@@ -1,0 +1,34 @@
+<div class="col__inner scrollable">
+    <div class="title">
+        Search
+    </div>
+    <div class="search-form">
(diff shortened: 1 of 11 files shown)


facia-tool/public/js/widgets/columns/fronts-search.html
```diff
@@ -0,0 +1,31 @@
+<div class="col__inner scrollable">
+    <div class="title">
+        Search
+    </div>
+    <div class="search-form">
+        <input
+          type="text"
+          placeholder="search for fronts"
```
piuccio: 'search for collections'

facia-tool/public/js/widgets/columns/fronts-search.html
```diff
@@ -0,0 +1,31 @@
...
+          data-bind='
+            value: searchTerm,
+            valueUpdate: ["afterkeydown", "afterpaste"]
```
piuccio: I know that I've told you to copy from autocomplete, but can you try [textInput binding](http://knockoutjs.com/documentation/textinput-binding.html). Should work better

facia-tool/public/js/widgets/columns/fronts-search.js
```diff
@@ -0,0 +1,72 @@
...
+import ColumnWidget from 'widgets/column-widget';
+import CONST from 'constants/defaults';
+import generateCollections from 'utils/generate-collections';
+
+export default class SearchConfig extends ColumnWidget {
+    constructor(params, element) {
+        super(params, element);
+        this.searchTerm = ko.observable('');
+        this.searchedFronts = ko.observableArray([]);
+        var originalFronts = _.reduce(JSON.parse(JSON.stringify(this.baseModel.frontsList())), function (frontList, front) {
```
piuccio: I don't like `JSON.parse(JSON.stringify` here. I think it's more readable if you reduce `this.baseModel.frontsList()` and then clone `front` inside the callback.
If you want to clone a nested object you can use `utils/clean-clone` otherwise a simple `_.extend` works fine.

You don't necessarily have to keep the same `front` structure, you can create a new object that better suits your needs, for instance you can do `front.id.toLowerCase()` here instead of having to do it on every search.

Keep in mind that originalFronts changes with time and can be updated while the app is running. For instance if you create a new front or someone else does it, you should still be able to find those through a search.
You can subscribe to `vars.model.state()` or maybe there's already a notification mechanism, I can't remember.

facia-tool/public/js/widgets/columns/fronts-search.js
```diff
@@ -0,0 +1,72 @@
...
+        var self = this;
+        this.searchTerm.subscribe(function (newTerms) {
+
+            if (newTerms.length < 2) {
+                self.searchedFronts([]);
+
+            } else {
+
+                newTerms = newTerms.toLowerCase();
+                var allSearchTerms = newTerms.match(/\S+/g);;
```
piuccio: two `;;` here

facia-tool/test/public/spec/config.spec.js
```diff
@@ -133,4 +116,54 @@ describe('Config', function () {
...
+        var mockConfig = this.testInstance.mockConfig,
+            baseModel;
+
+        this.testInstance.load()
+        .then(model => {
+            baseModel = model;
+            return dom.type($('input'), 'uk');
+        })
+        .then(() => {
+            $('.content-pane:nth(0) .title--text:nth(1)').click();
```
piuccio: can you assert here that the search displays the collection you expect?

piuccio:
apart from comments, good stuff. We should decide if the search deserves a place in the default list of columns when someone hits `/priority/config`

facia-tool/public/js/widgets/columns/fronts-search.js
```diff
@@ -0,0 +1,81 @@
...
+            originalFronts =  _.reduce(self.baseModel.frontsList(), function (frontList, front) {
+
+                if (_.every(CONST.askForConfirmation, function (element) {
+                    return front.id !== element;
+                }) ) {
+
+                    let frontToSearch = { id: front.id.toLowerCase() };
+
+                    frontToSearch.collections = generateCollections(front.collections);
+                    frontList.push(frontToSearch);
```
piuccio: nitpicking, how about

```js
frontList.push({
   id: front.id.toLowerCase(),
   collections: generateCollections(front.collections)
});
```

facia-tool/public/js/widgets/columns/fronts-search.js
```diff
@@ -0,0 +1,81 @@
...
+                }
+                return frontList;
+            }, []);
+        }
+
+        this.searchTerm.subscribe(function (newTerms) {
```
piuccio: it is safe to subscribe to observables in the same class, but I still prefer to use `subscribeOn`, for consistency.

Also it would be nice to put the body of this function in its own method, say `search`

```js
class {
   constructor() {}
   search() {}
}
```

piuccio:
:+1:

piuccio:
:+1:

PR #36, opened 2015-10-16
Update panda

piuccio:
And get rid of the old play google auth dependency.

@lindseydew @janua


janua:
Bye play googleauth!
:+1:

PR #37, opened 2015-10-16
modify group item list at the correct index

Reettaphant:
Fixes a bug where, if a front had multiple groups, closing articles that belonged to a group that did not belong to the first group on the list, would create duplicate copies of that article.


piuccio:
:+1:

PR #38, opened 2015-10-16
add eslint plugin for jasmine

Reettaphant:
Add eslint plugin for jasmine and display errors if there are focused or disabled tests.


piuccio:
:+1:

PR #39, opened 2015-10-16
Better validation for images coming from the Grid

piuccio:
Any image URL containing a Grid id is now transformed into it's actual crop image.
If an image doesn't have a valid crop, show an error.
It's now possible (again) to drag the main image from an article page.
There's also some better logic for imgix URLs

### DIFF
--- a/facia-tool/test/public/spec/validate.image.spec.js
+++ b/facia-tool/test/public/spec/validate.image.spec.js
@@ -1,122 +1,336 @@
-import {CONST} from 'modules/vars';
+import ko from 'knockout';
+import * as vars from 'modules/vars';
 import validate from 'utils/validate-image-src';
+import GridUtil from 'grid-util-js';
+import images from 'test/utils/images';
 
 describe('Validate images', function () {
     beforeEach(function () {
-        this.originalCDNDomain = CONST.imageCdnDomain;
+        images.setup();
+
+        this.originalModel = vars.model;
+        vars.setModel({
+            state: ko.observable({
+                defaults: { apiBaseUrl: '/api.grid', mediaBaseUrl: 'http://media' }
+            })
+        });
+
+        validate.gridInstance = new GridUtil('/api.grid');
     });
     afterEach(function () {
-        CONST.imageCdnDomain = this.originalCDNDomain;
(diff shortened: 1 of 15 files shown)


piuccio:
@paperboyo you're gonna like this :grin:

paperboyo:
:gem:

facia-tool/test/public/spec/validate.image.spec.js
```diff
@@ -1,122 +1,336 @@
...
+                    expect(image.height).toBe(140);
+                    expect(image.src).toMatch(/square\.png$/);
+                    expect(image.origin).toBe('http://media/image/1234567890123456789012345678901234567890');
+                })
+                .then(done)
+                .catch(done.fail);
+            });
+        });
+
+        describe('- link does not include crop id', function () {
```
theefer: How do you choose a crop in that case?

piuccio: display an alert saying there are no valid crops. In the next PR you'll also have a button to open the grid and create a crop

theefer: From the tests below it seems to pick the first acceptable export?

The reason I'm asking is because we're thinking of what to do when people drag a non-crop image into the different tools, so it might be nice if the behaviour was uniform.

piuccio: I guess you just figured out that I replied without even looking at the question.

`link does not include crop id` means you paste something like `https://grid.yeah/images/1234..40` instead of the crop url `https://grid.yeah/images/1234..40?crop=xyz`
what happens after that is
- find any crop that matches the constraints (5:3 and min max width)
- if success use the crop
- if fail give a warning

theefer: Can we have a quick chat on Monday? Would be nice to think ahead so the behaviour is eventually consistent with the behaviour in Composer & other tools.

paperboyo: What if an image is not 5:3 but contains a 5:3, but not one, but several? Very slight misspelling: `extreactImageElements`

PR #41, opened 2015-10-19
Move uncached assets out of common

piuccio:
I'm trying to get rid of `/common` so I'm moving all the assets related logic out of `common` and inside `facia-tool`.

@janua does this look sane to you?


janua:
Kind of don't know what some of that code is. But a lot of it is scala, so if it compiles it should be fine!

piuccio:
:smile: I don't know either, but it compiles, and loads assets...

PR #45, opened 2015-10-28
do not reorder group items if target item and source item are same

Reettaphant:
When dragging and dropping an article back to its original location, there no reordering of group items should be visible. Currently the dropped article gets first placed at the end of the group list and then is moved back to its original location.


piuccio:
:+1:

PR #47, opened 2015-10-30
New switches

Reettaphant:
Use the new switchboard app to manage switches in facia tool.

### DIFF
--- a/common/app/common/configuration.scala
+++ b/common/app/common/configuration.scala
@@ -368,6 +368,10 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
     lazy val key = playConfiguration.getStringFromStage("media.key");
   }
 
+  object switchBoard {
+    val bucket = playConfiguration.getMandatoryStringFromStage("aws.switchesBucket")
+  }
+
   object aws {
 
     lazy val region = playConfiguration.getMandatoryStringFromStage("aws.region")
--- a/common/app/common/configuration.scala
+++ b/common/app/common/configuration.scala
@@ -369,8 +369,8 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
   }
 
   object switchBoard {
-    val bucket = playConfiguration.getMandatoryStringFromStage("aws.switchesBucket")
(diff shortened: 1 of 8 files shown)


common/app/common/configuration.scala
```diff
@@ -368,6 +368,11 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
     lazy val key = playConfiguration.getStringFromStage("media.key");
   }
 
+  object switchBoard {
+    val bucket = playConfiguration.getMandatoryStringFromStage("aws.switchesBucket")
```
piuccio: let's call them

```
switchboard.bucket  (instead of aws.switchesBucket)
switchboard.object   (instead of newSwitches.key)
```

`key` gives me more the idea of an api key, not an object key in S3

facia-tool/app/Global.scala
```diff
@@ -11,7 +12,7 @@ object Global extends WithFilters(Gzipper)
   with GlobalSettings
   with CloudWatchApplicationMetrics
   with ConfigAgentLifecycle
-  with SwitchboardLifecycle
```
piuccio: we still need the old `SwitchboardLifecycle`

switchboard/Switchboard.scala
```diff
@@ -0,0 +1,47 @@
...
+import switches.switchManager
+import play.api.GlobalSettings
+import play.api.{Logger, Application, GlobalSettings}
+import play.api.libs.json.Json
+
+trait NewSwitchboardLifecycle extends GlobalSettings with ExecutionContexts with Logging {
+
+  override def onStart(app: Application) {
+    super.onStart(app)
+    Jobs.deschedule("SwitchBoardRefreshJob")
```
piuccio: I just realized that `Jobs` comes from `common` and it's quite a complex class.

Can you try using `Akka.scheduler`

Should work like this
```
Akka.scheduler.schedule(0.seconds, 1.minute) { refresh() }
```

piuccio:
looks good apart from small comments

piuccio:
:+1:

PR #49, opened 2015-10-30
Display publisher

Reettaphant:
Display the name of the publisher in breaking news: 
screen shot 2015-10-30 at 15 26 38


piuccio:
:+1:

PR #50, opened 2015-10-30
Front limit warnign

Reettaphant:
Display a front limit warning in config if the number of fronts remaining is below 10. 
screen shot 2015-10-30 at 16 10 06

### DIFF
--- a/facia-tool/public/js/widgets/columns/fronts-config.js
+++ b/facia-tool/public/js/widgets/columns/fronts-config.js
@@ -17,6 +17,14 @@ export default class FrontConfig extends ColumnWidget {
         this.populate();
         this.subscribeOn(this.baseModel.state, this.populate);
         this.loaded = Promise.resolve();
+        this.calculateRemainingFronts();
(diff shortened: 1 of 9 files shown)


facia-tool/public/js/widgets/columns/fronts-config.js
```diff
@@ -7,6 +7,8 @@ import frontCount from 'utils/front-count';
 import ColumnWidget from 'widgets/column-widget';
 import alert from 'utils/alert';
 
+var FRONT_ALERT_LIMIT = 10;
```
piuccio: constants go into `constants/defaults.js`

facia-tool/public/js/widgets/message.js
```diff
@@ -0,0 +1,9 @@
+import ko from 'knockout';
+
+class Message {
+    constructor() {
+        this.frontsRemainingMessage = ko.observable();
```
piuccio: message is a generic class/widget, let's rename `frontsRemainingMessage` into some reusable, like `textMessage` or something like that

piuccio:
:+1:

PR #52, opened 2015-11-02
replace article when dropped above itself

Reettaphant:
When article is dragged above itself in another column, any edited changes in the source article should appear in the target article.

### DIFF
--- a/facia-tool/public/js/utils/drag-dispatcher.js
+++ b/facia-tool/public/js/utils/drag-dispatcher.js
@@ -58,18 +58,21 @@ function handleInternalClass ({sourceItem, sourceGroup}, targetItem, targetGroup
 
     var {position, target, isAfter} = normalizeTarget(sourceItem, targetItem, targetGroup);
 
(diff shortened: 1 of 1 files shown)


facia-tool/public/js/utils/drag-dispatcher.js
```diff
@@ -56,16 +56,22 @@ function handleMedia ({sourceItem, mediaItem}, targetItem, targetGroup) {
...
-    if (targetItem instanceof Article && targetItem.id() === sourceItem.id) {
-        return;
-    }
-
     var {position, target, isAfter} = normalizeTarget(sourceItem, targetItem, targetGroup);
 
-    removeById(targetGroup.items, urlAbsPath(sourceItem.id));
+    var insertAt = 0;
+
+    for (var i = 0; i < targetGroup.items().length; i++) {
```
piuccio: I'd cache `targetGroup.items()` outside the for loop to avoid calling that function twice at each iteration

piuccio:
Apart for comment, I've tested on all use cases that come to mind, seems fine :+1:

PR #53, opened 2015-11-02
Get rid of panda secrets for local dev

piuccio:
We never used them for prod anyway

@jamespamplin thanks for the tip


jamespamplin:
:+1: nice one!

PR #57, opened 2015-11-06
Use sts assume role to access panda bucket

lindseydew:
Following the retro action this makes the fronts tool access the panda bucket using assume role, so we can deprecate the specific bucket policy.

cc @piuccio @jamespamplin @sihil

### DIFF
--- a/common/app/common/configuration.scala
+++ b/common/app/common/configuration.scala
@@ -102,6 +102,7 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
     lazy val pandomainHost = playConfiguration.getMandatoryStringFromStage("faciatool.pandomain.host")
     lazy val pandomainDomain = playConfiguration.getMandatoryStringFromStage("faciatool.pandomain.domain")
     lazy val pandomainService = playConfiguration.getMandatoryStringFromStage("faciatool.pandomain.service")
+    lazy val pandomainRoleArn = playConfiguration.getMandatoryStringFromStage("faciatool.pandomain.roleArn")
 
     lazy val logStream = playConfiguration.getMandatoryStringFromStage("logging.kinesis.stream")
     lazy val logStreamRegion = playConfiguration.getMandatoryStringFromStage("logging.kinesis.region")
--- a/facia-tool/app/auth/PanDomainAuthActions.scala
+++ b/facia-tool/app/auth/PanDomainAuthActions.scala
@@ -1,16 +1,15 @@
 package auth
 
-import com.amazonaws.auth.BasicAWSCredentials
-import conf.Configuration
+import com.amazonaws.auth.profile.ProfileCredentialsProvider
+import com.amazonaws.auth.{AWSCredentialsProviderChain, STSAssumeRoleSessionCredentialsProvider}
+import com.gu.pandomainauth.PanDomainAuth
 import com.gu.pandomainauth.action.AuthActions
 import com.gu.pandomainauth.model.AuthenticatedUser
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
+import conf.Configuration
 import play.api.Logger
 import play.api.mvc._
-import conf.aws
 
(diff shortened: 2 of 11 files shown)


facia-tool/app/auth/PanDomainAuthActions.scala
```diff
@@ -42,8 +45,8 @@ trait PanDomainAuthActions extends AuthActions with Results {
 
   override lazy val domain: String = Configuration.faciatool.pandomainDomain
   override lazy val system: String = Configuration.faciatool.pandomainService
+  override def awsCredentialsProvider  = new AWSCredentialsProviderChain(
+    new STSAssumeRoleSessionCredentialsProvider(Configuration.faciatool.pandomainRoleArn, "workflow")
```
piuccio: this file probably conflicts with master, once resolved, I think `workflow` should be renamed as `fronts`.

facia-tool/build.sbt
```diff
@@ -20,3 +20,5 @@ javaOptions in Universal ++= Seq(
     "-J-XX:+PrintGCDateStamps",
     s"-J-Xloggc:/var/log/${packageName.value}/gc.log"
 )
+
+ivyScala := ivyScala.value map { _.copy(overrideScalaVersion = true) }
```
piuccio: not sure what's this for

facia-tool/cloudformation/facia.json
```diff
@@ -25,6 +25,11 @@
             "Description": "Composer role to assume access to the kinesis log stash stream",
             "Type": "String",
             "Default": "arn:aws:iam::743583969668:role/FrontsRoleForLogStash-FaciaToolRole-GX5KBXQZGL5X"
+        },
+        "PandaRoleToAssume": {
+            "Description": "Composer role to assume access to the kinesis log stash stream",
```
piuccio: the description is the one for composer + logs

piuccio:
other than comments it looks good, I prefer STS to whatever other solution we were using

facia-tool/cloudformation/facia.json
lindseydew: copy and paste fail. thanks

facia-tool/app/auth/PanDomainAuthActions.scala
lindseydew: Ah ok, I chose "workflow" because that's the account the of the role to assume, but I'm happy for this to be "fronts"

adamnfish: Is it worth leaving the secret in the chain for applications that don't live in AWS (which of course includes local development)?

facia-tool/build.sbt
lindseydew: Added this to resolve compilation errors which occurred as a result of upgrading panda, but it turned out the fix was to upgrade sbt. Can remove.

common/app/common/configuration.scala
```diff
@@ -363,6 +363,8 @@ class GuardianConfiguration(val application: String, val webappConfDirectory: St
     lazy val pandomainHost = playConfiguration.getMandatoryStringFromStage("faciatool.pandomain.host")
     lazy val pandomainDomain = playConfiguration.getMandatoryStringFromStage("faciatool.pandomain.domain")
     lazy val pandomainService = playConfiguration.getMandatoryStringFromStage("faciatool.pandomain.service")
+    lazy val pandomainRoleArn= playConfiguration.getMandatoryStringFromStage("faciatool.pandomain.roleArn")
```
adamnfish: *if* you're completely removing the secret then the config properties should come out too. As I mentioned it might be worth leaving them in the chain, in which case perhaps they should be optional properties (line 368)

piuccio: they were removed already in master, the credentials provider should include
```
new ProfileCredentialsProvider("CMS Fronts")
```

adamnfish: Much better :-)

PR #63, opened 2015-11-11
used mirror image thumbnail for slow-v-half

Reettaphant:
fixed/small/slow-V-half container should appear as a mirror image in thumbnail. 
screen shot 2015-11-11 at 16 35 30


piuccio:
:+1:

PR #65, opened 2015-11-16
Permissions for breaking news

piuccio:
The ability to modify collections inside `breaking-news` front is now controlled by a permission in permissions API and a switch in the switchboard

@janua


facia-tool/app/controllers/DefaultsController.scala
```diff
@@ -0,0 +1,102 @@
...
+          Configuration.media.baseUrl.get,
+          Configuration.media.apiUrl.get,
+          FixedContainers.all.keys.toSeq.map(id => ContainerJsonConfig(id, None)),
+          DynamicContainers.all.keys.toSeq.map(id =>
+            if (id == "dynamic/package") {
+              ContainerJsonConfig(id, Some(DynamicPackage))
+            } else if (id == "dynamic/slow-mpu") {
+              ContainerJsonConfig(id, Some(DynamicMpu))
+            } else {
+              ContainerJsonConfig(id, Some(DynamicGroups))
```
janua: Are you sure you want the above inline?

piuccio: the above didn't change, I'll probably hardcode all that into `defaults.js` because that list comes from `common` which is out of sync with `frontend`

facia-tool/app/util/Acl.scala
```diff
@@ -0,0 +1,51 @@
...
+import play.api.libs.json.{JsBoolean, JsValue, Json, Writes}
+import switchboard.SwitchManager
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
+
+object Authorization {
+  implicit val authorizationWrites = new Writes[Authorization] {
+    def writes(access: Authorization): JsValue = access match {
+      case AccessGranted => JsBoolean(true)
+      case AccessDenied => JsBoolean(false)}}}
```
janua: :smile:

facia-tool/app/util/Acl.scala
```diff
@@ -0,0 +1,51 @@
...
+
+case class AclJson (
+  fronts: Map[String, Authorization],
+  permissions: Map[String, Authorization]
+)
+
+object Acl {
+  def testUser(permission: Permission, switch: String)
+              (email: String): Future[Authorization] = {
+    implicit def permissionsUser: PermissionsUser = PermissionsUser(email)
```
janua: `val` maybe? Why does it need to be `implicit`, is the `Permissions.get` using it?

janua:
Looks good! :+1:

facia-tool/app/util/Acl.scala
piuccio: `implicit` because it's used by Permissions, `def` because @jamespamplin and I tried few options to get it working, let me check if it works as `val`

jamespamplin: Unfortunately yes, this needs to be implicit as the api in the permissions client requires it. I went through a phase of liking implicit things, sorry about this.

PR #67, opened 2015-11-24
don't match hash when getting article path

Reettaphant:
Fixes the following bug: When copying a link from viewer tools, if the link ends in a hash, the copying fails.


piuccio:
:+1:  nice

PR #72, opened 2015-11-30
put private if packages project

Reettaphant:
Use putPrivate if using the packages project.


piuccio:
:+1: if it compiles on travis :smile:

PR #76, opened 2015-12-01
Stream update all the things

piuccio:
Any action that triggers a database change is sent to the update stream. The message also includes time and the affected fronts.

@janua mind having a quick look?


janua:
No plus 1?

piuccio:
implicit

janua:
No implicit here!

PR #77, opened 2015-12-02
packages project

piuccio:
Define another project `packages` using the same source folder but a different target


piuccio:
packages was forked into https://github.com/guardian/story-packages

PR #84, opened 2015-12-16
Advertisement priority

Reettaphant:
Add a route for fronts that have priority advertisement, are always hidden and don't allow for editing metadata.

screen shot 2015-12-16 at 15 17 22

### DIFF
--- a/public/src/js/widgets/columns/fronts-config.html
+++ b/public/src/js/widgets/columns/fronts-config.html
@@ -126,7 +126,10 @@
              <!-- ko if: id -->
                 <div class="instructions">
                     <!-- ko if: !state.isOpenProps() && collections.items().length -->
-                        <span class="linky tool--metadata" data-bind="click: openProps">edit metadata</span>
+                        <span class="linky tool--metadata" data-bind="
+                          click: openProps,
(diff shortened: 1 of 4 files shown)


public/src/js/widgets/columns/fronts-config.html
```diff
@@ -126,7 +126,10 @@
              <!-- ko if: id -->
                 <div class="instructions">
                     <!-- ko if: !state.isOpenProps() && collections.items().length -->
-                        <span class="linky tool--metadata" data-bind="click: openProps">edit metadata</span>
+                        <span class="linky tool--metadata" data-bind="
+                          click: openProps,
+                          visible: !state.isMetaLocked
```
piuccio: does this works for other priorities? you only define `state.isMetaLocked` when priority is `advertisement`. Which by the way should be an observable, just for consistency

Reettaphant: Yes it works, but I guess because it is not an observable. I tried to replicate the behaviour of the isTypeLocked property, so shall I change that to an observable as well?

piuccio:
sorry, but we decided a new priority is overkill, we'll just create a new hidden front

PR #86, opened 2015-12-18
include link to live front is pasted from preview

Reettaphant:
Include a link to the live front if a front is pasted from preview. 
screen shot 2015-12-18 at 11 32 23


piuccio:
:+1:

PR #87, opened 2015-12-21
mobile notification client

piuccio:
Send breaking news update to the new mobile notification client. It'll take care of sending alert to phones


piuccio:
tested with @alexduf works fine on CODE

PR #90, opened 2016-01-04
display error message only after multiple fails

Reettaphant:
Display error message if fetching latest articles fails only after it has failed at least two times.


piuccio:
:+1:

PR #91, opened 2016-01-05
drag pictures from grid that are not crops

Reettaphant:
Allow for dragging pictures from the grid that are not crops to the clipboard.


piuccio:
:+1:  mind doing the same PR on story packages as well?

Reettaphant:
Yes, I'll do so later.

PR #96, opened 2016-01-08
Hidden front notification

Reettaphant:
Display a notification on editor if front is hidden.

screen shot 2016-01-08 at 12 07 58

### DIFF
--- a/public/src/css/style.css
+++ b/public/src/css/style.css
@@ -1337,7 +1337,7 @@ button {
     color: red;
 }
 
-.front-hidden > span {
+.front-hidden-label {
     padding-left: 2px;
 }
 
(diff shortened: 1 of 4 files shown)


piuccio:
can we make it more prominent? maybe using the blue bar or adding the message below the front picker

public/src/css/style.css
```diff
@@ -1331,6 +1331,16 @@ button {
     max-width: 100%;
 }
 
+.front-hidden {
+    font-size: 12px;
+    padding: 3px 0px 3px 20px;
+    color: red;
+}
+
+.front-hidden > span {
```
piuccio: Instead of targeting a `span` it's better to add a class to that element

piuccio:
:+1:

PR #97, opened 2016-01-08
don't show contributor image if more than one contributor

Reettaphant:
Do not show the contributor image cutout if the article has more than contributor.


piuccio:
nice :+1:

PR #107, opened 2016-01-15
Breaking news collection history

Reettaphant:
Display the time that last alert was sent on each collection
screen shot 2016-01-15 at 11 03 15

### DIFF
--- a/public/src/css/style.css
+++ b/public/src/css/style.css
@@ -34,6 +34,33 @@ a:hover,
     color: #2ea3eb;
 }
 
+.collections-published {
+    background-color: #fff;
+    padding: 5px 10px;
+    box-shadow: 0 1px 6px rgba(0, 0, 0, 0.32), 0 1px 6px rgba(0, 0, 0, 0.32);
+    border: 1px solid #ddd;
+    margin: 16px 8px 16px 8px;
+    display: flex;
+    flex-flow: row wrap;
+}
+
+.published-collection {
+    margin: auto;
+    padding: 4px;
+    flex: 1 0 90px;
+}
+
+.collection-age {
+    font-size: 12px;
+    color: #CCCCCC;
+}
+
+.collection-title {
+    font-size: 14px;
+    color: #666666;
+    font-family: Georgia;
+}
+
 .configPane {
     position: relative;
     width: 100%;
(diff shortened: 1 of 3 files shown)


public/src/css/style.css
```diff
@@ -34,6 +34,33 @@ a:hover,
     color: #2ea3eb;
 }
 
+.collections-published {
+    background-color: #fff;
+    padding: 5px 10px;
+    box-shadow: 0 1px 6px rgba(0, 0, 0, 0.32), 0 1px 6px rgba(0, 0, 0, 0.32);
```
piuccio: `box-shadow` and few other properties are already defined in `cardContainer` we can avoid duplication

public/src/js/models/collections/collection.js
```diff
@@ -509,6 +509,24 @@ define([
         }
     };
 
+    Collection.prototype.lastAlertSent = function () {
+        var groupWithPublicationTime = _.find(this.groups, function (group) {
```
piuccio: once you rebase you can use arrow functions

public/src/js/models/collections/collection.js
```diff
@@ -509,6 +509,24 @@ define([
         }
     };
 
+    Collection.prototype.lastAlertSent = function () {
```
piuccio: I think `lastAlertSent` should be a `ko.pureComputed` otherwise it won't update when you send an alert

public/src/css/style.css
piuccio: forget about this, it only exists in story packages

public/src/js/models/collections/collection.js
```diff
@@ -93,6 +93,25 @@ export default class Collection extends BaseClass {
         this.setPending(true);
         this.loaded = this.load().then(() => onDomLoad);
         this.state.visibleCount({});
+
+
+        this.lastAlertSent = ko.pureComputed (() => {
```
piuccio: can we keep the constructor as small as possible?

`this.lastAlertSent = ko.pureComputed(this.getLastAlertTime, this)`

and then the logic is implemented in that function

piuccio:
great :+1:

PR #109, opened 2016-01-15
collection history background to grey

Reettaphant:
Light grey background for collection history
screen shot 2016-01-15 at 12 54 54


piuccio:
:+1:

PR #114, opened 2016-01-18
store default truthy values

piuccio:
prior to this, client and server had a list of defaults that depends on the article tags.
with this, when a value is not specified it always means `false`.

@janua if I understand it correctly, this PR alone doesn't have any effect, new comments will have `showQuotedHeadline:true` while old comments have it empty.

Step 2: Rewrite all collections and be explicit about those defaults

Step 3: Shuffle around the default handling in frontend.

right?


janua:
Yes, step 3 is correct. Not much will need to be done; in fact it may just already work with explicit booleans. It's when booleans **don't** exist that causes this implicit comment style issue to arise.

Otherwise it will just be something quite simple in the backend, such as not falling back to `getOrElse` for when the getting fails of a `metadata` field.

PR #115, opened 2016-01-18
add an extra check for groups when creating commercial fronts

Reettaphant:
Require specifying a group for all new commercial fronts as they are created.

### DIFF
--- a/public/src/js/models/config/front.js
+++ b/public/src/js/models/config/front.js
@@ -149,6 +149,10 @@ export default class ConfigFront extends BaseClass {
     }
 
     isUnderCreation() {
+        if (this.hideCreationForm()) {
+            return false;
+        }
+
         var id = this.id();
         if (!id) {
             return true;
--- a/public/src/js/models/config/front.js
+++ b/public/src/js/models/config/front.js
@@ -136,33 +136,23 @@ export default class ConfigFront extends BaseClass {
 
         this.groups = ko.observableArray(vars.CONST.frontGroups);
 
-        this.hideCreationForm = ko.observable(false);
-        this.underCreation = ko.pureComputed(this.isUnderCreation, this);
+        this.isNew = ko.observable(false);
     }
 
-    hideCreation() {
+    createFront() {
         if (!this.id() || (this.props.priority() === 'commercial' && !this.props.group())) {
             alert('You must select all properties');
             return;
         }
-        this.hideCreationForm(true);
-    }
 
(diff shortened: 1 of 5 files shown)


public/test/spec/config.front.spec.js
```diff
@@ -89,28 +89,31 @@ describe('Config Front', function () {
             $('.title .linky').click();
             // some form of validation
             dom.type('.input-url-path', '/something////here/');
-            // There's an onchange event, no need to click on save
 
-            expect(textInside('.title--text')).toBe('something/here');
+            $('.create-new-front').click();
+            return wait.ms(4000)
```
piuccio: we should do without this delay

public/src/js/models/config/front.js
```diff
@@ -135,6 +135,29 @@ export default class ConfigFront extends BaseClass {
...
+            return;
+        }
+        this.hideCreationForm(true);
+    }
+
+    isUnderCreation() {
+        if (!this.id()) {
+            return true;
+        }
+        if (!(_.find(vars.model.frontsList(), front => {
```
piuccio: you can cachec `id = this.id()` outside the loop and simplify the `_.find`. Moreover it's better to avoid negating the condition, just negate the result, it's easier to read.

public/src/js/models/config/front.js
```diff
@@ -135,6 +135,24 @@ export default class ConfigFront extends BaseClass {
...
+        this.isNew = ko.observable(false);
+    }
+
+    createFront() {
+        if (!this.id() || (this.props.priority() === 'commercial' && !this.props.group())) {
+            alert('You must select all properties');
+            return;
+        }
+
+        if (_.some(Object.keys(vars.model.state().config.fronts), frontId => {
```
piuccio: you can rewrite this as `if (vars.model.state().config.fronts[this.id()])`

public/test/spec/config.spec.js
```diff
@@ -65,6 +65,7 @@ describe('Config', function () {
             return configAction(mockConfig, baseModel, () => {
                 dom.click(dom.$('.title .linky'));
                 dom.type($('.cnf-form input[type=text]'), 'test/front');
+                dom.click(dom.$('.create-new-front'));
                 // There's an onchange event, no need to click on save
```
piuccio: can you remove this comment please?

piuccio:
:+1:

PR #116, opened 2016-01-20
create new front after dropping collection

Reettaphant:
Allow for saving a new front when a collection is dropped into an empty front.


piuccio:
looks good, the only piece missing is updating `also on`

piuccio:
:+1:  this is not anymore about creating a front but prevent dragging until the front is created

PR #133, opened 2016-02-04
Write to two buckets

Reettaphant:
Write two buckets, and history to the new bucket only.


piuccio:
@janua here starts the migration

piuccio:
:+1:

PR #134, opened 2016-02-10
add backfill to to collection config json if api query

Reettaphant:


piuccio:
:+1: tested on CODE with @alexduf , doesn't break MAPI

alexduf:
tested indeed! :shipit:

PR #136, opened 2016-02-11
log throwable on refresh

piuccio:
@janua is this sensible?


janua:
Yep, looks good!
:+1:

PR #148, opened 2016-03-11
count articles in individual collections

Reettaphant:
Count articles in individual collections when limiting number of articles.

### DIFF
--- a/public/src/js/widgets/columns/fronts.js
+++ b/public/src/js/widgets/columns/fronts.js
@@ -8,6 +8,7 @@ import humanTime from 'utils/human-time';
 import mediator from 'utils/mediator';
 import * as sparklines from 'utils/sparklines';
 import ColumnWidget from 'widgets/column-widget';
+import deepGet from 'utils/deep-get';
 
 export default class Front extends ColumnWidget {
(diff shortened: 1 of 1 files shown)


public/src/js/widgets/columns/fronts.js
```diff
@@ -290,14 +290,12 @@ export default class Front extends ColumnWidget {
         }
 
         var collectionCap  = this.baseModel.state().defaults.collectionCap;
-        var articleNumber = this.collections().reduce((numberOfArticles, collection) => {
-            return numberOfArticles + collection.groups.reduce((numberOfArticles, group) => {
-                return numberOfArticles + group.items().length;
-            }, 0);
+        var articleNumber = item.group.parent.groups.reduce((numberOfArticles, group) => {
```
piuccio: the long accessor is quite scary `item.group.parent.groups`, it's safer to use `deepGet`.

Can you also check that this function is not called when you drag a sublink?

Reettaphant: I've checked, it's not called with sublinks.

piuccio:
:+1:

PR #149, opened 2016-03-14
Remove api query

Reettaphant:
Remove api query and upgrade fapi client.


piuccio:
fails in the tests but the code looks good

piuccio:
:+1:

PR #159, opened 2016-03-23
refresh last alert times

Reettaphant:
Update the last alert times for breaking news collections without refreshing the page.


piuccio:
nice and clean :+1:

PR #160, opened 2016-03-24
Compile time dependency injection

piuccio:
In preparation of Play 2.5

thanks to @cb372


piuccio:
tested on CODE all seems fine. The only noticeable change is cloudwatch metrics, they're now called `facia-tool` instead of `frontend-facia-tool` which is a fine thing but radiator won't have any metrics for a while.

PR #163, opened 2016-04-07
Fix a race condition

piuccio:
maybe

@janua

### DIFF
--- a/app/services/S3.scala
+++ b/app/services/S3.scala
@@ -92,14 +92,14 @@ trait S3 {
     metadata.setContentType(contentType)
     metadata.setContentLength(value.getBytes("UTF-8").length)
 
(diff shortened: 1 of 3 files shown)


janua:
This won't fix the race condition if it's a race condition...

piuccio:
it won't fix two users writing at the same time but it prevents the same user from issuing another request until the previous put is complete.

maybe 😁

janua:
Seeing as this is synchronous anyway, it probably is waiting already even without the return type.

piuccio:
ok this new update should really solve it. While a request is on progress we apply a `pending` class on the collection, this prevents user actions such as adding another story. However the class is not applied on the publish/discard buttons so it's possible to publish before the previous edit completes.

PR #164, opened 2016-04-08
Specify image source

Reettaphant:
Save image source as an object with src and origin attributes.

### DIFF
--- a/public/src/js/constants/article-meta-fields.js
+++ b/public/src/js/constants/article-meta-fields.js
@@ -41,7 +41,7 @@ export default Object.freeze([
     },
     {
         key: 'imageSrc',
-        editable: true,
+        editable: false,
         dropImage: true,
         omitForSupporting: true,
         visibleWhen: 'imageReplace',
@@ -62,6 +62,30 @@ export default Object.freeze([
             }
         },
         type: 'text'
+    },
+    {
+        key: 'imageSource',
+        editable: true,
+        dropImage: true,
+        omitForSupporting: true,
+        visibleWhen: 'imageReplace',
+        label: 'replacement image URL',
+        validator: {
+            params: {
+                src: 'imageSrc',
+                width: 'imageSrcWidth',
+                height: 'imageSrcHeight',
+                origin: 'imageSrcOrigin',
+                imageSource: 'imageSource',
+                options: {
+                    maxWidth: 1000,
+                    minWidth: 400,
+                    widthAspectRatio: 5,
+                    heightAspectRatio: 3
+                }
+            }
+        },
+        type: 'text-image'
     },
     {
         key: 'imageSrcWidth',
--- a/public/src/js/constants/article-meta-fields.js
+++ b/public/src/js/constants/article-meta-fields.js
@@ -72,11 +72,13 @@ export default Object.freeze([
         label: 'replacement image URL',
         validator: {
             params: {
+                // transition only - inherited from imageSrc
                 src: 'imageSrc',
                 width: 'imageSrcWidth',
                 height: 'imageSrcHeight',
                 origin: 'imageSrcOrigin',
-                imageSource: 'imageSource',
+                // when transition is complete this is the only one needed
+                source: 'imageSource',
                 options: {
                     maxWidth: 1000,
                     minWidth: 400,
--- a/public/src/js/models/article/editor.js
+++ b/public/src/js/models/article/editor.js
@@ -56,11 +56,10 @@ export default class Editor extends BaseClass {
             if (validator && _.isFunction(this[validator.fn])) {
                 this.subscribeOn(meta, () => {
                     if (this.performValidation !== false) {
-                        this[validator.fn](validator.params, meta);
+                        this[validator.fn](validator.params);
                     }
                 });
             }
-
         } else if (type === 'text-image') {
             this.updateText = ko.pureComputed({
                 read: this.imageValue,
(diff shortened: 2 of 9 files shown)


public/src/js/constants/article-meta-fields.js
```diff
@@ -64,6 +64,30 @@ export default Object.freeze([
...
+        label: 'replacement image URL',
+        validator: {
+            params: {
+                src: 'imageSrc',
+                width: 'imageSrcWidth',
+                height: 'imageSrcHeight',
+                origin: 'imageSrcOrigin',
+                imageSource: 'imageSource',
+                options: {
+                    maxWidth: 1000,
```
paperboyo: This wouldn't have anything to do with https://github.com/guardian/frontend/issues/12214#issuecomment-197263238? Just asking :blush:

piuccio: 😆  it's exactly it

public/src/js/models/article/editor.js
```diff
@@ -56,16 +56,24 @@ export default class Editor extends BaseClass {
             return opts.maxLength && this.value().length > opts.maxLength;
         });
         this.displayEditor = ko.pureComputed(this.isVisible, this);
-        this.updateText = ko.pureComputed({
-            read: this.value,
-            write: this.writeText,
-            owner: this
-        });
+        if (type === 'text') {
```
piuccio: can you group all the `if (type ===` together after line 49, all the type dependent logic should be in the same place

public/src/js/models/article/editor.js
```diff
@@ -56,16 +56,24 @@ export default class Editor extends BaseClass {
...
+            });
+        } else if (type === 'text-image') {
+            this.updateText = ko.pureComputed({
+                read: this.imageValue,
+                write: this.writeImage,
+                owner: this
+            });
+        }
 
-        if (validator && _.isFunction(this[validator.fn])) {
```
piuccio: We need to keep both, `validator && _.isFunction(this[validator.fn])` is still used by cutout images

public/src/js/models/article/editor.js
```diff
@@ -75,6 +83,13 @@ export default class Editor extends BaseClass {
         return this.meta() || this.field() || '';
     }
 
+    imageValue() {
+        if (this.meta() && this.meta().src) {
```
piuccio: cache `this.meta()` in a `const` to avoid calling the function 3 times

public/src/js/constants/article-meta-fields.js
paperboyo: But changing it to pick the biggest wouldn't automatically choose correctly `/master/` (there is another crappier crop of same width next to `/master/`, we should choose it only if the `/master/` is not there - like with old assets). Sorry if preaching to the choir!

BTW - now I know why I have to have GIFs over 400px wide while trying to satisfy Grant's 100KB size limit (a line below). Not that we want more GIFs. Or JPGs below 400px wide... I don't!

public/src/js/models/article/editor.js
```diff
@@ -146,22 +168,66 @@ export default class Editor extends BaseClass {
         this.meta(value === this.field() ? undefined : value.replace(rxScriptStriper, ''));
     }
 
+    writeImage(value) {
+        var checkedValue = this.field() === value ? undefined : value.replace(rxScriptStriper, '');
+        this.meta({
```
piuccio: Because this is the only place where we update the `meta` we can simplify the logic. Instead of doing `this.meta` and subscribing to the value change, we can call `validateImage` directly.

We can get rid of a subscription and we never store a possibly invalid value in the meta.

public/src/js/models/article/editor.js
```diff
@@ -146,22 +168,66 @@ export default class Editor extends BaseClass {
...
+                this.assignImageToObjectElement(params, null);
+            }
+        }
+    }
+
+    assignImageToObjectElement(params, imgOrError, origin) {
+
+        if (!imgOrError || imgOrError instanceof Error) {
+
+            assign(this, [this.article.meta.imageSource], [{
```
piuccio: instead of assigning an object with empty values, it's better assigning undefined

public/src/js/models/article/images.js
```diff
@@ -19,8 +19,12 @@ export function main() {
         fields = this.fields,
         state = this.state;
 
-    if (meta.imageReplace() && meta.imageSrc()) {
-        return meta.imageSrc();
+    if (meta.imageReplace()) {
+        if (meta.imageSource()) {
+            return meta.imageSource().origin;
```
piuccio: `meta.imageSource().src`

public/src/js/widgets/trail-editor.html
```diff
@@ -27,6 +27,28 @@
...
+        <div class="editor" data-bind="css: {'length--alert': lengthAlert}">
+            <textarea data-bind="
+                click: open,
+                clickBubble: false,
+                hasFocus: hasFocus,
+                attr: {
+                    class: 'element__' + key + (hasFocus() ? ' active' : ''),
+                    placeholder: field() || (label + '...'),
+                    title: label,
+                    href: meta() ? meta().origin || meta().src : ''
```
piuccio: only `meta().src`

public/src/js/models/collections/article.js
```diff
@@ -309,7 +309,7 @@ export default class Article extends DropTarget {
             mediator.emit(
                 'ui:open',
                 _.chain(this.editors())
-                 .filter(function(editor) { return editor.type === 'text' && editor.displayEditor(); })
+                 .filter(function(editor) { return (editor.type === 'text' || editor.type === 'text-image') && editor.displayEditor(); })
```
piuccio: how about `editor.indexOf('text') === 0`? We won't have to change it if we add more text based editors

public/src/js/constants/article-meta-fields.js
piuccio: Yes, with this change we're simply getting ready to select a different image. Neither of us has figured out yet how to choose the master. We'll keep it for next PR

paperboyo: Sure thing! Frontend uses CAPI's [`isMaster` field](https://github.com/guardian/frontend/pull/12254/files#diff-e349969fa55f6fac9fb629179b58be55R33)... Grid and cropper emits an element named `master` as opposed to other `assets`. But I wouldn't know. @kenoir would, though. He's the `/master/`'s master.

piuccio:
I've pushed a small update to this. However image replacements are not displayed on pressed pages even with facia press on latest version.

piuccio:
Closing this, the changes is scala client are not feasible. iOS needs to know the image aspect ration

PR #169, opened 2016-04-13
sanitize slashes in capi query

piuccio:
@davidfurey 

This prevents new `/search` from being added. I'll have to remove the ones already in.


davidfurey:
Thanks Fabio!

PR #195, opened 2016-06-17
Don't send HTML entities to the notification client

piuccio:
@davidfurey


davidfurey:
The headline from CAPI doesn't isn't html encoded, so presumably you are encoding it somewhere else and then decoding it again here?

👍

piuccio:
true, CAPI is not encoded, but if you write a custom headline, the browser converts `&` into `&amp;` when writing the headline into a textarea.

I don't trust JS to fix encoding issues, so I unescape in the backend.

PR #205, opened 2016-09-12
Display front tags

Reettaphant:
This pr makes front collection tags more visible in the tool. 

- It displays collection tags on the front configuration page without having to click the collection open: 

screen shot 2016-09-12 at 08 15 16

- It displays the tags on the front edit page without having to click through to the configuration page: 

screen shot 2016-09-12 at 08 14 52


clloyd:
👍

PR #206, opened 2016-09-16
Image drag validation

Reettaphant:
- Avoid reloading an image that has crop information when dragging it to a front to ensure faster loading.


Reettaphant:
@piuccio

PR #210, opened 2016-09-26
Do not allow http image and use https domains

mchv:
- We had some fronts embedding `http` images which then prevent the front to be displayed as secure, so do not allow to add `http` image

- Always use our `https` domains

### DIFF
--- a/public/src/js/constants/defaults.js
+++ b/public/src/js/constants/defaults.js
@@ -3,8 +3,8 @@ export default {
     editions: ['uk', 'us', 'au'],
 
     environmentUrlBase: {
-        'prod': 'http://theguardian.com/',
-        'code': 'http://m.code.dev-theguardian.com/'
+        'prod': 'https://theguardian.com/',
+        'code': 'https://m.code.dev-theguardian.com/'
     },
 
     types: [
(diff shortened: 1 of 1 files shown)


public/src/js/constants/defaults.js
```diff
@@ -209,10 +209,10 @@ export default {
     reauthTimeout:         60000,
 
     imageCdnDomain:        '.guim.co.uk',
-    imageCdnDomainExpr:    /^https?:\/\/(.*)\.guim\.co\.uk\//,
-    imgIXDomainExpr:       /^https?:\/\/i\.guim\.co\.uk\/img\/static\//,
+    imageCdnDomainExpr:    /^https:\/\/(.*)\.guim\.co\.uk\//,
```
piuccio: Instead of changing this regular expression I think it's better to automatically replace `http:` with `https:` in `validate-image`

public/src/js/constants/defaults.js
```diff
@@ -209,10 +209,10 @@ export default {
...
+    imgIXDomainExpr:       /^https:\/\/i\.guim\.co\.uk\/img\/static\//,
```
piuccio: There's no need to change this expression because the code rewrite the URL to include https already

public/src/js/constants/defaults.js
mchv: I think not allowing `http` is better as it less ambiguous

mchv:
@jennysivapalan @philmcmahon @susiecoleman @Reettaphant are you happy with those changes? could we quickly check how many collection there is that currently references `http` images?

PR #211, opened 2016-09-27
readme cleanup

Reettaphant:
Removes mention of frontend credentials and editing of credentials file as this now comes from janus, and clarifies how tests are run. 

@susiecoleman

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,8 @@ $ feria cmsFronts && feria --access s3-read workflow && feria --access sqs-consu
 ```
 (You should first checkout the repository and follow the install instructions for *feria*.)
 
-
+You can run the fronts tool without frontend credentials, but you will not be able to check how your changes to fronts appear on frontend without
+these credentials. You will need to test your changes on `CODE` to see these changes.
 ### Code Dependencies
 
 Inside the project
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -173,7 +163,7 @@ You need valid developer credentials for `cmsFronts` and `workflow`.
 You can get keys temporary keys from `janus`. Note that the recommended way to get Janus credentials for *facia-tool* is to use feria and run:
 
 ```
-$ feria cmsFronts && feria --access s3-read workflow && feria --access sqs-consumer frontend
```
piuccio: I think the sqs is still useful, but the app probably work without

Reettaphant: I've added it back, but I'll add a note about the frontend credentials after the tool is no longer writing on the frontend bucket.

piuccio: Even when we don't write anymore to the other bucket we still need to send messages to sqs queue to let facia press know that a change has happened.

Reettaphant: yes, so i'll just clarify what people can and cannot do without frontend credentials so that people can work on facia tool without having them.

Reettaphant:
@clloyd Do these changes to the readme make sense?

PR #212, opened 2016-09-28
wait before testing for update

Reettaphant:
The tests to update front config sometimes fails on Travis: https://travis-ci.org/guardian/facia-tool/builds/162715143

I think this may fix it. 

@piuccio


piuccio:
Thanks for looking into this, it's rather annoying indeed 

:+1:

PR #218, opened 2016-12-07
Update FAPI client

gidsg:


build.sbt
```diff
@@ -73,14 +75,15 @@ libraryDependencies ++= Seq(
...
-    "com.gu" %% "fapi-client" % "2.0.1",
+    "com.gu" %% "fapi-client" % "2.0.6",
     "com.gu" % "kinesis-logback-appender" % "1.3.0",
     "com.gu" %% "mobile-notifications-client-play-2-4" % "0.5.29",
     "com.gu" %% "pan-domain-auth-play_2-4-0" % "0.3.0",
     "com.gu" %% "scanamo" % "0.6.0",
     "com.gu" %% "thrift-serializer" % "1.1.0",
     "net.logstash.logback" % "logstash-logback-encoder" % "4.7",
     "org.julienrf" %% "play-json-variants" % "2.0",
+    "org.json4s" % "json4s-native_2.11" % "3.5.0",
```
gidsg: this is required as json4s is no longer bundled with the content-api client.

gidsg:
Do Not Merge this as we are investigating some issues with transitive dependencies

piuccio:
have fun with transitive dependencies 😆 

Keep in mind that `fapi-client` itself depends on capi models, things work better when they're on the same version.

gidsg:
This is blocked as `content-api-models` depends on an old version of Circe which then conflicts with the version of Cats required by Scanamo. So I am going to close this until we can resolve the issues upstream.

PR #222, opened 2016-12-22
Email front layouts

joelochlann:
Email fronts now have these layouts:

- Fast
- Medium
- Slow
- Most Popular

rather than the huge list of layouts that are only applicable to web.

Currently it looks a bit weird because for some reason its width shrinks right down when it has fewer than six items in it. In a future PR @superfrank is going to supply some SVGs for thumbnails and I'll try and sort out the width issue then:

picture 374

@Reettaphant


joelochlann:
@philmcmahon

philmcmahon:
👍 
email gif

PR #223, opened 2016-12-22
Allow unhiding of email fronts

joelochlann:
For some reason `isTypeLocked` also needs to be false to allow me to unhide an email front

@philmcmahon


philmcmahon:
👍

PR #225, opened 2017-02-07
Add backfill limit for email fronts

davidfurey:
Allows email fronts to use backfill but still restrict the maximum number of articles in a collection.  Depends on https://github.com/guardian/facia-scala-client/pull/182, and will require a change to frontend to respect the new field.

screen shot 2017-02-10 at 14 04 13

### DIFF
--- a/public/src/js/utils/observable-numeric.js
+++ b/public/src/js/utils/observable-numeric.js
@@ -3,12 +3,10 @@ import ko from 'knockout';
 export default function(initialValue) {
     var actual = ko.observable(initialValue);
     return ko.dependentObservable({
-        read: function ()
-        {
+        read() {
             return actual();
         },
-        write: function (newValue)
-        {
+        write(newValue) {
             var parsedValue = parseFloat(newValue);
             actual(isNaN(parsedValue) ? newValue : parsedValue);
         }
--- a/public/src/js/utils/observable-numeric.js
+++ b/public/src/js/utils/observable-numeric.js
@@ -1,0 +1,16 @@
+import ko from 'knockout';
+
+export default function(initialValue) {
+    var actual = ko.observable(initialValue);
+    return ko.dependentObservable({
+        read: function ()
+        {
+            return actual();
+        },
+        write: function (newValue)
+        {
+            var parsedValue = parseFloat(newValue);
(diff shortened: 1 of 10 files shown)


public/src/js/utils/observable-numeric.js
```diff
@@ -0,0 +1,16 @@
...
+    var actual = ko.observable(initialValue);
+    return ko.dependentObservable({
+        read: function ()
+        {
+            return actual();
+        },
+        write: function (newValue)
+        {
+            var parsedValue = parseFloat(newValue);
+            actual(isNaN(parsedValue) ? newValue : parsedValue);
```
joelochlann: so if we can't turn it into a number, we pass along the string and allow the backend to break?

public/src/js/widgets/columns/fronts-config.html
```diff
@@ -232,6 +232,11 @@
                 <span class="cnf-form__value" data-bind="text: meta.groups"></span>
             <!-- /ko -->
 
+            <div data-bind="visible: $root.priority == 'email' ">
+            <label for="backfillLimit">Maximum number of items to render</label>
```
joelochlann: should be indented within the `<div></div>`

public/src/js/widgets/columns/fronts-config.html
```diff
@@ -232,6 +232,11 @@
...
+            <input id="backfillLimit" type="number" data-bind="value: meta.displayHints.maxItemsToDisplay">
```
joelochlann: Also you may need to add a styling rule for inputs of type "number": https://github.com/guardian/facia-tool/blob/master/public/src/css/style.css#L1157-L1186

public/src/js/utils/observable-numeric.js
```diff
@@ -0,0 +1,16 @@
+import ko from 'knockout';
+
+export default function(initialValue) {
+    var actual = ko.observable(initialValue);
+    return ko.dependentObservable({
+        read: function ()
+        {
```
joelochlann: I'd always keep this opening `{` on the same line as `function ()`

joelochlann: oh, actually also in ES6 you can just do `read()` as shorthand for `read: function()`

PR #227, opened 2017-03-15
Update kinesis stream

jennysivapalan:
Uses new kinesis role (cloudformed in ELK stack) and updates the parameters in logging to match EC2 instance tags.

### DIFF
--- a/app/logging/LogStash.scala
+++ b/app/logging/LogStash.scala
@@ -62,9 +62,9 @@ object LogStash {
         lb.info("Configuring Logback")
         val context = lb.getLoggerContext
         val customFields = Map(
-          "stack" -> "fronts",
+          "stack" -> "cms-fronts",
           "stage" -> config.environment.stage.toUpperCase,
-          "app"   -> config.logging.app
+          "app"   -> config.environment.applicationName
         )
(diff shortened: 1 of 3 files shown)


app/logging/LogStash.scala
```diff
@@ -62,9 +62,9 @@ object LogStash {
         lb.info("Configuring Logback")
         val context = lb.getLoggerContext
         val customFields = Map(
-          "stack" -> "fronts",
+          "stack" -> "cms-fronts",
           "stage" -> config.environment.stage.toUpperCase,
-          "app"   -> config.logging.app
+          "app"   -> config.environment.applicationName
```
susiecoleman: Nothing in this pr changed the `ApplicationConfiguration` but I'm going to assume that `environment.applicationName` already exists...

jennysivapalan: It does

jennysivapalan:
So this didn't work as there is a dependency conflict on Fronts tools. The error we get is same stack trace to this http://stackoverflow.com/questions/41689194/kinesis-stream-putrecord-failing-unable-to-marshall-request-to-json.
I will raise a health task for addressing this.

PR #229, opened 2017-04-06
Add free-text layout type for email

joelochlann:
Adds a free text layout type to be used for containers which just contain a message from the editor, typically used at the top or bottom or an email.

For now, editorial will drop a card into this container and write all their text in the headline field (sadly, using `<br>` tags for newlines. Hopefully at some point this can be replaced with real rich text editor in the fronts tool).

Frontend will know to render this card differently because it's in a container with "free text" layout type.

I also removed the `most-popular` layout since we don't use it

### DIFF
--- a/public/src/js/constants/defaults.js
+++ b/public/src/js/constants/defaults.js
@@ -66,7 +66,7 @@ export default {
         { name: 'fast-images' },
         { name: 'medium' },
         { name: 'slow' },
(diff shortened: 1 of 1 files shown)


joelochlann:
A free text container may look like this when rendered:

 
email_sleeve_notes_text_box_sans_serif-top 1

PR #230, opened 2017-04-12
Upgrade panda lib and add check for user groups

marialivia16:
Adds user group validation and uses `cacheValidation` to avoid calling `validateUser` with every request.

Tested on CODE.

### DIFF
--- a/app/auth/PanDomainAuthActions.scala
+++ b/app/auth/PanDomainAuthActions.scala
@@ -4,6 +4,7 @@ import com.amazonaws.auth.profile.ProfileCredentialsProvider
 import com.amazonaws.auth.{AWSCredentialsProviderChain, STSAssumeRoleSessionCredentialsProvider}
 import com.gu.pandomainauth.action.AuthActions
 import com.gu.pandomainauth.model.AuthenticatedUser
+import com.gu.pandomainauth.service.GoogleGroupChecker
 import com.gu.pandomainauth.{PanDomain, PanDomainAuth}
 import conf.ApplicationConfiguration
 import play.api.Logger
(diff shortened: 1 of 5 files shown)


marialivia16:
[Fronts tool should use multiple google groups as per Composer](https://trello.com/c/EBC5zgFB/71-fronts-tool-should-use-multiple-google-groups-as-per-composer)

app/auth/PanDomainAuthActions.scala
```diff
@@ -26,7 +34,10 @@ trait PanDomainAuthActions extends AuthActions with PanDomainAuth with Results {
   override def invalidUserMessage(claimedAuth: AuthenticatedUser): String = {
     if( (claimedAuth.user.emailDomain == "guardian.co.uk") && !claimedAuth.multiFactor) {
       s"${claimedAuth.user.email} is not valid for use with the Fronts Tool as you need to have two factor authentication enabled." +
-       s" Please contact the Helpdesk by emailing 34444@theguardian.com or calling 34444 and request access to Composer CMS tools."
+      s" Please contact the Helpdesk by emailing 34444@theguardian.com or calling 34444 and request access to Composer CMS tools."
+    } else if (!userInGroups(claimedAuth)) {
+      s"${claimedAuth.user.email} is not valid for use with the Fronts Tool as you need to be granted access to the editorial tools." +
```
akash1810: minor: could use [multiline string](https://www.safaribooksonline.com/library/view/scala-cookbook/9781449340292/ch01s03.html)

build.sbt
```diff
@@ -56,7 +56,7 @@ TwirlKeys.templateImports ++= Seq(
     "play.api.Play.current"
 )
 
-val awsVersion = "1.11.18"
+val awsVersion = "1.11.8"
```
akash1810: 18 -> 8? Why are we making this change?

marialivia16: It was initially because of the deploy issue I had, but found the problem in the config file. I'll take this out to see if it's still working.

marialivia16:
Looking at other ways to implement this - trying not to rely on Google so much.

PR #231, opened 2017-04-27
Upgrade facia scala client

marialivia16:
Upgrade needed to label paid content as `paid` instead of `news`.


LATaylor-guardian:
👍

PR #233, opened 2017-05-16
New dynamo/election container layout

DiegoVazquezNanini:
In line with the change in the election container design, we are changing the implementation.
Apologies for the noise.
@NataliaLKB @Reettaphant

### DIFF
--- a/app/slices/Container.scala
+++ b/app/slices/Container.scala
@@ -1,7 +1,6 @@
 package slices
 
 import com.gu.facia.api.models.CollectionConfig
-import conf.ApplicationConfiguration
 import play.api.Logger
 
 class Containers(val fixedContainers: FixedContainers) {
@@ -10,6 +9,7 @@ class Containers(val fixedContainers: FixedContainers) {
     ("dynamic/fast", Dynamic(DynamicFast)),
     ("dynamic/slow", Dynamic(DynamicSlow)),
     ("dynamic/package", Dynamic(DynamicPackage)),
+    ("dynamo/election", Dynamic(DynamicElection)),
     ("dynamic/slow-mpu", Dynamic(DynamicSlowMPU)),
     ("nav/list", NavList),
     ("nav/media-list", NavMediaList),
--- a/app/slices/Container.scala
+++ b/app/slices/Container.scala
@@ -9,7 +9,7 @@ class Containers(val fixedContainers: FixedContainers) {
     ("dynamic/fast", Dynamic(DynamicFast)),
     ("dynamic/slow", Dynamic(DynamicSlow)),
     ("dynamic/package", Dynamic(DynamicPackage)),
-    ("dynamo/election", Dynamic(DynamicElection)),
+    ("dynamic/election", Dynamic(DynamicElection)),
     ("dynamic/slow-mpu", Dynamic(DynamicSlowMPU)),
     ("nav/list", NavList),
     ("nav/media-list", NavMediaList),
--- a/app/slices/DynamicElection.scala
+++ b/app/slices/DynamicElection.scala
@@ -1,0 +1,30 @@
+package slices
+
+import slices.Story._
+
+object DynamicElection extends DynamicContainer {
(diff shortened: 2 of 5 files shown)


NataliaLKB:
@DiegoVazquezNanini thanks for this! Looks good to me, but I will let @Reettaphant do the proper review.
This change will also add:
image

to the top of the container?

app/slices/Container.scala
```diff
@@ -10,6 +9,7 @@ class Containers(val fixedContainers: FixedContainers) {
     ("dynamic/fast", Dynamic(DynamicFast)),
     ("dynamic/slow", Dynamic(DynamicSlow)),
     ("dynamic/package", Dynamic(DynamicPackage)),
+    ("dynamo/election", Dynamic(DynamicElection)),
```
NataliaLKB: Should this be dynamic instead of dynamo?

stephanfowler:
Line 61 in the [container definitions file](https://github.com/guardian/facia-tool/blob/46332be525ebe0f231ced894fb32661b73f6a610/public/src/js/constants/defaults.js) needs to look like this:
```
{
   'name': 'dynamo/election',
   'groups': [
   'standard',
   'snap'
   ]
},
```
... otherwise the groups aren't offered in the Fronts Tool.

DiegoVazquezNanini:
Thanks for the review. @NataliaLKB and @stephanfowler

app/slices/DynamicElection.scala
```diff
@@ -0,0 +1,30 @@
...
+    // optional snap (1) - 2 - 4
+    stories.length match {
+      case 0 => Nil
+      case 1 => Nil
+      case 2 => Seq(ThreeQuarterTallQuarter1)
+      case 3 => Seq(ThreeQuarterTallQuarter1)
+      case 4 => Seq(ThreeQuarterTallQuarter1)
+      case 5 => Seq(ThreeQuarterTallQuarter1)
+      case 6 => Seq(ThreeQuarterTallQuarter1, QuarterQuarterQuarterQuarter)
+      case _ => Seq(ThreeQuarterTallQuarter1, QuarterQuarterQuarterQuarter)
```
TBonnin: Something like this would be more concise. Not sure about the readability though 😝 
```
stories.length match {
       case c if 0 until 1 contains c => Nil
       case c if 2 until 5 contains c => Seq(ThreeQuarterTallQuarter1)
       case _ => Seq(ThreeQuarterTallQuarter1, QuarterQuarterQuarterQuarter)
}
```

app/slices/DynamicElection.scala
```diff
@@ -0,0 +1,30 @@
+package slices
+
+import slices.Story._
+
+object DynamicElection extends DynamicContainer {
+  override protected def optionalFirstSlice(stories: Seq[Story]): Option[(Slice, Seq[Story])] = {
+    val byGroup = segmentByGroup(stories)
+    val snap = byGroup.getOrElse(1, Seq.empty)
```
TBonnin: Types of those vals would be nice to have.

app/slices/DynamicElection.scala
NataliaLKB: Personally I like how explicit it currently is 😄

app/slices/DynamicElection.scala
```diff
@@ -0,0 +1,30 @@
...
+
+    if (snap.nonEmpty) {
+      Some((FullMedia50, stories.drop(1)))
+    } else {
+      None
+    }
```
TBonnin: I think the same logic can be done this way.
```
segmentByGroup(stories)
      .get(1)
      .filter(_.nonEmpty)
      .map(stories => (FullMedia50, stories.drop(1)))
```

app/slices/DynamicElection.scala
DiegoVazquezNanini: @TBonnin I decided not to go with that because of readability, but that is a good comment.
Also, remember that we plan to delete this code as soon as the election finish.

PR #236, opened 2017-05-30
add viewUrl to snap links

shaundillon:
snap links were missing a viewUrl due to a refactor in https://github.com/guardian/facia-tool/commit/f736976083867b6b40bec4959ad512601b500111 missing them out.

/cc @Reettaphant

### DIFF
--- a/public/src/js/models/collections/article.js
+++ b/public/src/js/models/collections/article.js
@@ -138,12 +138,6 @@ export default class Article extends DropTarget {
             contentApi.decorateItems(this.meta.supporting.items());
         }
 
-        // Add viewUrl to snap links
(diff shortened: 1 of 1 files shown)


public/src/js/models/collections/article.js
```diff
@@ -137,6 +138,12 @@ export default class Article extends DropTarget {
             contentApi.decorateItems(this.meta.supporting.items());
         }
 
+        // Add viewUrl to snap links
+        const viewUrl = getViewUrl(this);
+        if (viewUrl) {
+            this.state.viewUrl = viewUrl;
+        }
+
```
Reettaphant: Can do this only when we are doing to conversion to snaps?

PR #238, opened 2017-06-01
update build to use riff-raff.yaml

shaundillon:


jennysivapalan:
You can remove the `deploy.json` too. 
Should we move this build from travis to Teamcity?

shaundillon:
@jennysivapalan - yeah, that would make sense. Perhaps in a different PR though! Also, I need to fix this one first haha.

PR #242, opened 2017-06-12
Removing the 2017 election container

NataliaLKB:
Reverting: https://github.com/guardian/facia-tool/pull/232 and https://github.com/guardian/facia-tool/pull/233

To go with corresponding change to dotcom: https://github.com/guardian/frontend/pull/17098

The election container was created just for the election which has now passed. Removing the code.

_Note_: There has been talk of elements of this container to be reimplemented in the future. Until that is decided though its best to revert this change to avoid dead code lying around.


NataliaLKB:
Hey @Reettaphant this morning I took a look and unfortunately I am not sure if this has worked. The `dynamo/election` container is still an option. What have I forgotten? :thinking: 

image

PR #247, opened 2017-06-30
Use small thumbnail for replacement image thumbnails

gtrufitt:
We now request the smallest crop of a **replacement** image from the grid and add it to the meta of a collection. This means we are loading 140px wide images for replaced images instead of the master crops (maybe 5000px wide). 

For articles that have images that haven't been replaced, we just use the existing thumb from CAPI which is 500px wider or so - maybe we want to reduce the size of this?

This PR doesn't currently handle when a user drops an image in - it will still load the master crop at that point.

- [x] Do we want to reduce the size of default image crop?
- [x] How do we use the small crop when dropping in?


paperboyo:
> Do we want to reduce the size of default image crop?

Back-of-an-envelope numbers [view of UK front at 23:39, 30.06.2017]:

| what | count | weight |
| ------ | ------- | ------------- |
|  500.jpg/png Grid thumbs | 197 | 5.7MB |
| static cutout PNGs | 48 | 5.2MB  |
| masters  | **1 (sic!)** |3MB  |
|Whole page | | 15.2MB|

Replacing 500px Grid thumbs with 140px ones: 5.7MB > 1.4MB.

Since all trails are shown at 83px wide at 100% zoom in frontstool, loading 140px should still satisfy a fair amount of zooming in (or almost a 2x HiDPI screen). So it should be safe (and lighter!) to switch to 140px.

Static cutout trails are impossible to fix, but cutouts from Grid should use the same logic to normal Grid images (they will also have the same array of thumbnails). And one day – they will prevail over static. Currently, cutouts from Grid use a `1000.png` Grid thumbnail (instead of much smaller `500.png`, or `140.png`). **Frontsool also incorrectly sends that `1000.png` Grid thumb to frontends (instead of master asset!).**

Slideshow trails should also use the same logic (at least these are sent to frontends as masters).

> How do we use the small crop when dropping in?

No idea. There is something attached to a mouse pointer while dragging – can we use this (before loading Grid thumb proper)?

app/controllers/FaciaToolController.scala
```diff
@@ -32,10 +33,19 @@ class FaciaToolController(val config: ApplicationConfiguration, val acl: Acl, va
...
+        case Some(f) => f.map(Some(_))
+        case None    => Future.successful(None)
+      }
+
     FaciaToolMetrics.ApiUsageCount.increment()
-    frontsApi.amazonClient.collection(collectionId).map { configJson =>
-      NoCache {
-        Ok(Json.toJson(configJson)).as("application/json")}}}
+    val collection = frontsApi.amazonClient.collection(collectionId).flatMap{ collectionJson =>
+      f(collectionJson.map(mediaServiceClient.addThumbnailsToCollection))
```
mbarton: Could we just inline the helper function here?

akash1810:
> > How do we use the small crop when dropping in?

> No idea. There is something attached to a mouse pointer while dragging – can we use this (before loading Grid thumb proper)?

Grid adds its complete api response to the drag event if that's helpful

piuccio:
I think the drag event already uses the thumbnail, if not you might find [this](https://github.com/guardian/grid-util-js/blob/master/src/index.js#L98) useful

paperboyo:
👋 :jp: 🗻[This!](http://www.imdb.com/title/tt0084390/) 🏯 :jp: 👋

jfsoul:
For the record this doesn't affect what sizes get passed to frontend?

app/controllers/FaciaToolController.scala
```diff
@@ -32,10 +33,19 @@ class FaciaToolController(val config: ApplicationConfiguration, val acl: Acl, va
         Ok(Json.toJson(configJson)).as("application/json")}}}
 
   def getCollection(collectionId: String) = APIAuthAction.async { request =>
+    def f[A](x: Option[Future[A]])(implicit ec: ExecutionContext): Future[Option[A]] =
+      x match {
+        case Some(f) => f.map(Some(_))
+        case None    => Future.successful(None)
+      }
```
jfsoul: In a few places I think you could replace this pattern match with 'Option.fold' which might be neater.

gtrufitt:
> For the record this doesn't affect what sizes get passed to frontend? 

Nope just for rendering inside the tool, doesn't update any field that Frontend uses.

Thx @akash1810 & @piuccio - I'll take a look at the event.

app/controllers/FaciaToolController.scala
gtrufitt: "What would Odersky do?" 😛

https://www.reddit.com/r/scala/comments/449td4/fold_or_pattern_matching_with_an_option/czp154s/

gtrufitt:
Finally got a chance to look at this again! I now use the thumb image that is a part of the event and add it as the meta property that it's also assigned to on the back end. That means that dropping in will also use the much smaller image size as well as on page load. 

@akash1810 Any chance of another gander?

paperboyo:
💎 This should allow to fix a last thing (not directly related) – to correctly pass a master asset for cutouts from Grid (currently, a 1000px Grid crops are passed to frontends: see a cake on training-site-2).

app/services/MediaServiceClient.scala
```diff
@@ -0,0 +1,59 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class MediaServiceClient(val mediaApi: MediaApi) {
+
+  def getThumbnail(imageSrc: String):Future[Option[String]] = {
+    val imageIdRegex = "[0-9a-z]{40}".r
+
+    val thumbnail = for {
+      imageId <- imageIdRegex findFirstIn imageSrc
+      cropIdRegex = s"\\/$imageId\\/([^\\/]+)".r
```
akash1810: 🤢

app/services/MediaApi.scala
```diff
@@ -0,0 +1,40 @@
...
+    val gridUrl = s"${config.media.apiUrl}/images/$imgId"
+
+    ws.url(gridUrl).withHeaders("X-Gu-Media-Key" -> config.media.key).get().map(
+      response => extractExportsFromJsonResponse(response.json, cropId))
+  }
+
+  def extractExportsFromJsonResponse(response: JsValue, cropId: String) = {
+    val exports = (response \ "data" \ "exports").as[List[Export]]
+    exports.find(_.id == cropId) match {
+      case Some(export) => Some(export.assets.sortBy(asset => asset.size).head.secureUrl)
```
akash1810: are we sure we want the smallest crop asset? It can be as small as [140px wide](https://github.com/guardian/grid/blob/Tidy_s3watcher-lambda/cropper/app/lib/Config.scala#L40)

public/src/js/models/article/editor.js
```diff
@@ -169,17 +169,18 @@ export default class Editor extends BaseClass {
         var imageSrc = this.article.meta[params.src],
             imageSrcWidth = this.article.meta[params.width],
             imageSrcHeight = this.article.meta[params.height],
-            imageSrcOrigin = this.article.meta[params.origin];
+            imageSrcOrigin = this.article.meta[params.origin],
+            imageSrcThumb = this.article.meta[params.imageSrcThumb] || function(){return '';};
```
akash1810: why is this a function?

public/src/js/models/article/images.js
```diff
@@ -3,7 +3,9 @@ export function thumbnail() {
         fields = this.fields,
         state = this.state;
 
-    if (meta.imageReplace() && meta.imageSrc()) {
+    if (meta.imageReplace() && meta.imageSrcThumb() && meta.imageSrcThumb() !== ''){
```
akash1810: minor: could create a `const` for `meta.imageSrcThumb()` to avoid multiple evaluations

app/services/MediaApi.scala
paperboyo: If I’m correct ☝️ (and that’s a uuge if), this should be fine if we are talking abouts frontstool small thumbs in current design:

> Since all trails are shown at 83px wide at 100% zoom in frontstool, loading 140px should still satisfy a fair amount of zooming in (or almost a 2x HiDPI screen). So it should be safe (and lighter!) to switch to 140px.

public/src/js/models/article/editor.js
gtrufitt: It's expected to be a Knockout observable I believe, so it's called to get a value

gtrufitt:
Holy moly. Finally deploying live.

PR #248, opened 2017-07-25
ensure deploys can deploy static files too

shaundillon:


franziskas:
@shaundillon can we salvage this PR?

franziskas:
Chatted with Shaun, closing this.
> I think we should scrap it, it's pretty stale and facia is deploying fine as is

PR #249, opened 2017-08-01
Adding a warning to shared containers on fronts

rcrphillips:
BEFORE
screen shot 2017-08-01 at 15 34 40

AFTER
screen shot 2017-08-01 at 15 32 00

Some fronts editors have mentioned that other users often edit shared containers, affecting them in multiple places. This should provide a quick mental check that they're approaching the task in the right way.

### DIFF
--- a/public/src/css/style.css
+++ b/public/src/css/style.css
@@ -2042,6 +2042,11 @@ ol.search_list > li::before {
     color: #555;
 }
 
+span.also a{
+  color: red;
+  font-size: 20px;
+}
+
 .also {
     margin: 6px;
     line-height: 24px;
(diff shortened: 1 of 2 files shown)


public/src/css/style.css
```diff
@@ -2042,6 +2042,11 @@ ol.search_list > li::before {
     color: #555;
 }
 
+span.also a{
```
piuccio: usually it's better to avoid mentioning the tag name, `.also` is more future-proof than`span.also` and less verbose

rcrphillips: Thanks @piuccio Further down in the collection html there are other links which picked up the red styling, so tried to add the specificity to lear it up.

PR #250, opened 2017-08-03
alert if more than 20 stories are added to a collection

shaundillon:
picture 257


TBonnin:
I think there is already a feature to limit the number of stories in a container (with a current limit at 35)

screen shot 2017-08-03 at 15 14 57

shaundillon:
@TBonnin Ahh, I'm unfamiliar with the tool so I didn't know. Thanks!

PR #251, opened 2017-08-03
Change collection cap

shaundillon:
Far easier than I thought

### DIFF
--- a/app/conf/Configuration.scala
+++ b/app/conf/Configuration.scala
@@ -129,7 +129,7 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
 
   object facia {
     lazy val stage = getString("facia.stage").getOrElse(stageFromProperties)
-    lazy val collectionCap: Int = 35
+    lazy val collectionCap: Int = 20
   }
 
   object logging {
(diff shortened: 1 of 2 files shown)


TBonnin:
lgtm.

What happens if the container already contains more than 20 stories and one tries to add an additional one. Does it drop everything below 20?

Also, what do we do with containers that currently have more than 20 stories. Would they need to be edited for this limit to take enacted? Or should we truncate them?

rcrphillips:
@TBonnin @shaundillon we can probably test those negative paths on code quite easily. I'll deploy it now as I've set a container ready with >20

shaundillon:
@TBonnin - As far as the client is concerned, the alert just shows if a new item is added and the number of articles exceeds the collection cap. It would not remove any existing stories from there.

That said, inside https://github.com/guardian/facia-tool/blob/c7e43eb5e6b53b4e13ad0936d7c7dddebeb60379/app/model/frontsapi.scala#L142 When collections are updated they're capped.

I'm not sure when this is called, but it looks like it definitely would affect any collections that currently have more than 35 stories in them.

That said, it seems excessive to have more than that

TBonnin:
Great. That answers my question. Thx @shaundillon

PR #257, opened 2017-10-31
depopulate container without validating groups

shaundillon:
The validation in place makes it impossible to create a front without a group, or to remove it. For some reason, the front mentioned in the tasks did not have a group. This shouldn't affect deletion of containers though. This fixes that.


shaundillon:
[Fronts config tool makes you add metadata before removing a container from a front](https://trello.com/c/4XWTu2Wk/201-fronts-config-tool-makes-you-add-metadata-before-removing-a-container-from-a-front)

PR #269, opened 2018-01-25
Ed tools users

Reettaphant:
Check that all fronts users are in ed tools users group. Set cache validation to true to prevent checking groups on every request.


app/conf/Configuration.scala
```diff
@@ -42,6 +42,10 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
   private def getMandatoryBoolean(property: String): Boolean = getBoolean(property)
     .getOrElse(throw new BadConfigurationException(s"$property of type boolean not configured for stage $stageFromProperties"))
 
+  def getMandatoryStringPropertiesSplitByComma(propertyName: String): List[String] = {
+    getMandatoryString(propertyName).split(",").toList
```
sihil: To prevent an empty string being parsed into List("") you should `filter(_.nonEmpty)` here.

PR #271, opened 2018-02-16
Use an encrypted AMI

sihil:


shtukas:
Thanks for the techtime :)

PR #273, opened 2018-03-01
Platform-specific collections

tomrf1:
Introduces Any/Web/App collections. A very simple implementation while we test app-only collections.

1. Collections are Any by default.
2. Does not allow an Any collection to exist with the same name as an App/Web collection. But there may be both an App and Web version of a collection.
3. Does not allow App/Web collections to be dragged into other fronts.

An Any collection can be split into Web/App if it only belongs to one front:
picture 6

picture 5

A platform-specific collection can be reverted to an Any if it's the only one:
picture 7

Web/App collections are indicated on the main page:
picture 4

### DIFF
--- a/public/src/css/style.css
+++ b/public/src/css/style.css
@@ -1535,6 +1535,10 @@ button {
     margin: 0 10px 0 0;
 }
 
+.tool--remove-button {
+    background: #ed5935 !important;
+}
+
 .tool:hover {
     background: #BBBBBB;
     color: #ffffff;
@@ -1906,6 +1910,12 @@ ol.search_list > li::before {
     padding: 10px 0;
     clear: left;
     height: 30px;
+}
+
+.cnf-form .platform-edit {
+    padding: 10px 0;
+    clear: left;
+    height: 50px;
 }
 
 .cnf-form .action-buttons {
--- a/public/src/js/models/config/collection.js
+++ b/public/src/js/models/config/collection.js
@@ -10,6 +10,7 @@ import deepGet from 'utils/deep-get';
 import fullTrim from 'utils/full-trim';
 import populateObservables from 'utils/populate-observables';
 import urlAbsPath from 'utils/url-abs-path';
+import isPlatformSpecificCollection from 'utils/platform';
 
 export default class ConfigCollection extends DropTarget {
     constructor(opts = {}) {
@@ -36,7 +37,8 @@ export default class ConfigCollection extends DropTarget {
                 'hideShowMore',
                 'backfill',
                 'description',
-                'metadata'
+                'metadata',
+                'platform'
             ]),
             {
                 displayHints: asObservableProps([
@@ -44,7 +46,6 @@ export default class ConfigCollection extends DropTarget {
                 ], observableNumeric)
             }
         );
-
 
         populateObservables(this.meta, opts);
 
@@ -70,6 +71,21 @@ export default class ConfigCollection extends DropTarget {
         });
 
         this.typePicker = this._typePicker.bind(this);
+
+        this.thisIsPlatformSpecificCollection = isPlatformSpecificCollection(this.meta.platform());
+    }
+
+    getPlatform() {
+        switch (this.meta.platform()) {
+            case vars.CONST.platforms.app: return 'app only';
+            case vars.CONST.platforms.web: return 'web only';
+            default: return 'any';
+        }
+    }
+
+    getDisplayName() {
+        const platform = isPlatformSpecificCollection(this.meta.platform()) ? ` (${this.meta.platform()} only)` : '';
+        return this.meta.displayName() + platform || '(no title)';
     }
 
     toggleOpen() {
--- a/public/src/js/models/config/front.js
+++ b/public/src/js/models/config/front.js
@@ -13,6 +13,7 @@ import populateObservables from 'utils/populate-observables';
 import generateCollections from 'utils/generate-collections';
 import {validateImageSrc} from 'utils/validate-image-src';
 import alert from 'utils/alert';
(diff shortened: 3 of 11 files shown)


public/src/js/widgets/columns/fronts-config.html
```diff
@@ -264,6 +264,16 @@
             <input type="checkbox" data-bind="
                 checked: meta.uneditable" />
 
+            <div class="platform-edit">
+                <span>Platform: </span><span data-bind="text: getPlatform()"></span>
+                <!-- ko if: !thisIsPlatformSpecificCollection() && parents().length <= 1 -->
```
susiecoleman: Do the comments need to stay?

public/src/js/models/config/front.js
```diff
@@ -221,9 +222,44 @@ export default class ConfigFront extends BaseClass {
...
 
+    hasDuplicatePlatformSpecificCollection() {
+        const dups = this.collections.items().filter(c => isPlatformSpecificCollection(c.meta.platform()) && c.parents().length > 1);
+        return dups.length > 0;
+    }
+
+    isUniqueCollection(name) {
+        //Checks if this is the only collection in this front with this name
+        const others = this.collections.items().filter(c => c.meta.displayName() === name);
+        return others.length < 2;
```
susiecoleman: I thought displayName() always included the platform specific information?

tomrf1: `meta.displayName` is the name in the actual collection data (stored in s3). We render it with the platform type in the UI, but do not add it in s3

susiecoleman: 👍 thanks

public/src/js/widgets/columns/fronts-config.html
tomrf1: this is actually how you conditionally create html in knockout. It's great!

public/src/js/models/config/collection.js
```diff
@@ -72,6 +73,23 @@ export default class ConfigCollection extends DropTarget {
...
+            default: return 'any';
+        }
+    }
+
+    getDisplayName() {
+        const platform = isPlatformSpecificCollection(this.meta.platform()) ? ` (${this.meta.platform()} only)` : '';
+        return this.meta.displayName() + platform || '(no title)';
+    }
+
+    thisIsPlatformSpecificCollection() {
```
Reettaphant: Why is this a function rather than a constant?

public/src/js/models/config/front.js
```diff
@@ -221,9 +222,44 @@ export default class ConfigFront extends BaseClass {
...
         }
+        if (this.hasDuplicatePlatformSpecificCollection()) {
+            alert('Cannot share platform-specific collections with other fronts');
+            return;
+        }
         return persistence.front.update(this);
     }
 
+    hasDuplicatePlatformSpecificCollection() {
+        const dups = this.collections.items().filter(c => isPlatformSpecificCollection(c.meta.platform()) && c.parents().length > 1);
```
Reettaphant: Would using find here instead of filter better describe what is happening?

public/src/js/utils/platform.js
```diff
@@ -0,0 +1,5 @@
+function isPlatformSpecificCollection(platform) {
+    return platform === 'App' || platform === 'Web';
```
Reettaphant: These strings seem to be appearing in a few places - they could be defined in `/constants/defaults.js` for clarity

public/src/js/widgets/columns/fronts-config.html
```diff
@@ -264,6 +264,16 @@
...
+                <span><button class="tool" data-bind="click: $parents[1].splitCollection.bind($parents[1], meta, parents)">Split into app/web</button></span>
+                <!-- /ko -->
+                <!-- ko if: thisIsPlatformSpecificCollection() && $parents[1].isUniqueCollection(meta.displayName()) -->
+                <span><button class="tool" data-bind="click: $parents[1].changeToAnyPlatform.bind($parents[1], meta)">Change to any</button></span>
```
Reettaphant: Should we display and disable this button if the collection is not a unique collection for clarity?

public/src/js/models/config/collection.js
tomrf1: I assumed it's necessary to pick up on updated values, but perhaps it's only ever called once on creation anyway?

public/src/js/widgets/columns/fronts-config.html
susiecoleman: Fun (hopefully on react's roadmap to introduce this feature)

public/src/js/models/config/front.js
tomrf1: good point

tomrf1: This is weird, `find` works when I run the app, but when I run the tests it's `undefined`...which suggests it's not a real array!? I'll leave it as `filter` as that works

public/src/js/models/config/collection.js
Reettaphant: Yes unless you need it to change because a user did something to the collection but it could be an observable in that case.

public/src/css/style.css
```diff
@@ -1535,6 +1535,10 @@ button {
     margin: 0 10px 0 0;
 }
 
+.tool--remove-button {
+    background: #ed5935 !important;
```
Reettaphant: Annoying if you have to use the `important` here, the fronts tool css is a bit messed up in places.

tomrf1: yes, this is not the first use of `important` here :)

public/src/js/models/config/front.js
tomrf1: ahh that makes sense, will do

PR #280, opened 2018-03-16
Add in a default setup for V2

RichieAHB:
@susiecoleman @Reettaphant 

The README will layout most of what we're up to here but basically this is a base setup with:

- Yarn (because we're using `yarn.lock` and it's referenced in the `scripts`)
- Webpack (also with a hot-reloading dev server setup)
- Babel (with `preset-env`, auto-polyfilling and supporting Chrome 49 and FF 48 *ahem*)
- React / Redux (and some other base libs)
- Jest
- Flow (although I'd __*LOVE*__ to use Typescript, but I'll let it drop 🙄 )
- ES Lint / Prettier

Currently for our styled components I've added in `styled-components`. Dotcom use `emotion` so maybe it's worth swapping it out for that but I think they're mostly doing that for bundle-size reasons - the two do have a very similar API. Additionally, adding default props other than `style` (such as `type` as `password` for an input), is much nicer in `styled-components` and the community is much larger at this point. However, currently `styled-components` doesn't have flow typings (although there is a PR for it [here](https://github.com/flowtype/flow-typed/pull/1814)) so this is a bit of a pain.

I'm on holiday at the start of next week but I'm not doing much so if anything confuses you then send me a hangout chat. I know @gtrufitt was keen to look at this next quarter so between the three of you have a look and see what you think ... I'm open to any ideas*

*Typescript ... I didn't let it drop 😞


client-v2/src/components/App.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import styled from 'styled-components';
+
+const AppContainer = styled('div')`
```
piuccio: you can also use <code>styled.div``</code>

piuccio:
Glad to see you guys are working on it. I wish I had done it ages ago.

client-v2/src/components/App.js
RichieAHB: I would like to but there’s currently an issue with Flow when using that method 😞 https://github.com/styled-components/styled-components/issues/1321 - another reason to use TS!

RichieAHB:
@reettaphant - the script would just be `yarn install` or `yarn dev` if you want the dev server setup (`yarn dev` runs `yarn install` as part of it) or do you mean installing other things?

Reettaphant:
@RichieAHB Yes, it would just be running `yarn install` and making sure yarn itself is actually installed. I think it's a nice thing to have because it follows the pattern of other apps. People can just run it without thinking about the tooling they need installed.

shaundillon:
@Reettaphant and @RichieAHB, PFI has an example script that checks for nvm, runs it, then runs yarn. Just a little extra measure to ensure setup is as smooth as possible. There's also some messaging to help people get deps if they need them

PR #282, opened 2018-03-26
Move play to 2.6

susiecoleman:
[DO NOT MERGE] All testing was done in firefox fronts constantly refreshes on chrome.
Moving to latest version of play 2.5 and latest version of riffraff.


Reettaphant:
This is working now @mbarton!

Reettaphant:
See also: https://github.com/guardian/editorial-tools-platform/pull/170

app/controllers/FaciaContentApiProxy.scala
```diff
@@ -3,27 +3,23 @@ package controllers
...
 import scala.concurrent.ExecutionContext.Implicits.global
 
-class FaciaContentApiProxy(ws: WSAPI, val config: ApplicationConfiguration) extends Controller with PanDomainAuthActions {
+class FaciaContentApiProxy(val deps: BaseFaciaControllerComponents) extends BaseFaciaController(deps) {
   implicit class string2encodings(s: String) {
     lazy val urlEncoded = URLEncoder.encode(s, "utf-8")
   }
 
-  override lazy val actorSystem = ActorSystem()
+  lazy val actorSystem = ActorSystem()
```
mbarton: I don't think this is used? (at least according to IntelliJ)

app/controllers/FaciaToolController.scala
```diff
@@ -20,11 +18,11 @@ import updates._
...
 import scala.concurrent.Future
 
-class FaciaToolController(val config: ApplicationConfiguration, val acl: Acl, val frontsApi: FrontsApi, val faciaApiIO: FaciaApiIO, val updateActions: UpdateActions,
+class FaciaToolController(val acl: Acl, val frontsApi: FrontsApi, val faciaApiIO: FaciaApiIO, val updateActions: UpdateActions,
                           breakingNewsUpdate: BreakingNewsUpdate, val auditingUpdates: AuditingUpdates, val faciaPress: FaciaPress, val faciaPressQueue: FaciaPressQueue,
-                          val configAgent: ConfigAgent, val s3FrontsApi: S3FrontsApi, val mediaServiceClient: MediaServiceClient) extends Controller with PanDomainAuthActions with BreakingNewsEditCollectionsCheck {
+                          val configAgent: ConfigAgent, val s3FrontsApi: S3FrontsApi, val mediaServiceClient: MediaServiceClient, val deps: BaseFaciaControllerComponents) extends BaseFaciaController(deps) with BreakingNewsEditCollectionsCheck {
 
-  override lazy val actorSystem = ActorSystem()
+  lazy val actorSystem = ActorSystem()
```
mbarton: As above, I don't think this is used

app/model/Cached.scala
```diff
@@ -50,6 +49,8 @@ object NoCache {
 
 case class NoCache[A](action: Action[A]) extends Action[A] {
 
+  implicit lazy val executionContext: ExecutionContext = ExecutionContext.Implicits.global
```
mbarton: We shouldn't import the execution context here but rather leave it for each controller using this action to inject as appropriate. We might also need to make this a trait to do that since the case class won't be allowed to have abstract members?

app/permissions/Permissions.scala
```diff
@@ -6,13 +6,13 @@ import conf.ApplicationConfiguration
 import scala.concurrent.ExecutionContext
 
 class Permissions(val appConfig: ApplicationConfiguration) extends PermissionsProvider {
-  override implicit lazy val executionContext: ExecutionContext = play.api.libs.concurrent.Execution.Implicits.defaultContext
+  override implicit lazy val executionContext: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
```
mbarton: I would pass in the execution context in the constructor parameters instead

app/permissions/PermissionsActionCheck.scala
```diff
@@ -1,15 +1,18 @@
...
 import play.api.mvc._
 import services.ConfigAgent
 import util.{AccessDenied, AccessGranted, Acl, Authorization}
 
-import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.{ExecutionContext, Future}
 
 trait PermissionActionFilter extends ActionFilter[UserRequest] {
+
+  implicit lazy val executionContext: ExecutionContext = ExecutionContext.Implicits.global
```
mbarton: As above, I think this should be injected per-controller

build.sbt
```diff
@@ -53,8 +51,8 @@ javaOptions in Universal ++= Seq(
 
 routesGenerator := InjectedRoutesGenerator
 
-scalacOptions := Seq("-unchecked", "-optimise", "-deprecation", "-target:jvm-1.8",
-    "-Xcheckinit", "-encoding", "utf8", "-feature", "-Yinline-warnings")
+scalacOptions := Seq("-unchecked", "-opt:l:method", "-deprecation", "-target:jvm-1.8",
```
mbarton: I'd remove `optimise` altogether unless there's a non-performance reason to have it?

Using this looks fine and the optimisation [is solid](https://www.scala-lang.org/news/2.12.0/#new-optimizer) but a few years ago the optimiser had been known to cause issues in libraries eg older versions of [Akka](https://doc.akka.io/docs/akka/2.4/intro/getting-started.html#Do_not_use_-optimize_Scala_compiler_flag)

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=0.13.11
+sbt.version=0.13.15
```
mbarton: Might as well bump to SBT 1?

README.md
```diff
@@ -123,6 +123,12 @@ Modify the files
 
    > merge_slashes off;
 
+* `/usr/local/etc/nginx/sites-enabled/fronts.conf` and add this line inside the `location /` block which is in the `server` block
+
+   > proxy_http_version 1.1;
+
+   This HTTP 1.1 is required by play (as opposed to nginx default of 1.0) in order to send chunked responses.
```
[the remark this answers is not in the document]
Reettaphant: I don't know why this line is added here as I couldn't find where in the fronts tool chunked responses are being sent and my fronts tool works fine even without this line.

app/conf/Configuration.scala
```diff
@@ -47,8 +47,13 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
   }
 
   object environment {
-    val stage = stageFromProperties.toLowerCase
+    val stage = stageFromProperties
```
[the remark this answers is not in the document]
Reettaphant: This is actually a bit annoyingly complicated because we don't want the `stage` value to distinguish between code and dev environments. This is because `stage` helps us distinguish if we are reading and writing from the code bucket (on code and dev) or the prod bucket (prod only). I've made a slightly ugly fix for this - let me know what you think...

app/controllers/FaciaContentApiProxy.scala
Reettaphant: I don't think so either.

README.md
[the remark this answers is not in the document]
mbarton: Cool lets remove it from the README

app/conf/Configuration.scala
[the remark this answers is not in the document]
mbarton: LGTM

app/controllers/CollectionController.scala
```diff
@@ -8,6 +8,7 @@ import services.Press
 import updates._
 import util.Acl
 import util.Requests._
+import scala.concurrent.ExecutionContext.Implicits.global
```
mbarton: This will work fine but I think we should go a step further and add the execution context as an implicit parameter on the constructor of the controllers. It will then get filled in with the implicit `executionContext` Play constructs in `BuiltInComponents`.

Same goes for all the global context imports below.

PR #283, opened 2018-04-03
Temporarily disable Panda group checks

mbarton:
We have had recent incidents where the Google Directory API returns a rate limit exception. The reasons remain unclear as we are confident that we remained under the limit.

In such situations we would normally enable the emergency login switch however that was not fully reliable during the last outage. Pending further testing this PR will disable the checks that were calling the offending API.

It should only be merged if the problem recurs.


mbarton:
We should close this once #352 is merged

mbarton:
Closing now that facia is using permissions for access

PR #284, opened 2018-04-05
Browse fronts

Reettaphant:
Adds the ability to browse fronts with a specific priority in a pretty dropdown menu. 

This pr also adds types for the fronts config, fronts and collections and for the app config. It also adds enzyme to allow for testing components with jest and adds in checks for flow coverage.


public/client-v2/src/actions/FrontsConfig.js
```diff
@@ -0,0 +1,38 @@
...
+  return {
+    type: 'CAUGHT_ERROR',
+    message: 'Could not fetch fronts config',
+    error,
+    receivedAt: Date.now()
+  };
+}
+
+export default function getFrontsConfig() {
+  return (dispatch: Dispatch) => {
```
RichieAHB: I wonder if this code is getting covered with just this `Dispatch` type? Not the end of the world but may need to type it with a `Thunk`. Although I think the `Thunk` type that I added into `types/Stores.js` might be unfit for purpose. Something like the below might do it:

```
type ThunkDispatch = (action: ThunkAction) => void;

export type Store = ReduxStore<State, Action>;
export type GetState = () => State;
export type Dispatch = ReduxDispatch<Action> & ThunkDispatch;
export type ThunkAction = (dispatch: Dispatch, getState: GetState) => any;
```

public/client-v2/src/components/FrontsEdit/Edit.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import React from 'react';
+import Fronts from './Fronts';
+
+type Props = {
+  match: Object
```
RichieAHB: Could we type this with the `Match` type from `redux-router`. Object will just have an `any` type for all string keys.

public/client-v2/src/components/FrontsEdit/Fronts.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  renderSelectPrompt = () => {
+    if (!this.state.selectedFront) {
+      return <option value="">Choose a front...</option>;
+    }
+    return null;
+  };
+
+  renderFrontOption = (front: FrontDetail) => (
```
RichieAHB: Could this method be inline inside render?

public/client-v2/src/constants/priorities.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+import type { Priorities } from '../types/Fronts.js';
+
+export const priorities: Priorities = {
```
RichieAHB: Perhaps we could remove priorities from the front config entirely if we're specifying them by hand here? Seems like a duplication for no reason?

public/client-v2/src/components/FrontsEdit/Fronts.js
```diff
@@ -0,0 +1,84 @@
...
+    </option>
+  );
+
+  render() {
+    if (!this.props.frontsConfig.fronts) {
+      return <div>Loading</div>;
+    }
+
+    const { match: { params: { priority } } } = this.props;
+    const filteredFronts: Array<FrontDetail> = filterFrontsByPriority(
```
RichieAHB: Here is a really good case for using `reselect` so that we can memoize this data when calling `mapStateToProps` so that other components can take advantage of it. I can come over and show you what I mean as I'm sure this makes no sense!

public/client-v2/src/constants/priorities.js
Reettaphant: The priorities here are settings for each priority we are displaying on the client. They are empty objects for now as we only need their names but they will have some data in them in the future. The priorities of fronts coming back in the fronts config are just names of priorities for each front and the priority types are just making sure that priorities are always conforming to the shape here. Maybe it makes sense to export the priorities type from this file rather than from `types/Fronts` ?

public/client-v2/src/components/FrontsEdit/Fronts.js
Reettaphant: Maybe but I'm not too keen to shuffle this around now as this component should be getting more complicated quite soon.

PR #286, opened 2018-04-06
[WIP] First commit with feed

RichieAHB:
Bit odd to PR into your branch but wanted to check with you before doing it!

Just added some basic feed stuff in, a few `CAPI` components that will potentially get swapped out for other `FrontsAPI` components but should provide a good basis for getting feeds in.


public/client-v2/src/components/CAPI/CAPISearchQuery.js
```diff
@@ -0,0 +1,79 @@
+// @flow
+
+import * as React from 'react';
+import capiQuery from '../../services/capiQuery';
+import Async, { type AsyncChild } from '../util/Async';
+
+type ImageAsset = {
```
Reettaphant: I'd image a lot of the types defined here, such as images and articles, are going to be used across the application (clipboard, rendering collections) so maybe they should live in a separate file?

public/client-v2/src/services/capiQuery.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+
+import { qs } from '../util/qs';
+
+const API_BASE = 'https://content.guardianapis.com/';
+
+const capiQuery = (apiKey: string) => ({
```
Reettaphant: Should we return appropriate errors to users in there is something wrong with capi? What do we display here if capi is not returning any search results?

public/client-v2/src/services/capiQuery.js
```diff
@@ -0,0 +1,20 @@
...
+  search: async (params: Object) => {
+    const response = await fetch(
+      `${API_BASE}search${qs({
+        ...params,
+        'api-key': apiKey
+      })}`
+    );
+
+    return response.json();
```
Reettaphant: I was looking at the checks the old app does on capi results and it checks that an article has in an internalPageCode before showing it in the dropdown - we should probably repeat this check.

public/client-v2/src/services/capiQuery.js
RichieAHB: Yeah definitely, I wasn't quite sure whether we were actually going to use CAPI (for the thing you said about preview content which I hadn't thought about) so didn't do everything there. Let's work out what out plan is there and we can definitely spec this up! 👍

public/client-v2/src/components/Feed.js
```diff
@@ -0,0 +1,65 @@
...
+import CAPISearchQuery from './CAPI/CAPISearchQuery';
+import { type Element } from './CAPI/CAPISearchQuery';
+/* eslint-enable import/no-duplicates */
+import ScrollContainer from './ScrollContainer';
+import FeedItem from './FeedItem';
+import Loader from './Loader';
+
+// TODO: get apiKey from context (or speak directly to FrontsAPI)
+
+const getThumbnail = (_elements: Element[]) => {
```
Reettaphant: All the logic that extract data from capi requests will probably lived in some shared file in the end as they will be used across the app but I guess we can start moving them when multiple components start making capi requests.

public/client-v2/src/services/capiQuery.js
RichieAHB: Is that a Front API specific thing? I don't think I've ever seen it in CAPI but yeah, we should definitely do that ... do you know what it represents?

public/client-v2/src/components/Feed.js
RichieAHB: I did do this but for some reason flow just started marking my types as `any` and I couldn't be bothered to figure it out 😞

public/client-v2/src/components/CAPI/CAPISearchQuery.js
RichieAHB: See the below comment - I have some figuring out to do with flow about re-exporting types but I totally agree!

RichieAHB:
Going to PR into master now instead.

PR #289, opened 2018-04-09
Move to fronts API and tidy up request components

RichieAHB:
This changes the client to use the Fronts API instead of CAPI. With various tidies for components to do with querying CAPI.


public/client-v2/src/util/qs.js
```diff
@@ -2,17 +2,21 @@
...
  */
-const qs = (o: StringableObject) =>
+const qs = (o: { [key: string]: ?string }) =>
   `?${Object.keys(o)
-    .map(key => `${enc(key)}=${enc(o[key])}`)
+    .map(key => {
+      const val = o[key];
+
+      if (!val) {
+        return '';
```
Reettaphant: I know it's not much but I think we should add the new behaviour to tests

public/client-v2/src/components/CAPI/CAPITagQuery.js
```diff
@@ -0,0 +1,67 @@
...
+    ) {
+      this.setupCAPI(this.props.baseURL, this.props.fetch);
+    }
+  }
+
+  setupCAPI(baseURL?: string, fetch?: Fetch): void {
+    this.capi = capiQuery(baseURL, fetch).tags;
+  }
+
+  capi: $ElementType<$Call<typeof capiQuery, string>, 'tags'>;
```
Reettaphant: I think we want to restrict the tag types we are searching here as they are not all relevant to fronts. The current tool also allows for choosing the section option to search for section tags but this can be added in another pr.

public/client-v2/src/components/CAPI/CAPITagInput.js
```diff
@@ -0,0 +1,65 @@
...
+              params={{ q: inputValue }}
+            >
+              {({ value }) =>
+                value &&
+                value.response.results.map((tag, index) => (
+                  <div
+                    {...getItemProps({ item: tag })}
+                    key={tag.id}
+                    style={{
+                      backgroundColor:
```
Reettaphant: Can we abstract the styling logic out as there is quite a lot of going on here already?

PR #291, opened 2018-04-10
Error Display

Reettaphant:
Add an error banner displaying an error to the user if we get an error back from an action. Error gets cleared on route change.

### DIFF
--- a/public/client-v2/src/components/ErrorBanner.js
+++ b/public/client-v2/src/components/ErrorBanner.js
@@ -1,0 +1,25 @@
+// @flow
+
+import React from 'react';
+import styled from 'styled-components';
+import { error } from '../styleConstants';
+import type { ActionError } from '../types/Action';
+
+type Props = {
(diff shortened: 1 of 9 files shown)


public/client-v2/src/components/ErrorBanner.js
```diff
@@ -0,0 +1,25 @@
...
+import styled from 'styled-components';
+import { error } from '../styleConstants';
+import type { ActionError } from '../types/Action';
+
+type Props = {
+  error: ActionError
+};
+
+const ErrorContainer = styled('div')`
+  background-color: ${error.primary};
```
RichieAHB: I don't think we should change it now but we should probably think about using `styled-components` [themeing](https://www.styled-components.com/docs/advanced#theming) for this. I forgot about it when we talked earlier but it will make it easier to theme things nicely. For now though this can definitely stay as it is!

PR #292, opened 2018-04-10
Routing

Reettaphant:
- FrontId is q part of the front edit route and changes as users select different fronts
- Restricts valid priorities when routing
- Adds a generic not found component.


public/client-v2/src/components/App.js
```diff
@@ -20,8 +22,13 @@ const AppContainer = styled('div')`
 const App = () => (
   <AppContainer>
     <Switch>
-      <Route exact path="/:priority" component={FrontsEdit} />
+      <Route
+        exact
+        path={`/:priority(${Object.keys(priorities).join('|')})/:frontId?`}
```
RichieAHB: Didn't know this exited for `react-router` 👍

public/client-v2/src/components/FrontsEdit/Fronts.js
```diff
@@ -26,15 +30,23 @@ class Fronts extends React.Component<FrontsComponentProps, ComponentState> {
     selectedFront: null
   };
 
+  componentWillMount() {
```
RichieAHB: These methods are being deprecated in future version of React.  For this one we can just set the state in the `constructor`. I actually don't think we should be using state here but actually just using `frontId` directly and using the prop to set the select value as it'll always be synced up no?

public/client-v2/src/components/FrontsEdit/Fronts.js
```diff
@@ -26,9 +26,17 @@ type ComponentState = {
 };
 
 class Fronts extends React.Component<FrontsComponentProps, ComponentState> {
-
   componentDidMount() {
-    this.props.frontsActions.getFrontsConfig();
+    this.props.frontsActions.getFrontsConfig().then(() => {
```
RichieAHB: I wonder whether we should handle this in the component above this? This seems like an odd side effect to have (a quick render then redirect when we had this information in the parent). Can we not check if the selected front is valid in the parent component then either render the selected front here or render the select to say "choose front". What do you think?

Reettaphant: Yes, I was thinking about it and it's a bit weird but we are fetching the fronts in this component. We could fetch the fronts in the component where we do the routing, we wouldn't need them for anything else except for deciding which routes are valid.

PR #295, opened 2018-04-11
Ensure CAPI URL is correct for each stage

RichieAHB:
This PR makes a few changes:
- Adds CAPI URLs into the config (done by querying the routes, which seems nice but I might be wrong!)
- Adds a way of providing CAPI params via React 'context`- this means we can drop the FrontsCAPI components as all they did was provide arguments to the raw CAPI components. This now happens through context.
- Specifies the feed to be pulling from `capiLiveUrl` at all times. I'll make a PR to pull in the `capiPreviewUrl` and add a toggle switch to demo that functionality.
- Also typed some of the `Config` areas with the pre-existing type
- Added polling to the `CAPISearch` component - currently set to `30000`


client-v2/src/components/CAPI/__tests__/SearchQuery.spec.js
```diff
@@ -13,49 +13,53 @@ const succesfulReturn = {
...
-    global.fetch = jest.fn().mockImplementation(() =>
-      Promise.resolve({
-        json: () => succesfulReturn
-      })
-    );
+  it('needs the tests turning back on when enzym is updated', () => {
+    console.log('turn on these tests when enzyme get Context API');
+    expect(true).toBeTruthy();
   });
+  // beforeEach(() => {
```
Reettaphant: Why are there so many commented out tests?

app/controllers/DefaultsController.scala
```diff
@@ -30,11 +30,13 @@ case class Defaults(
   collectionCap: Int,
   navListCap: Int,
   navListType: String,
-  collectionMetadata: Iterable[Metadata]
+  collectionMetadata: Iterable[Metadata],
+  capiLiveUrl: String,
+  capiPreviewUrl: String
```
Reettaphant: If you make these urls optional, then you don't need to include them in the v1 config, you can only send them back to the v2 app.

app/controllers/DefaultsController.scala
```diff
@@ -61,7 +63,9 @@ class DefaultsController(val config: ApplicationConfiguration, val acl: Acl, val
           config.facia.navListType,
           Metadata.tags.map{
             case (_, meta) => meta
-          }
+          },
+          routes.FaciaContentApiProxy.capiLive("").absoluteURL(true),
```
Reettaphant: Is this going to know what the correct domain is?

client-v2/src/components/CAPI/__tests__/SearchQuery.spec.js
RichieAHB: Enzyme doesn't support React 16.3 `context` so I've had to comment them out. What I will actually do now I think about it is export the components that haven't been wrapped in a context consumer (how they were when I wrote the test) and just test the raw component.

app/controllers/DefaultsController.scala
RichieAHB: Good plan 👍

app/controllers/DefaultsController.scala
RichieAHB: Yeah, any route builders inside the Play app will have to use something similar to this (or something similar) to build absolute URLs.

RichieAHB:
@Reettaphant hopefully this passes and I've addressed those fixes. Interestingly, if you look at the preview feed it's full of workflow stuff, hence the `internalPageCode` thing though. But I'm assuming we can't query CAPI to ask for things with only `internalPageCode`? If we can't then we will always show a random amount of results, which is a bit annoying ...

Reettaphant:
I'm not sure if you can only ask for this with `internalPageCode`, I can't see anything in the docs at least. But if we are returning a large number of results each time I don't think anyone is going to notice.

RichieAHB:
@reetta yup - you’re probably right 👍

PR #297, opened 2018-04-12
remove accidental .vscode from version control

RichieAHB:
Happy to keep it checked in but I assume other people will want other things - plus it's always there, unstaged, every time I edit it!


RichieAHB:
@shaundillon 👍 😉

PR #303, opened 2018-04-18
Fronts dropdown improvements

Reettaphant:
- Sort fronts alphabetically (we can do something smarter later)
- Don't display breaking news in fronts dropdown
- Fronts dropdown should be a separate component
- I removed most of the fronts dropdown tests because they didn't really make any sense (we are testing fronts filtering elsewhere). I added them just to get enzyme into the project in the first place.

### DIFF
--- a/client-v2/src/components/FrontsEdit/Fronts.js
+++ b/client-v2/src/components/FrontsEdit/Fronts.js
@@ -7,8 +7,9 @@ import { withRouter } from 'react-router';
 import getFrontsConfig from '../../actions/FrontsConfig';
 import { GetFrontsConfigStateSelector } from '../../selectors/frontsSelectors';
 import CollectionContainer from './CollectionContainer';
+import FrontsDropDown from './FrontsDropdown';
 import { getFrontCollections } from '../../util/frontsUtils';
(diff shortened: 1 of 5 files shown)


client-v2/src/components/FrontsEdit/Fronts.js
```diff
@@ -54,13 +34,12 @@ class Fronts extends React.Component<FrontsComponentProps> {
...
     return (
       <div>
-        <select
-          value={this.props.frontId}
-          onChange={e => this.selectFront(e.target.value)}
-        >
-          {this.renderSelectPrompt()}
-          {fronts.map(this.renderFrontOption)};
-        </select>
+        <FrontsDropDown
```
RichieAHB: This here is a perfect candidate for a container that doesn't take any props but instead gets its data from the store (or redux-router, wherever the Source of Truth™️ is). This container can do all of the data shaping we need (probably through selectors) inside of mapStateToProps and pass this data directly to a custom DropdownComponent that knows nothing about a "front" or "priority". I might make a PR here as an example because I know many of these things don't seem very concrete! I think after then there are a few other areas we can look at setting up more specific containers with less knowledge of the application.

PR #304, opened 2018-04-19
Live collections

Reettaphant:
Allows for toggling between live and preview collections on a front.
Also uses selectors for getting collections and articles as this is what we should be doing anyway.

### DIFF
--- a/client-v2/src/components/FrontsEdit/CollectionContainer.js
+++ b/client-v2/src/components/FrontsEdit/CollectionContainer.js
@@ -6,11 +6,12 @@ import { bindActionCreators } from 'redux';
 import getFrontCollection from '../../actions/Collection';
 import getArticlesForCollection from '../../actions/Articles';
 import CollectionDetail from './CollectionDetail';
+import Button from '../Button';
 
 import type { ConfigCollectionDetailWithId } from '../../types/FrontsConfig';
-import type { Collection } from '../../types/Collection';
-import type { CapiArticle } from '../../types/Capi';
+import type { Collection, CollectionArticles } from '../../types/Collection';
 import type { State } from '../../types/State';
+import type { CapiArticle } from '../../types/Capi';
 
 type Props = {
   collection: ConfigCollectionDetailWithId
@@ -22,11 +23,25 @@ type ConnectedComponentProps = Props & {
     [string]: Collection
   },
   collectionArticles: {
-    [string]: Array<CapiArticle>
+    [string]: {
+      draft: CollectionArticles,
+      live: CollectionArticles
+    }
   }
 };
 
-class CollectionContainer extends React.Component<ConnectedComponentProps> {
+type ComponentState = {
+  browsingState: 'draft' | 'live'
+};
+
+class CollectionContainer extends React.Component<
+  ConnectedComponentProps,
+  ComponentState
+> {
+  state = {
+    browsingState: 'draft'
+  };
+
   componentDidMount() {
     const { props: { collection: { id } } } = this;
     const currentCollection = this.props.collections[id];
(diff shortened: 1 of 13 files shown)


client-v2/src/components/FrontsEdit/CollectionContainer.js
```diff
@@ -6,55 +6,56 @@ import { bindActionCreators } from 'redux';
...
 import type { ConfigCollectionDetailWithId } from '../../types/FrontsConfig';
 import type { Collection } from '../../types/Collection';
-import type { CapiArticle } from '../../types/Capi';
 import type { State } from '../../types/State';
+import type { CapiArticle } from '../../types/Capi';
 
 type Props = {
-  collection: ConfigCollectionDetailWithId
+  collectionConfig: ConfigCollectionDetailWithId,
+  browsingStage: string // eslint-disable-line react/no-unused-prop-types
```
shaundillon: If this isn't used, does it need to be here?

client-v2/src/selectors/collectionArticleSelectors.js
```diff
@@ -0,0 +1,38 @@
...
+const collectionIdSelector = (state: State, id: string) => id;
+
+const articleStageSelector = (state: State, id: string, stage: string) => stage;
+
+const getCollectionArticles = (
+  collectionArticles: { [string]: CollectionArticles },
+  id: string,
+  stage: string
+): Array<CapiArticle> => {
+  if (!collectionArticles) {
```
shaundillon: This can be combined with the check below
```javascript
if (!collectionArticles || !collectionArticles[id]) {
  return [];
}
```

or, if you like a ternary, you can combine all 3:
```javascript
return !collectionArticles || !collectionArticles[id] ? [] : collectionArticles[id][stage];
```

client-v2/src/selectors/collectionSelectors.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import { createSelector } from 'reselect';
```
shaundillon: I've never come across selectors before, this is very cool.

client-v2/src/components/FrontsEdit/CollectionContainer.js
Reettaphant: It is used because we are passing it to the article selector but as this happens outside the component itself, eslint gets confused and thinks it's an error.

client-v2/src/selectors/collectionSelectors.js
Reettaphant: Yes, I know! @RichieAHB introduced me to them.

client-v2/src/components/FrontsEdit/CollectionContainer.js
```diff
@@ -6,55 +6,56 @@ import { bindActionCreators } from 'redux';
...
       this.props.frontsActions.getFrontCollection(id).then(() => {
         this.props.frontsActions.getArticlesForCollection(
-          this.props.collections[id],
+          this.props.collection,
           id
         );
       });
     } else {
-      this.props.frontsActions.getArticlesForCollection(currentCollection, id);
+      this.props.frontsActions.getArticlesForCollection(collection, id);
```
RichieAHB: Here this looks like every time we unmount this component (i.e. change fronts), we'll be refetching all the articles despite them already being in the store?

client-v2/src/components/FrontsEdit/CollectionContainer.js
```diff
@@ -6,55 +6,56 @@ import { bindActionCreators } from 'redux';
...
     );
   }
 }
 
-const mapStateToProps = (state: State) => ({
-  collections: state.collections,
-  collectionArticles: state.collectionArticles
+const mapStateToProps = (state: State, props: Props) => ({
+  collection: collectionSelector(state, props.collectionConfig.id),
+  collectionArticles: collectionArticlesSelector(
```
RichieAHB: Passing separate props as multiple arguments after the state argument to reselect can work but it quite often breaks the memoization so you don't actually get any of the benefits over just using plain functions - see here for a bit more detail: https://github.com/reduxjs/reselect/issues/253

PR #310, opened 2018-05-02
Refactor dropdown

RichieAHB:
This refactors the `FrontsDropdown` component away from being a "fronts" dropdown to just being a dropdown (while adding a `FrontsDropdown` container). Most if not all of the logic is the same as before but all of the data handling around fronts that is required for the dropdown is being done outside the dumb component now (and in some slightly changed selectors). E.g. our component is now not dependent on `react-router` (through using `history.push`) and `priority`. This pattern can be used in other places in the app too. Find a fronts component; describe it in terms of it's own domain; take any data munging logic out of the component and instead do this when mapping our state (or router params) to props; pass in in through a container.

I've also split out `collection` and `fronts` out from `frontsConfig` in our selectors so that we can use them separately nice and deep in the component tree only where needed.

This PR is open to discussion and I may have missed some reasons for doing other bits and pieces in here but I think some of these patterns will be really useful as the app grows. Arguably doing it here is a bit pedantic but there are some other places in the app (specifically around collections etc.) that would benefit from this going forward and would take a bit more of a discussion to draw some boundaries.


RichieAHB:
@Reettaphant @jonathonherbert can we get these changes merged or are there any issues with this approach (aside from the slightly pedantic PR). Just looking at sorting out the facia types I've touched them in here so wanted to make the changes on top of these ones.

client-v2/src/selectors/frontsSelectors.js
```diff
@@ -4,64 +4,68 @@ import { createSelector } from 'reselect';
...
 
-const frontsIdSelector = createSelector([frontsSelector], fronts => {
+const getFronts = createSelector(
+  [rawFrontsSelector],
+  (fronts: Front): FrontsWithIds =>
+    Object.keys(fronts)
+      .map((id): FrontDetailFull => ({
+        id,
+        ...fronts[id],
+        priority: fronts[id].priority || 'editorial'
```
Reettaphant: we should remove breaking news fronts here

client-v2/src/selectors/frontsSelectors.js
```diff
@@ -4,64 +4,68 @@ import { createSelector } from 'reselect';
...
+        };
+      },
+      {}
+    )
+);
+
+const getFrontsWithPriority = (
+  state: State,
+  priority: string
+): FrontDetailFull[] => getFrontsByPriority(state)[priority] || [];
```
Reettaphant: Users are unlikely to do a lot of switching around different priorities so sorting all fronts by priorities in the selectors might be too much complexity but also it is a nice thing to have

RichieAHB: So the reason I did this is that it means the selectors don't break when we pass a different props object. It's annoyingly subtle but basically, when we were passing a props object before to the selectors, even if the `priority` key was the same, if we passed an object with a different reference (i.e. when another prop on the container that was using it changed) then `reselect` would recompute the selector because it does shallow equality checking. This might not make sense so maybe we can talk about it as it's not intuitive at all.

RichieAHB: ☝️  This is the reason that I've refactored a few of the other selectors.

client-v2/src/selectors/frontsSelectors.js
RichieAHB: Yeah good plan, is that just a precaution? I need to quickly run through with you the way a breaking news front may appear not sure I quite understand it!

client-v2/src/selectors/frontsSelectors.js
Reettaphant: Ok, I think I get it. So if we are passing in component props to selectors, we should separate out the logic that uses the prop from all other data manipulation we are doing.

client-v2/src/selectors/frontsSelectors.js
Reettaphant: The only thing to understand about the breaking news front is that we just don't want to show it here at all, it has it's on separate breaking news page which has slightly different behaviour from everything else. I saw you filtering the breaking news front further down this file but I don't think we should be adding it an editorial priority here as it is not an editorial front and does not get pressed.

PR #311, opened 2018-05-03
Add in the outbound links for CAPI

RichieAHB:
As above - just trying to get the feed to parity with v1

### DIFF
--- a/client-v2/src/services/capiQuery.js
+++ b/client-v2/src/services/capiQuery.js
@@ -28,7 +28,13 @@ type Element = ImageElement;
 type Article = {
   webTitle: string,
   webUrl: string,
-  elements?: Element[]
+  elements?: Element[],
+  fields?: {
+    trailText?: string
+  },
+  frontsMeta: {
+    tone: string
+  }
 };
 
 type CAPISearchQueryReponse = {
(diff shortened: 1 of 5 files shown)


client-v2/src/services/capiQuery.js
```diff
@@ -28,7 +28,13 @@ type Element = ImageElement;
 type Article = {
   webTitle: string,
   webUrl: string,
-  elements?: Element[]
+  elements?: Element[],
+  fields?: {
+    trailText?: string
+  },
+  frontsMeta: {
```
Reettaphant: Does `tone` refer to the tone tag coming from capi? If this is the case, storing it in the an object called `frontsMeta` is a bit misleading as in the old tool (and I've been replicating this in the new tool) meta refers to overrides we make in the fronts tool.

RichieAHB: This is how they come back directly from CAPI and this is the type that represents the response. We should definitely transform them though at the point these get bought into a front (drag and drop or similar)

Reettaphant: Ah, I think I understand now. There is some default meta for articles stored in capi - we need to figure out how this works and what we should do with it.

PR #317, opened 2018-05-04
Fix 'also on' not appearing for commercial collections used elsewhere, and improve the messaging.

jonathonherbert:
'Also on' now appears when collections that appear in a front with a commercial priority are used elsewhere.

The 'Also on' notice is now followed by a list of the priorities of other fronts that use that collection.

If users are working on a collection that also appears in a front with a commercial priority, they receive a 'Careful: ' warning, too.

### DIFF
--- a/public/src/js/widgets/collection.html
+++ b/public/src/js/widgets/collection.html
@@ -38,7 +38,13 @@
         </span>
 
         <span class="also" data-bind="if: alsoOn.length">
-            <a class="alsoOnToggle" data-bind="click: alsoOnToggle">also on&hellip;</a>
+            <a class="alsoOnToggle" data-bind="click: alsoOnToggle">also on&hellip;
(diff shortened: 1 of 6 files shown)


public/src/js/widgets/collection.html
```diff
@@ -37,8 +37,22 @@
             (<span class="non-zero">updating...</span>)
         </span>
 
-        <span class="also" data-bind="if: alsoOn.length">
-            <a class="alsoOnToggle" data-bind="click: alsoOnToggle">also on&hellip;</a>
+        <span class="also" data-bind="if: alsoOn && alsoOn.length">
```
Reettaphant: Could we simplify the logic here by only doing one conditional check: 
```
<span data-bind="if: !alsoOnMeritsWarning>Also on&hellip;</span>
<span data-bind="if: alsoOnMeritsWarning"> 
    Careful: also on
 </span>
```

PR #319, opened 2018-05-04
Add d option to Webpack development start up 

gtrufitt:
My Webpack knowledge is stale, so feel free to reject for alternatives, but turning on sourcemaps with `-d` is a shortcut for `--debug --devtool cheap-module-eval-source-map --output-pathinfo`

> -d is a shortcut for --debug --devtool source-map --output-pathinfo where:

> - --debug activates debug mode for loaders and its behaviour depends on each loader. You can activate it via param { debug: true } in your webpack.config.js
> - --devtool is basically allow to select type of sourcemaps for your files via param { devtool: "sourcemap" } or if you use webpack@2: { devtool: 'eval-cheap-module-source-map' }(See all options)
> - --output-pathinfo will write original filepath into webpack's requires like this: __webpack_require__(/* ./src/containers/App/App.js */15). 
You can activate it via: { output: { pathinfo: true } }

## Before

before

## After
after

Longest description for a 2 char change.

### DIFF
--- a/client-v2/config/webpack.config.dev.js
+++ b/client-v2/config/webpack.config.dev.js
@@ -4,6 +4,7 @@ import common from './webpack.config.common.js';
 export default {
   ...common,
   mode: 'development',
+  devtool: 'cheap-module-eval-source-map',
   devServer: {
     contentBase: path.join(__dirname, '../'),
     compress: true,


gtrufitt:
Maybe we want to add specifically in the config rather than via CLI.

RichieAHB:
Looks good to me although like you say we can probably add it in config and probably don't need `output-pathinfo` but I don't mind if other people need it ... can't imagine it adds any time to build / rebuild.

gtrufitt:
Added debug and devtool to the config rather than via the CLI. Left out output path info for now as source maps probably make it less useful anyway.

PR #320, opened 2018-05-08
Add Alias mapping

gtrufitt:
Alias Mapping in Webpack means that we can import files as `Types/xxx`, `Components/xxx` etc instead of relative path imports. I've found this clearer and easier to work with in the past (rather than knowing how deep I am in a file). 

*Up for discussion, do you find this easier to work with too?*

- [Webpack Resolve Alias](https://webpack.js.org/configuration/resolve/#resolve-alias)
- [Flow Module Name Mapper](https://flow.org/en/docs/config/options/#toc-module-name-mapper-regex-string)
- [ESLint Module Resolver Webpack](https://www.npmjs.com/package/eslint-import-resolver-webpack)

Most of the PR is updating relative imports to mappings but [this commit](https://github.com/guardian/facia-tool/commit/a3b7d32c493b1eafe070c6a74957d0b6d1801495) shows the config changes for this and [this one](https://github.com/guardian/facia-tool/commit/02f13bb947a0821efc1470bf777e98d40ed92a39) shows what the remapping looks like.

I'd also maybe suggest that we import *all* files in this way (as some are imported relative to current file)?

image


gtrufitt:
> I didn't used to like this as I've had edge cases where aliases have mapped to node_modules and vice versa

Interesting, I've not seen this, seems a little edge-casey? Especially when being explicit like this? Frontend uses Aliases, so might find out what their experiences were.

> specially as VSCode autocompletes them for me

@RichieAHB  I've added [jsconfig](https://github.com/guardian/facia-tool/pull/320/commits/98031bfaba5d165d0aed4cf3a60a684e71408107) (yet another path mapping..)

pathmap

> explicitness of the relative imports

Yeah, I think maybe on a component level, relative imports are a good thing (`./childComponent` or whatever) but on a top level, especially with the types for example, having to work out where I am and `../` or `../../` etc or `./` rather than explicit `Types/x` is less clear in both usage and quickly being able to reason with what's being imported.

RichieAHB:
@gtrufitt it definitely is edge casey - but it took _forever_ too debug a missing named import from what turned out to be the wrong module!

However I think when we had this issue the source maps we were using were pretty poor so perhaps with the better source maps you can get "these days" then it'll be much easier to debug? I'm happy to go with it though just wanted to flag my few issues with this but I think there are probably much better things in place to catch that 👍 

Could you add a `Containers` path too - as this dir will be added pretty soon and don't want to forget to do it

gtrufitt:
@RichieAHB @jonathonherbert Mapping to directory without alias mapping works https://github.com/guardian/facia-tool/pull/320/commits/34fda97e89bf06397819fec8f649528d56bb5758 though did have to turn of Eslint case insensitivity if we want to use that convention? https://github.com/guardian/facia-tool/pull/320/commits/d6ddc4ef59763a5d498398907bf6aa5a348b0fcd

PR #326, opened 2018-05-21
Genericise common functionality

jonathonherbert:
Genericise functionality we think is likely to be common across different applications - fronts, breaking news, story packages - into code stored in a shared folder.


client-v2/src/actions/ExternalArticles.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+import type { Action } from 'types/Action';
+
+function errorReceivingArticles(error: string): Action {
+  return {
+    type: 'CAUGHT_ERROR',
+    message: 'Could not fetch articles from Capi',
```
RichieAHB: What's the difference between 'error' and 'message' here. Might be better to leave rendering a generic message to the view layer?

client-v2/src/actions/FrontsConfig.js
```diff
@@ -28,6 +45,79 @@ function errorReceivingConfig(error: string): Action {
...
+        dispatch(
+          batchActions([
+            collectionReceived(collection),
+            articleFragmentsReceived(articleFragments)
+          ])
+        );
+        return collectionWithDraftArticles.live
+          ? [
+              // We use a set to dedupe our article ids
+              ...new Set([
```
RichieAHB: Pretty sure this will be polyfilled automatically by 'useBuiltIns' in babel but might be worth a sanity check!

client-v2/src/actions/FrontsConfig.js
```diff
@@ -28,6 +45,79 @@ function errorReceivingConfig(error: string): Action {
   };
 }
 
+function getFrontCollection(collectionId: string) {
+  return (dispatch: Dispatch, getState: () => State) =>
+    getCollection(collectionId)
+      .then((res: Object) => {
+        const collectionConfig = getCollectionConfig(getState(), collectionId);
```
RichieAHB: Why are there selectors in an action? Would it not be sending the raw data back to the reducer and deferring the derivations until they get used in the view?

client-v2/src/types/FaciaApi.js
```diff
@@ -0,0 +1,76 @@
...
+  },
+  collections: {
+    [string]: CollectionConfigResponse
+  }
+};
+
+type FrontConfigResponseWithoutPriority = $Diff<
+  FrontConfigResponse,
+  { priority: PriorityName | void }
+>;
```
Reettaphant: Can we just `FrontConfigResponseWithoutPriority` here, we don't seem to be using anywhere.

client-v2/src/shared/selectors/__tests__/shared.spec.js
```diff
@@ -0,0 +1,141 @@
...
+      const selector = createArticleFromArticleFragmentSelector();
+      expect(selector(state, 'invalid')).toEqual(null);
+    });
+    it('should handle a case where the ExternalArticle referenced by the ArticleFragment does not exist', () => {
+      const selector = createArticleFromArticleFragmentSelector();
+      expect(selector(state, 'afWithInvalidReference')).toEqual(null);
+    });
+  });
+
+  describe('createArticlesInCollectionGroupSelector', () => {
```
Reettaphant: We are missing a test for when the meta block is entirely missing from the article.

client-v2/src/selectors/frontsSelectors.js
```diff
@@ -48,4 +70,72 @@ const frontsIdSelector = createSelector([getFronts], (fronts): string[] => {
...
+      ) {
+        return [...acc, fronts[key]];
+      }
+      return acc;
+    },
+    []
+  );
+  return {
+    fronts: frontsWithPriority,
+    collections
```
RichieAHB: Is the scope of this selector too big? It just passes through `collections` from argument to return. Could we not split this into two selectors? This is now doing the same as `getFrontsWithPriority` and `getCollections` combined?

client-v2/src/shared/components/Collection.js
```diff
@@ -0,0 +1,74 @@
...
+`;
+
+const collectionDetail = ({ collection, stage }: Props) =>
+  collection ? (
+    <CollectionContainer>
+      <CollectionHeadline>{collection.displayName}</CollectionHeadline>
+      {collection.groups
+        ? collection.groups
+            .slice()
+            .reverse()
```
RichieAHB: Is there any reason this doesn't come reversed from the selector (there may will be!)

client-v2/src/types/FaciaApi.js
jonathonherbert: Agreed 👍

client-v2/src/actions/ExternalArticles.js
jonathonherbert: Good question! I've ported any error handling from the original project here; there's definitely another conversation to have about error handling in a more generic sense, which hopefully will be another PR in the not-too-distant future, if that makes sense.

RichieAHB: 👍

client-v2/src/shared/components/Collection.js
jonathonherbert: Nope, we could definitely do this earlier. Arguably at the point at which we insert the model into the state.

client-v2/src/selectors/frontsSelectors.js
jonathonherbert: This is a pretty lazy port of logic that (IIRC) previously lived in a render function somewhere, and I haven't taken the time to refactor it; arguably this belongs in a separate PR, but I'm happy to do that here also if needed.

client-v2/src/shared/selectors/__tests__/shared.spec.js
jonathonherbert: Excellent point! Also a test for the idea of articles without a 'meta' key belonging to group 0. I'll add both.

client-v2/src/actions/FrontsConfig.js
jonathonherbert: Good point - our supported browsers as indicated in babel.config.js are fine with this, but I'll add missing imports for babel-polyfill, so future use of unsupported ESx-y things should be polyfilled as needed.

client-v2/src/actions/FrontsConfig.js
jonathonherbert: (me and Rich discussed this - we're happy it's necessary to do the necessary munging before our final collection type is added to the app state)

PR #327, opened 2018-05-22
add a readme to the new shared library

Reettaphant:
This leaves out quite a few things (e.g. error handling, making ands saving edits) which we can add to it once the project progresses.

### DIFF
--- a/client-v2/src/shared/README.md
+++ b/client-v2/src/shared/README.md
@@ -1,0 +1,91 @@
+# Shared Collections Library
+(Still waiting for a better name for this)
+
+The shared collections library is a library written using React and Redux and flow for typing.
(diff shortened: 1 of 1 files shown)


client-v2/src/shared/README.md
```diff
@@ -0,0 +1,90 @@
+# Shared Collections Library
+(Still waiting for a better name for this)
+
+The shared collections library is a library written using React and Redux and flow for typing.
+It makes it easier to develop tools which need to provide the ability to curate collections of content.
+
+The shared collections library provides:
+- Components which render collections
+- Reducers and actions to store articles and collections in application state.
```
jonathonherbert: ... and selectors for common selection use cases, e.g. selecting all articles in a particular group for a given collection.

PR #330, opened 2018-05-25
Display draft and take down labels

Reettaphant:
To accomplish replicating the current behaviour of the fronts tool: 

- We fetch all external articles from preview capi
- We also fetch `isLive` and `firstPublicationDate` fields from capi
- We use this info to determine if we need to display draft or taken down labels

This approach is not entirely correct as we might end up displaying unpublished changes in live articles without flagging this. We need to speak with ux and find out if we can keep the old, slightly faulty logic in the tool in the future. Should be ok for now though!

### DIFF
--- a/client-v2/src/shared/components/Article.js
+++ b/client-v2/src/shared/components/Article.js
@@ -28,6 +28,11 @@ const ArticleComponent = ({ article }: ComponentProps) =>
   article && (
     <ArticleContainer key={article.headline}>
       {article.headline}
+      <div>
+        {!article.isLive &&
+          (article.firstPublicationDate ? 'Taken Down' : 'Draft')}
+      </div>
+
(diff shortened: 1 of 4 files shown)


client-v2/src/shared/components/Article.js
```diff
@@ -28,6 +28,8 @@ const ArticleComponent = ({ article }: ComponentProps) =>
   article && (
     <ArticleContainer key={article.headline}>
       {article.headline}
+      {!article.isLive && !article.firstPublicationDate && <div>Draft</div>}
```
RichieAHB: Just a slight tweak to remove some redundancy?
```js
<div>
  {!article.isLive && (article.firstPublicationDate ? 'Taken Down' : 'Draft')}
</div>
```

PR #334, opened 2018-05-31
display error message on error

akash1810:
When we don't get a 2XX back from the server, we log to sentry and then display a success modal. This is a little confusing.

Now, we display the error message and stop.

I've also removed some seemingly random `.catch`es.


akash1810:
[Sending a second breaking news alert for the same story (e.g. live-blog) if you delete the story from the bundle and re-add it](https://trello.com/c/SHpQm2rv/414-sending-a-second-breaking-news-alert-for-the-same-story-eg-live-blog-if-you-delete-the-story-from-the-bundle-and-re-add-it)

public/src/js/models/collections/collection.js
```diff
@@ -234,29 +236,30 @@ export default class Collection extends BaseClass {
...
-        .then(() => this.load().then(detectPressFailures))
-        .then(() => {
-            if (opts.sendAlert) {
-                success(requestData);
-            }
-        })
-        .catch(() => {});
+        .catch(err => {
+            this.setPending(false);
+            alert(err.message);
```
Reettaphant: I think it would be good to have a custom client facing text here without any of the errors which are coming back from the server. Breaking news collections should have a different message from normal collections as well from normal collections just to make it clear what is happening. Calling `confirmSendingAlert()` will tell you if we are dealing with a breaking news collection.

akash1810: Updated now. In two minds about whether the server response should go to sentry or not, I think its useful if it does.

Reettaphant: I think it might as well although we log it on server side as well

PR #335, opened 2018-06-04
Also on warnings

Reettaphant:
The look of the also on warnings isn't great since we are still waiting for input from UX. This also fixes a small bug we have in the current tool where we can't click on the alsoOn front links if the front we are navigating to has a different priority from the current front.


jonathonherbert:
Looks great!

PR #338, opened 2018-06-06
Add 'last modified' for fronts

jonathonherbert:
Pings off a request to front/lastmodified, and adds an action, reducer handler and selector. Displays the resulting data (no fancy styling!) in the appropriate component.


client-v2/src/components/FrontsEdit/Fronts.js
```diff
@@ -50,6 +58,9 @@ class Fronts extends React.Component<FrontsComponentProps, ComponentState> {
         nextProps.selectedFront.collections
       );
     }
+    if (this.props.frontId !== nextProps.frontId || !this.props.lastPressed) {
+      this.props.frontsActions.fetchLastPressed(nextProps.frontId);
```
Reettaphant: Should we be polling this endpoint?

jonathonherbert: ... yes. Yes we should 😁

Reettaphant: Let's do this in a separate pr as discussed.

PR #339, opened 2018-06-07
New collection endpoints

Reettaphant:
New endpoint now prunes unnecessary metadata on save. We don't need to worry about the weird live/draft format behaviour if we don't allow edits to live collections - the publish endpoint will take care of all of this.

### DIFF
--- a/app/controllers/FaciaToolV2Controller.scala
+++ b/app/controllers/FaciaToolV2Controller.scala
@@ -33,17 +33,15 @@ class FaciaToolV2Controller(
       case Some(update) => {
           val identity = request.user
 
-          val shouldUpdateLive: Boolean = update.isLive
-
           updateActions.v2UpdateCollection(update.id, update.collection, identity)
 
           faciaPress.press(PressCommand(
             Set(update.id),
-            live = update.isLive,
-            draft = (update.collection.draft.isEmpty && shouldUpdateLive) || !shouldUpdateLive)
-          )
+            live = false,
+            draft = true
+          ))
 
-          structuredLogger.putLog(LogUpdate(V2CollectionUpdate(update.id, update.updateType), identity.email))
+          structuredLogger.putLog(LogUpdate(V2CollectionUpdate(update.id, update.collection), identity.email))
 
           Ok(Json.toJson(update.collection)).as("application/json")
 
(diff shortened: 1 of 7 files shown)


app/controllers/FaciaToolV2Controller.scala
```diff
@@ -0,0 +1,54 @@
...
+    val v2Update: Option[V2Update] = request.body.asJson.flatMap(jsValue => jsValue.asOpt[V2Update])
+    v2Update match {
+      case Some(update) => {
+          val identity = request.user
+
+          updateActions.v2UpdateCollection(update.id, update.collection, identity)
+
+          faciaPress.press(PressCommand(
+            Set(update.id),
+            live = false,
```
RichieAHB: Is this the bit where we're ignoring the weird live front editing from v1?

app/model/frontsapi.scala
```diff
@@ -126,11 +130,27 @@ trait UpdateActionsTrait {
...
+    }
+
   def putMasterConfig(config: ConfigJson, identity: User): Option[ConfigJson] = {
     faciaApiIO.archiveMasterConfig(config, identity)
     faciaApiIO.putMasterConfig(config)
   }
 
+  def v2UpdateCollection(id: String, collection: CollectionJson, identity: User): CollectionJson = {
+    val collectionWithGroupsRemoved = removeGroupIfNoLongerGrouped(id, collection)
+    val prunedCollection = pruneBlock(collectionWithGroupsRemoved)
```
RichieAHB: What does this line actually prune in reality? What sort of things are removed (or more to the point, what might we put here on the client that isn't wanted)?

app/controllers/FaciaToolV2Controller.scala
Reettaphant: Yes

app/model/frontsapi.scala
Reettaphant: This removes empty metadata objects and removes empty  group objects if the collection is no longer grouped. The only thing I could think of why a collection is not grouped anymore is if someone changes the layout. These transformations are not really changing anything in the collection, just cleaning up empty stuff.

RichieAHB: 👍 makes perfect sense

Reettaphant:
Commits were cherry-picked to another branch

PR #340, opened 2018-06-07
Add drag and drop for collections

RichieAHB:
This brings in [https://github.com/RichieAHB/guration](https://github.com/RichieAHB/guration) to handle validation, of dragging and dropping into a tree (I'll need to move it over to the guardian github / npm but was quicker to set up on mine)(also we _can_ change the name 🙄 ).

This is definitely still a work in progress, styles are odd and things like snaplinks and dropping articles is missing but there is a clear way forward for this. I also don't want to keep rebasing!

I've added some actions to shared (add and remove ones mainly) so have a look there and see whether they fit with the model you guys had in mind. I think it maybe easier to talk about much of this IRL ...

As a side note we don't have a type for the tree that we get returned but will add this in a follow up PR.


client-v2/src/shared/actions/ArticleFragments.js
```diff
@@ -12,4 +12,47 @@ function articleFragmentsReceived(articleFragments: {
...
+      id,
+      supportingArticleFragmentId,
+      index
+    }
+  };
+}
+
+function changeArticleGroup(id: string, newGroup: string): Action {
+  return {
+    type: 'SHARED/CHANGE_ARTICLE_GROUP',
```
jonathonherbert: <nitpick>CHANGE_ARTICLE_FRAGMENT_GROUP?</nitpick> :)

RichieAHB: This is / was a good point and will change.

PR #341, opened 2018-06-08
Add a reducer creator to reduce(r) boilerplate

jonathonherbert:
Adds createAsyncResourceBundle, a function to create bundles of selectors, actions, and a reducer for async resources, and applies them to the FrontsConfig, Collection and ExternalArticle resources. 

There's a significant reduction in boilerplate. This is (I hope!) good.

~~Typing in the resulting bundle is slightly weaker than before. This is bad: the resulting actions can't be refined by the 'type' property, because it's provided by string interpolation and not hardcoded. As a result, not only do we have to refine by an arbitrary string property on each action in the reducer, we can't add the resulting types to the Actions union in the consumer module, because other reducers don't refine by this type. This is a tougher problem, and I'd really appreciate any suggestions on this point!~~ No longer the case, thanks for the suggestions below!

I've called the resulting modules 'bundles' and stuck them in 'bundles' folders, but all naming and structural decisions are obviously pretty arbitrary, please let me know if there's something to change here! ('Bundles' is somewhat ambiguous - should it be redux/bundles alongside redux/actions etc.)


RichieAHB:
I've not reviewed this yet but do you think, in order to firm up the typing, we can ask for the action names to be passed in by the consumer? It's definitely a bit more boilerplate I just wonder whether the compromise is worth it? Additionally I know @Ap0c uses flow on contributions and I think I remember him saying he used reducer factories so I wonder what they've done in the face of this issue?

client-v2/src/reducers/frontsReducer.js
```diff
@@ -17,34 +21,38 @@ type State = {
...
-    frontsConfig: { fronts: {}, collections: {} },
+    frontsConfig: initialState,
     lastPressed: {
       draft: {},
       live: {}
     }
   },
   action: Action
 ): State => {
+  const frontsConfig = frontsConfigReducer(state.frontsConfig, action);
```
RichieAHB: Is there any reason this is nested inside this reducer? It seems like it could live outside but maybe I'm missing something?

client-v2/src/shared/components/Collection.js
```diff
@@ -58,17 +56,16 @@ const collectionDetail = ({ collection, stage }: Props) =>
     <span>Waiting for collection</span>
   );
 
-const createMapStateToProps = () => {
```
RichieAHB: How come the change from a factory here?

client-v2/src/shared/util/createAsyncResourceBundle.js
```diff
@@ -0,0 +1,249 @@
...
+) {
+  const { indexById } = options;
+  const selectLocalState = options.selectLocalState
+    ? options.selectLocalState
+    : (state: any) => state[entityName];
+  const baseActionKey = snakeCase(entityName).toUpperCase();
+  const actionKey = options.namespace
+    ? `${options.namespace.toUpperCase()}/${baseActionKey}`
+    : baseActionKey;
+  const FETCH_START = `${actionKey}_FETCH_START`;
```
RichieAHB: Now I'm looking at this obviously you couldn't ask for this as a config as you need the string at compile time to discriminate the union. Will the explicit generic arguments help here when they come to babel?

Ap0c:
@RichieAHB I don't know if this is of any help to you, not being familiar with this codebase, but we've scoped our actions like [this](https://github.com/guardian/support-frontend/blob/master/assets/containerisableComponents/contributionSelection/contributionSelectionActions.js), with the corresponding reducers looking something like [this](https://github.com/guardian/support-frontend/blob/master/assets/containerisableComponents/contributionSelection/contributionSelectionReducer.js#L132). We separated out our action `type` and `scope` into separate properties in the action, so that it can still be tagged by the `type` at compile time. I'm assuming that our `scope` correlates loosely to your `actionKey` here?

RichieAHB:
@Ap0c that is useful thanks! More specifically our issue was around something like:

```js
const createReducerAndActionsAndStuff = (type /* more args */) => {
  // ...
  return {
    // ...
    reducer: (state, action) => {
      switch (action.type) {
        // no type checking :/
        case `${type}_FETCH`: {
          // ...
        }
      }
    }
    actions: {
      fetch: payload => ({ type: `${type}_FETCH`, payload })
    }
  };
}
```
Here we can no longer type check on our action type as it's not know at compile time 😢 

Have you had to deal with this?

jonathonherbert:
@RichieAHB, @Ap0c - thanks for the comments, this is really useful!

 Jamie came by and showed me his solution. Support-frontend use a dynamic 'scope' property to differentiate between modules, and then use the hardcoded 'type' property to differentiate between actions, whereas we do the opposite - use a hardcoded 'scope' property ('localType') and a dynamic 'type' property.

This works fine locally, but as soon as we export the action types into other parts of the app, other reducers can't disambiguate the new actions as the type property is now of type 'string' rather than a union of string literals. @Ap0c your solution solves this issue, so we now have better type safety. Thanks!

The only drawback I can think of - action types aren't module specific, and so looking at a stream of actions in devtools requires checking 'scope' to understand which module triggered the action in question, but I think type safety outweighs that concern 👍

RichieAHB:
@jonathonherbert @Ap0c nice! I wondered if we could do something like that but dismissed it but of course the action types' shape will be the same for every scope so the type checking will work nicely. If I'm getting that wrong then hopefully one of you can correct me tomorrow IRL!

client-v2/src/shared/components/Collection.js
jonathonherbert: selectById doesn't require memoization, it's a lookup :)

client-v2/src/bundles/frontsConfigBundle.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+import createAsyncResourceBundle from 'shared/util/createAsyncResourceBundle';
```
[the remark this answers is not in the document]
jonathonherbert: Good question, 'shared' is obviously heinous and needs correcting ... 'bundle' is more idiomatic but there's definitely room for improvement, super up for discussing!

client-v2/src/reducers/frontsReducer.js
jonathonherbert: The frontsConfig operates on part of the fronts state as a nested reducer, so it needs to operate on the state passed in to the fronts reducer.

client-v2/src/shared/util/createAsyncResourceBundle.js
jonathonherbert: The factor re: @Ap0c 's suggestion should help here!

client-v2/src/bundles/frontsConfigBundle.js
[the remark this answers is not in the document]
RichieAHB: If it's an idiom that I've missed that that makes sense ... happy to stay with it if it's something people will understand 👍

client-v2/src/shared/components/Collection.js
RichieAHB: Simple as that!

client-v2/src/reducers/frontsReducer.js
RichieAHB: I'll chat to you tomorrow about why it's nested. I don't see where the parent reducer touches frontsConfig so it looks a bit tucked away there but I'm sure it's well considered and again, I'm only questioning it because it's not how I would do it - not because it's not a good way of doing it!

client-v2/src/shared/util/createAsyncResourceBundle.js
RichieAHB: Yeah sounds like the two of you sorted that riiight out 🎉

PR #342, opened 2018-06-14
Move npm package over to @guardian

RichieAHB:
Keep things in the faaaamily ...


RichieAHB:
@franziskas Yes - I'll wait for Reetta's big PR to go in as this will be a bit easier to rebase than the other way round I think but hopefully I can do that tomorrow 👌

PR #343, opened 2018-06-19
Launch collections

Reettaphant:
Adds the ability to launch a collection. Launch button is only displayed when we have unpublished changes in a collection.

### DIFF
--- a/client-v2/src/components/Button.js
+++ b/client-v2/src/components/Button.js
@@ -2,11 +2,19 @@
 
 import styled from 'styled-components';
 
+const getColour = ({ selected, dark }) => {
+  if (selected) return '#123';
+  if (dark) {
+    return '#000';
+  }
+  return '#fff';
+};
+
 export default styled(`button`)`
   appearance: none;
   background: ${({ selected }) => (selected ? 'white' : 'transparent')};
   border: 1px solid #fff;
-  color: ${({ selected }) => (selected ? '#213' : '#fff')};
+  color: ${getColour};
 
   :hover {
     background: #fff;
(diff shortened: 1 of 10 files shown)


client-v2/src/services/faciaApi.js
```diff
@@ -65,6 +65,26 @@ async function fetchLastPressed(frontId: string): Promise<string> {
     });
 }
 
+async function publishCollection(collectionId: string): Promise<void> {
+  // The server does not respond with JSON
+  try {
+    await pandaFetch(`/collection/publish/${collectionId}`, {
```
RichieAHB: Can we not just return this promise directly rather than a void promise created by the async function, just in the event that we start returning something useful for the response? Not a deal breaker, just a thought ....

client-v2/src/components/Button.js
```diff
@@ -2,11 +2,19 @@
...
+  }
+  return '#fff';
+};
+
 export default styled(`button`)`
   appearance: none;
   background: ${({ selected }) => (selected ? 'white' : 'transparent')};
   border: 1px solid #fff;
-  color: ${({ selected }) => (selected ? '#213' : '#fff')};
+  color: ${getColour};
```
RichieAHB: Ahhh the fight between English and American English ;)

client-v2/src/services/faciaApi.js
Reettaphant: Yeah, this felt a bit weird but so did returning the promise because I don't use it. I guess we can return it when we actually end up doing something with it.

PR #346, opened 2018-06-27
Add normalisation library

RichieAHB:
# What are you doing?!
Normalising (and coercing) our slightly odd `group` tags on `articleFragments` out to nodes in a tree (the model they are trying to represent) such that our model goes from:

```json
{
  "type": "collection",
  "live": [
    {
      "type": "articleFragment",
      "group": "group-index",
      "meta": {
        "supporting": [
          {
            "type": "articleFragment"
          }
        ]
      }
    }
  ]
}
```

... to:

```json
{
  "type": "collection",
  "live": [
    {
      "type": "group",
      "id": "group-index",
      "articleFragments": [
        {
          "meta": {
            "supporting": [
              {
                "type": "articleFragment"
              }
            ]
          }
        }
      ]
    }
  ]
}
```

To do this we use [this](https://github.com/guardian/normalize-with-fields), which is basically like [this](https://github.com/paularmstrong/normalizr) (the community standard) but with moooaaar features (and currently slightly less error handling).

## What's does this mean?
So now we normalise (and turn groups into nodes) as we come in to the application and then we denormalise (and turn groups into fields) as we send it out. One question is **could this actually be done inside the faciaApi service**, which means any client could work with a slightly more sane model when interacting with the facia tool? So far I haven't done this.

## What's changed
- Selectors (hooray for keeping out lensing in one place!)
- The drag and drop stuff is now simpler
- We now keep the `internal-code/page/` in the articleFragment ids.`internal-code/page/
- A bunch of tests
- A slight fix to the collection saving middleware (works correctly when all actions passed as a batch aren't persistence actions - perhaps not valid but worth accounting for)
- I've also moved over the `guration` package to `@guardian/guration`, which will close #342 
- Remove over 400 loc, even sneaking in a few other fixes 😍 

---

I've changed a fair bit here written by others so have a good long look at that ... but there are no flow errors and all the tests pass so it must be fine, right?!


client-v2/src/shared/bundles/collectionsBundle.js
```diff
@@ -23,101 +16,35 @@ const collectionSelectors = {
...
-            state.articleFragments[fragmentId].meta &&
-            state.articleFragments[fragmentId].meta.supporting &&
-            state.articleFragments[fragmentId].meta.supporting.indexOf(
-              articleFragmentId
-            ) !== -1
-          ) {
+      ['live', 'draft', 'previously'].some(stage => {
+        const groups = state.collections.data[id][stage] || [];
+
+        return groups.some(gId => {
```
susiecoleman: Personally I'd call it `groupID` you're using `collectionId`

client-v2/src/shared/bundles/collectionsBundle.js
```diff
@@ -23,101 +16,35 @@ const collectionSelectors = {
...
+          const articleFragments = state.groups[gId].articleFragments || [];
+          if (articleFragments.indexOf(articleFragmentId) !== -1) {
             collectionId = id;
             return true;
           }
-          return false;
+
+          return articleFragments.some(afId => {
```
susiecoleman: I'd also switch to `articleFragmentId` but the more verbose names are just personal preference. I find them more readable.

client-v2/src/shared/selectors/shared.js
```diff
@@ -105,84 +81,41 @@ const collectionIdsSelector = (
...
-                ...articleFragments[articleFragmentId].meta,
-                supporting: articleFragments[
-                  articleFragmentId
-                ].meta.supporting.map(
-                  (supportingFragmentId: string) =>
-                    articleFragments[supportingFragmentId]
+    (collections, groups, articleFragments, collectionIds, stage) => ({
+      collections: collectionIds
+        .map(
+          cId =>
```
susiecoleman: I prefer `collectionId`

client-v2/src/shared/selectors/shared.js
```diff
@@ -105,84 +81,41 @@ const collectionIdsSelector = (
...
+            collections[cId] && {
+              id: cId,
+              groups: (collections[cId][stage] || []).map(gId => ({
```
susiecoleman: I also prefer `groupId`

client-v2/src/shared/selectors/shared.js
```diff
@@ -105,84 +81,41 @@ const collectionIdsSelector = (
...
+                ...groups[gId],
+                articleFragments: (groups[gId].articleFragments || []).map(
+                  afId => ({
```
susiecoleman: And `articleFragmentId`

client-v2/src/shared/util/schema.js
```diff
@@ -0,0 +1,62 @@
+// @flow
+
+import { createType, build, createFieldType } from 'normalise-with-fields';
+import v4 from 'uuid/v4';
+
+const postProcessArticleFragment = (_af: {
```
susiecoleman: Same `af` vs `articleFragment` comment...

client-v2/src/components/FrontsEdit/CollectionComponents/Collection.js
```diff
@@ -49,15 +38,9 @@ const Collection = ({
...
-    <div style={{ marginLeft: 10 }}>
-      {groups
-        .reduce(getArticleFragmentLengths, [])
-        .map(([group, offset]) => (
-          <React.Fragment key={group.id}>
-            {children(group, offset)}
-          </React.Fragment>
-        ))}
-    </div>
+    <Guration.Level arr={groups} type="group" getKey={({ uuid: key }) => key}>
```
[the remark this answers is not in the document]
RichieAHB: Good question, I thought I'd keep the namespace to make it clear where the library was acting. Would it be clearer without?

client-v2/src/shared/bundles/collectionsBundle.js
RichieAHB: You're definitely (cont ...)

client-v2/src/shared/bundles/collectionsBundle.js
RichieAHB: ... right: will change.

client-v2/src/components/CAPI/SearchQuery.js
```diff
@@ -30,6 +30,8 @@ class SearchQuery extends React.Component<
     params: {}
   };
 
+  state = {};
+
```
Reettaphant: Why in this PR?

client-v2/src/shared/actions/Groups.js
```diff
@@ -0,0 +1,41 @@
+// @flow
```
Reettaphant: I understand why this has been renamed but I also want to use these for the clipboard which don't have groups. Maybe we could talk about `ContainerArticleFragments` to distinguish them from supporting article fragments?

client-v2/src/components/CAPI/SearchQuery.js
RichieAHB: Because I was getting some linting error and I got annoyed and just did it 😞

client-v2/src/shared/actions/Groups.js
RichieAHB: This sounds like a good plan - let's talk IRL 👍

franziskas:
@RichieAHB can we salvage this month-old PR?

PR #351, opened 2018-07-05
Access permission

mbarton:
Adds a switch to check the `fronts_access` user permission before allowing access.

The compile errors are because it requires v2 of the permissions client: https://github.com/guardian/permissions/pull/103.

Once we are happy that the permission is effective I will remove the switch and the google group checker.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -87,7 +87,7 @@ libraryDependencies ++= Seq(
     "com.gu" %% "content-api-models" % capiModelsVersion,
     "com.gu" %% "content-api-models-json" % capiModelsVersion,
     "com.gu" %% "content-api-client-aws" % "0.5",
-    "com.gu" %% "editorial-permissions-client" % "0.8",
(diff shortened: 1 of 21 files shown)


build.sbt
```diff
@@ -87,7 +87,7 @@ libraryDependencies ++= Seq(
     "com.gu" %% "content-api-models" % capiModelsVersion,
     "com.gu" %% "content-api-models-json" % capiModelsVersion,
     "com.gu" %% "content-api-client-aws" % "0.5",
-    "com.gu" %% "editorial-permissions-client" % "0.8",
+    "com.gu" %% "editorial-permissions-client" % "2.0-SNAPSHOT",
```
LATaylor-guardian: Should this be the `SNAPSHOT` version?

mbarton: Yeah this was for local testing, I'll update it once I've published

PR #353, opened 2018-07-09
Add info about types to readme

Reettaphant:
Add some documentation around the types we are using in the shared library. We must remember to keep updating this as we keep developing the libaray.

### DIFF
--- a/client-v2/src/shared/README.md
+++ b/client-v2/src/shared/README.md
@@ -12,35 +12,22 @@ The shared collections library provides:
 The consumer application is responsible for fetching data (collections, details of the articles in the collections) from its own data store and from an external store in (e.g. capi) and passing
 this data on to the shared collections module.
 
+Collections and articles should be passed to the shared collections library in a normalised format. There are utility functions to normalised collections as they are stored in facia-tool or story packages but if the consumer uses a different datastructure, they will have to do perform this normalisation. The `Data types in the library` section contains details of what normalised collections and articles look like.
+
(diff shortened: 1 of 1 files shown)


franziskas:
@jonathonherbert can you please take a look at the proposed changes?

client-v2/src/shared/README.md
```diff
@@ -12,36 +12,23 @@ The shared collections library provides:
 The consumer application is responsible for fetching data (collections, details of the articles in the collections) from its own data store and from an external store in (e.g. capi) and passing
 this data on to the shared collections module.
 
+Collections and articles should be passed to the shared collections library in a normalised. There are utility functions to normalised collections as they are stored in facia-tool or story packages but if the consumer uses a different datastructure, they will have to do perform this normalisation. The `Data types in the library` section contains details of what normalised collections and articles look like.
```
jonathonherbert: 'in a normalised' - is this missing an end, e.g. 'format'?

client-v2/src/shared/README.md
```diff
@@ -12,36 +12,23 @@ The shared collections library provides:
...
-They look like this:
+## Data types in the library
 
-```
-type ExternalArticle = {
-  id: string,
-  headline: string
-};
-```
+There are few data types in this library which represent articles and collections.
```
jonathonherbert: I guess this is 'there are a few', which is a bit pedantic but could potentially be misleading!

PR #356, opened 2018-07-18
Fix fetch config script

susiecoleman:
Config script wasn't creating a `gu` directory in the same place. `/etc` isn't writeable by default so I've added sudo. Adding sudo doesn't feel great. Maybe we should reconsider where config lives on local machines.


franziskas:
@RichieAHB can I bother you to review this? :)

PR #359, opened 2018-07-27
Invalid images

Reettaphant:
This pr makes sure the fronts tool can handle rendering collections containing images which have been deleted from the grid. Previously the fronts tool would fail to render the collection if this happened. Now 
- It renders the collection, with a gray blob instead of a picture. 
- Any article containing a deleted image will not appear on the front so ideally we never want to be in a situation where deleted images have not been removed from fronts. To prevent this, the fronts tool now posts details of every image used on fronts to grid usages so that people can be alerted to this before deleting an image.


app/services/MediaApi.scala
```diff
@@ -28,10 +28,15 @@ class MediaApi(val config: ApplicationConfiguration, val ws: WSClient) {
...
-      case _ => None
+
+    try {
+      val exports = (response \ "data" \ "exports").as[List[Export]]
+      exports.find(_.id == cropId) match {
+        case Some(export) => Some(export.assets.sortBy(asset => asset.size).head.secureUrl)
+        case _ => None
+      }
+    } catch {
+      case e: Exception => None
```
akash1810: Why add this try/catch?

conf/routes
```diff
@@ -85,3 +85,7 @@ GET         /troubleshoot/*section                   controllers.TroubleshootCon
 GET         /v2                                      controllers.V2App.index(path="")
 GET         /v2/*path                                controllers.V2App.index(path)
 
+# Grid Usages
+POST        /api/usage/add                           controllers.GridProxy.addUsage()
```
akash1810: minor: semantically, wouldn't a `POST` will always `add`.

app/services/MediaApi.scala
Reettaphant: If the image has been deleted from the grid, extracting these exports will fail. We still want to render the corresponding collection in the fronts tool, just without an image instead of failing to do so completely.

akash1810: Ah! Would parsing `JsValue` into an `Option[List[Export]]` and mapping over that be more intention revealing? Alternatively, could we add a comment as to why `Exception` will occur?

app/controllers/GridProxy.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.joda.time.DateTime
+import org.joda.time.format.ISODateTimeFormat
+
+object FrontUsageData {
+  implicit val jsonFromat: Format[FrontUsageData] = Json.format[FrontUsageData]
+}
+
+case class FrontUsageData(mediaId: String, front: String)
+
+case class GridUsageData(dateAdded: String, addedBy: String, front: String, mediaId: String)
```
akash1810: Can `dateAdded` be a DateTime 🙏? `String` dates make me really sad

PR #371, opened 2018-08-29
Press alerts

Reettaphant:
Replicates the fronts failing to press alert behaviour with some changes: 
- Because we don't allow edits on live content we only need to check if a front was pressed on launch
- Because users can have multiple fronts open at once, we show them a list of all the fronts which are failed to press after publish.

The beautifully styled alert banner can we restyled and integrated with the app error banner once we get input from ux. For now, this should central prod to start using the application.

### DIFF
--- a/client-v2/src/actions/Fronts.js
+++ b/client-v2/src/actions/Fronts.js
@@ -6,10 +6,12 @@ import { batchActions } from 'redux-batched-actions';
 import {
   fetchFrontsConfig,
   fetchLastPressed as fetchLastPressedApi,
-  publishCollection as publishCollectionApi
+  publishCollection as publishCollectionApi,
+  getCollection
 } from 'services/faciaApi';
 import { actions as frontsConfigActions } from 'bundles/frontsConfigBundle';
 import { recordUnpublishedChanges } from 'actions/UnpublishedChanges';
+import { isFrontStale } from 'util/frontsUtils';
 
 function fetchLastPressedSuccess(frontId: string, datePressed: string): Action {
   return {
@@ -31,11 +33,11 @@ function publishCollectionSuccess(collectionId: string): Action {
   };
 }
 
-function recordStaleFronts(frontId: string, isFrontStale: boolean): Action {
+function recordStaleFronts(frontId: string, frontIsStale: boolean): Action {
   return {
     type: 'RECORD_STALE_FRONTS',
     payload: {
-      frontId: isFrontStale
+      [frontId]: frontIsStale
     }
   };
 }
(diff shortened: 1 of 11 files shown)


client-v2/src/actions/Fronts.js
```diff
@@ -42,17 +53,34 @@ function fetchLastPressed(frontId: string): ThunkAction {
...
         dispatch(
           batchActions([
             publishCollectionSuccess(collectionId),
             recordUnpublishedChanges(collectionId, false)
           ])
-        )
-      )
+        );
+
+        return new Promise(resolve => setTimeout(resolve, 10000))
```
Reettaphant: Not sure if this is the best way to do this, suggestions are welcomed. I copied the value of the timeout from the old app.

client-v2/src/reducers/staleFrontsReducer.js
```diff
@@ -0,0 +1,21 @@
...
+
+type State = {
+  [string]: boolean
+};
+
+const staleFronts = (state: State = {}, action: Action): State => {
+  switch (action.type) {
+    case 'RECORD_STALE_FRONTS': {
+      const { payload } = action;
+      return { ...state, ...payload };
```
RichieAHB: I guess Flow might help us here, but (in tandem with the comment above) spreading directly into the state like this makes me feel irrationally uncomfortable.

client-v2/src/constants/fronts.js
```diff
@@ -6,3 +6,5 @@ export const frontStages = {
   draft: 'draft',
   live: 'live'
 };
+
+export const detectPressFailureMs = 10000;
```
RichieAHB: Is this what we were using in v1?

client-v2/src/actions/Fronts.js
```diff
@@ -31,6 +33,15 @@ function publishCollectionSuccess(collectionId: string): Action {
   };
 }
 
+function recordStaleFronts(frontId: string, frontIsStale: boolean): Action {
+  return {
+    type: 'RECORD_STALE_FRONTS',
+    payload: {
+      [frontId]: frontIsStale
```
RichieAHB: Minor nitpick I feel like `frontId` should be a property on the payload rather than just the key of the object as if we wanted to use this value in a different way in the reduce in future we'd have to use `Object.keys()`.

client-v2/src/actions/Fronts.js
RichieAHB: Could we use the constant declared further down for this value?

RichieAHB: It's sort of a pain that promises aren't cancelable. It looks like this code will just create a promise every time we publish a collection. That is potentially quite infrequently so it should be fine but part of me things we should move towards using something like `lodash.throttle` here, although that would be harder to implement. Maybe that's overkill but if we end up doing something like publishing on edit in some circumstance in future then this may be a memory leak. Be interested to have @jonathonherbert 's eye on this. Happy to let this go as long as we've thought about it and don't mind.

Reettaphant: Having talked to editorial on multiple desks I think it's going to be unlikely that we'll have to ever publish anything on edit so maybe this will be fine, what does @jonathonherbert think?

client-v2/src/constants/fronts.js
Reettaphant: Yes, just copied over most of the logic from v1 as users seem happy with it.

client-v2/src/actions/Fronts.js
RichieAHB: Then I'm happy with this if @jonathonherbert is 👍

PR #372, opened 2018-08-29
Persist front ids

Reettaphant:
Instead of just storing clipboard data in dynamo, we want to store both front ids in there as well. This userdata is sent to the client when the app loads. In the future, other kinds of settings could be added to this store as well. 

Cloudformation changes for this here: https://github.com/guardian/editorial-tools-platform/pull/229

### DIFF
--- a/app/controllers/UserDataController.scala
+++ b/app/controllers/UserDataController.scala
@@ -1,17 +1,12 @@
 package controllers
 
-import com.gu.facia.api.NotFound
 import com.gu.facia.client.models.Trail
 import com.gu.scanamo._
 import com.gu.scanamo.syntax._
-import play.api.libs.json._
 import services.Dynamo
 import model.UserData
 
-import scala.util.{Failure, Success, Try}
-
-
-class ClipboardController(dynamo: Dynamo, val deps: BaseFaciaControllerComponents) extends BaseFaciaController(deps) {
+class UserDataController(dynamo: Dynamo, val deps: BaseFaciaControllerComponents) extends BaseFaciaController(deps) {
   import model.UserData._
 
   private lazy val userDataTable = Table[UserData](config.faciatool.userDataTable)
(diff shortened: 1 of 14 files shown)


client-v2/src/actions/Clipboard.js
```diff
@@ -52,43 +51,39 @@ function fetchClipboardContentSuccess(clipboardContent: Array<string>) {
   };
 }
 
-function fetchClipboardContent(): ThunkAction {
```
RichieAHB: I think I need to understand this change a bit more before I can review this. How come we're not fetching the clipboard any more - where is being loaded from now?

jonathonherbert: I _think_ the idea is because Scala is in a position to provide us with this information on our first request, we plant it on window.config and parse it when the app starts up.

app/controllers/UserDataController.scala
```diff
@@ -0,0 +1,46 @@
...
+import com.gu.facia.client.models.Trail
+import com.gu.scanamo._
+import com.gu.scanamo.syntax._
+import services.Dynamo
+import model.UserData
+
+class UserDataController(dynamo: Dynamo, val deps: BaseFaciaControllerComponents) extends BaseFaciaController(deps) {
+  import model.UserData._
+
+  private lazy val userDataTable = Table[UserData](config.faciatool.userDataTable)
```
jonathonherbert: Not for now, but would it be worth thinning out our controllers by creating a UserDataDAO/UserDataService or equivalent and delegating persistence duties to that, especially as our persistence requirements for user data grow?

client-v2/src/actions/Clipboard.js
RichieAHB: I thought as much but I just couldn't see where that was happening but maybe it's already happening before this PR?

jonathonherbert: I think that's right! IIRC it happens on the Config instance dropped into the index template, which is extended in this PR to include our options.

app/controllers/UserDataController.scala
Reettaphant: Yep, I thought of this as well and I thought when we get a third type of user data it will be worth doing it.

client-v2/src/actions/Clipboard.js
RichieAHB: Cool - this all make much more sense then fankew 👍

PR #376, opened 2018-09-04
Add more granular permissions

Reettaphant:
Add permissions to the fronts tool to lock to the tool down by front priority (commercial, training, editorial). We lock down the front edit endpoints and collection edit endpoints. Because collections can be shared between fronts of different priorities, the user will be allowed to edit a collection if they have the permission to edit at least one of the front priorities that the collection belongs to.

These new permissions will initially be behind switches and will not immediately be operational when we merge the pr. Wanting to have new permissions behind switches has meant a little bit of code repetition here because we've had to keep old permissions (breaking news and config permissions) separate from the new ones.

### DIFF
--- a/app/controllers/FaciaToolController.scala
+++ b/app/controllers/FaciaToolController.scala
@@ -42,7 +42,7 @@ class FaciaToolController(
 
   def getCollection(collectionId: String) = AccessAPIAuthAction.async { implicit request =>
 
-    val collectionPriorities = getPriorityFilterFromCollectionId(collectionId)
+    val collectionPriorities = configAgent.getFrontsPermissionsPriorityByCollectionId(collectionId)
 
     withModifyGroupPermissionForCollections(collectionPriorities, Set()) {
 
@@ -61,16 +61,11 @@ class FaciaToolController(
     }
   }
 
-  def getPriorityFilterFromCollectionId(collectionId: String) = {
-    val result: Set[PermissionsPriority] = configAgent.getFrontsPermissionsPriorityByCollectionId(collectionId)
-    result
-  }
-
   def publishCollection(collectionId: String) = AccessAPIAuthAction.async { implicit request =>
 
     withModifyPermissionForCollections(Set(collectionId)) {
 
-      val collectionPriorities = getPriorityFilterFromCollectionId(collectionId)
+      val collectionPriorities = configAgent.getFrontsPermissionsPriorityByCollectionId(collectionId)
 
       withModifyGroupPermissionForCollections(collectionPriorities, Set(), true) {
         val identity = request.user
@@ -120,7 +115,7 @@ class FaciaToolController(
 
   def treatEdits(collectionId: String) = AccessAPIAuthAction.async { implicit request =>
 
-    val collectionPriorities = getPriorityFilterFromCollectionId(collectionId)
+    val collectionPriorities = configAgent.getFrontsPermissionsPriorityByCollectionId(collectionId)
 
     withModifyGroupPermissionForCollections(collectionPriorities, Set()) {
 
@@ -167,7 +162,7 @@ class FaciaToolController(
         case update: Update =>
           withModifyPermissionForCollections(Set(update.update.id)) {
 
-            val collectionPriorities = getPriorityFilterFromCollectionId(update.update.id)
+            val collectionPriorities = configAgent.getFrontsPermissionsPriorityByCollectionId(update.update.id)
 
             withModifyGroupPermissionForCollections(collectionPriorities, Set()) {
 
@@ -198,7 +193,7 @@ class FaciaToolController(
         case remove: Remove =>
           withModifyPermissionForCollections(Set(remove.remove.id)) {
 
-            val collectionPriorities = getPriorityFilterFromCollectionId(remove.remove.id)
+            val collectionPriorities = configAgent.getFrontsPermissionsPriorityByCollectionId(remove.remove.id)
 
             withModifyGroupPermissionForCollections(collectionPriorities, Set()) {
 
@@ -220,8 +215,8 @@ class FaciaToolController(
         case updateAndRemove: UpdateAndRemove =>
           withModifyPermissionForCollections(Set(updateAndRemove.update.id, updateAndRemove.remove.id)) {
 
-            val fromCollectionPriorities = getPriorityFilterFromCollectionId(updateAndRemove.update.id)
-            val toCollectionPriorities = getPriorityFilterFromCollectionId(updateAndRemove.remove.id)
+            val fromCollectionPriorities = configAgent.getFrontsPermissionsPriorityByCollectionId(updateAndRemove.update.id)
+            val toCollectionPriorities = configAgent.getFrontsPermissionsPriorityByCollectionId(updateAndRemove.remove.id)
 
             withModifyGroupPermissionForCollections(fromCollectionPriorities, toCollectionPriorities) {
               val identity = request.user
--- a/app/controllers/FaciaToolController.scala
+++ b/app/controllers/FaciaToolController.scala
@@ -32,7 +32,7 @@ class FaciaToolController(
                            val mediaServiceClient: MediaServiceClient,
                            val deps: BaseFaciaControllerComponents
                          )(implicit ec: ExecutionContext)
-  extends BaseFaciaController(deps) with BreakingNewsEditCollectionsCheck with Logging {
+  extends BaseFaciaController(deps) with BreakingNewsEditCollectionsCheck with ModifyCollectionsPermissionsCheck with Logging {
 
   def getConfig = AccessAPIAuthAction.async { request =>
     FaciaToolMetrics.ApiUsageCount.increment()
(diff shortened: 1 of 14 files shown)


app/permissions/PermissionsActionCheck.scala
```diff
@@ -27,6 +27,56 @@ trait PermissionActionFilter extends ActionFilter[UserRequest] with Logging {
...
+
+class LaunchCommercialFrontsPermissionCheck(val acl: Acl)(implicit ec: ExecutionContext) extends PermissionActionFilter {
+  override implicit val executionContext: ExecutionContext = ec
+  override val testAccess: String => Authorization = acl.testUser(Permissions.LaunchCommercialFrontsAlert, "facia-tool-allow-access-for-all")
+  override val restrictedAction: String = "Launch commercial fronts."
+}
+
+class LaunchEditorialFrontsPermissionCheck(val acl: Acl)(implicit ec: ExecutionContext) extends PermissionActionFilter {
+  override implicit val executionContext: ExecutionContext = ec
+  override val testAccess: String => Authorization = acl.testUser(Permissions.LaunchEditorialFrontsAlert, "facia-tool-allow-access-for-all")
```
akash1810: Is `facia-tool-allow-access-for-all` correct?

Reettaphant: They are not. I will update them

app/permissions/PermissionsActionCheck.scala
```diff
@@ -27,6 +27,56 @@ trait PermissionActionFilter extends ActionFilter[UserRequest] with Logging {
...
+
+    if (priorities.isEmpty) AccessGranted
+    else if (isLaunch)
+      acl.testUserGroupsAndCollections(
+        Permissions.LaunchEditorialFrontsAlert, Permissions.LaunchCommercialFrontsAlert,
+        Permissions.FrontsAccess, "facia-tool-allow-access-for-all")(email, priorities)
+    else
+      acl.testUserGroupsAndCollections(
+        Permissions.EditEditorialFrontsAlert, Permissions.LaunchCommercialFrontsAlert,
+        Permissions.FrontsAccess, "facia-tool-allow-access-for-all")(email, priorities)
```
Reettaphant: We need to add the correct switch name here as well, don't we @susiecoleman ? It's going to be messy because we are going to have to figure out the correct switch name from the priority names.

app/util/Acl.scala
```diff
@@ -47,4 +48,49 @@ class Acl(permissions: PermissionsProvider) extends Logging {
...
+             hasEditorialPermissions: Authorization,
+             hasTrainingPermissions: Authorization,
+             priorities: Set[PermissionsPriority]
+           ): Authorization =  {
+
+    val trainingPermissionIsValid = priorities.contains(TrainingPermission)
+    val editorialPermissionIsValid = priorities.contains(EditorialPermission) || priorities.contains(EmailPermission)
+    val commercialPermissionIsValid = priorities.contains(CommercialPermission)
+
+    if (trainingPermissionIsValid)
```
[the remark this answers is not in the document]
Reettaphant: I wrote tests for this to test it and also so document what we are expecting to be returned. I'm happy to accept any ideas for improving the logic.

app/permissions/PermissionsActionCheck.scala
susiecoleman: I think something might have got lost in a merge? As this is right on my machine. These should be `"facia-tool-allow-launch-editorial-fronts-for-all"` and `"facia-tool-allow-edit-editorial-fronts-for-all"` respectively. I'll push up the changes

app/permissions/PermissionsActionCheck.scala
```diff
@@ -27,6 +27,56 @@ trait PermissionActionFilter extends ActionFilter[UserRequest] with Logging {
...
+
+  def withModifyGroupPermissionForCollections[A](priorities: Set[PermissionsPriority], secondaryPriorities: Set[PermissionsPriority],
+                                                 isLaunch: Boolean = false)(block: => Future[Result])
+                                                (implicit request: UserRequest[A],
+                                                 executionContext: ExecutionContext): Future[Result] = {
+
+    (testAccess(request.user.email, priorities, isLaunch), testAccess(request.user.email, secondaryPriorities, isLaunch)) match {
+      case (AccessGranted, AccessGranted) => block
+      case _ => {
+        logger.info(s"User with e-mail ${request.user.email} not authorized to modify collection")
```
mbarton: I think it would help debugging to log this at `warn` and include what front types (priorities) they wanted to edit and what permissions they actually help.

app/permissions/PermissionsActionCheck.scala
```diff
@@ -27,6 +27,56 @@ trait PermissionActionFilter extends ActionFilter[UserRequest] with Logging {
...
+      acl.testUserGroupsAndCollections(
+        Permissions.LaunchEditorialFrontsAlert, Permissions.LaunchCommercialFrontsAlert,
+        Permissions.FrontsAccess, "facia-tool-allow-launch-editorial-fronts-for-all")(email, priorities)
+    else
+      acl.testUserGroupsAndCollections(
+        Permissions.EditEditorialFrontsAlert, Permissions.LaunchCommercialFrontsAlert,
+        Permissions.FrontsAccess, "facia-tool-allow-edit-editorial-fronts-for-all")(email, priorities)
+  }
+
+  def withModifyGroupPermissionForCollections[A](priorities: Set[PermissionsPriority], secondaryPriorities: Set[PermissionsPriority],
```
mbarton: Maybe define the check as an `ActionFilter` in the style of `PermissionsAccessCheck`? That would avoid an extra level of nesting in the calling code.

I think we'd need to pass in `configAgent` to the trait, maybe as a def that's then overridden in each controller.

app/util/Acl.scala
```diff
@@ -47,4 +48,49 @@ class Acl(permissions: PermissionsProvider) extends Logging {
       testUser(permission, switch)(email)
     else AccessGranted
   }
+
+  def testUserGroupsAndCollections(editorialPermission: PermissionDefinition, commercialPermission: PermissionDefinition,
```
mbarton: I'd just reference the permissions directly rather than passing them as parameters, since the variable names already match up to specific ones

test/util/AclTest.scala
```diff
@@ -0,0 +1,48 @@
...
+import org.scalatest.{DoNotDiscover, FreeSpec, Matchers}
+import permissions.{TrainingPermission, EmailPermission, EditorialPermission, CommercialPermission}
+
+@DoNotDiscover class AclTest extends FreeSpec with Matchers {
+
+  "Grant access according to collection priorities" - {
+
+    "Allow access to commercial fronts if commercial permissions" in {
+
+      PermissionsChecker.check(AccessGranted, AccessDenied, AccessDenied, Set(CommercialPermission)) should be(AccessGranted)
```
mbarton: It'll get a bit verbose but I think it's worth naming parameters in each test to make them easier to follow.

👍 for the tests, great stuff

app/services/ConfigAgentTrait.scala
```diff
@@ -54,4 +55,17 @@ class ConfigAgent(val config: ApplicationConfiguration, val frontsApi: FrontsApi
 
   def getConfig(id: String): Option[CollectionConfig] = configAgent.get().flatMap(_.collections.get(id).map(CollectionConfig.fromCollectionJson))
 
+  def getFrontsPermissionsPriorityByCollectionId(id: String): Set[PermissionsPriority] = configAgent.get().map(config => {
+    val fronts = config.fronts
+    fronts.foldLeft(Set[PermissionsPriority]())((acc, front) => {
```
mbarton: I think you can destructure in the lambda arguments to avoid using `_2`:

```scala
fronts.foldLeft(Set.empty[PermissionsPriority]) { case(acc, (_, front)) =>
    if(front.collections.contains(id)) {
       ...
    }
}
```

app/controllers/FaciaToolController.scala
```diff
@@ -31,39 +32,56 @@ class FaciaToolController(
                            val mediaServiceClient: MediaServiceClient,
                            val deps: BaseFaciaControllerComponents
                          )(implicit ec: ExecutionContext)
-  extends BaseFaciaController(deps) with BreakingNewsEditCollectionsCheck with Logging {
+  extends BaseFaciaController(deps) with BreakingNewsEditCollectionsCheck with ModifyCollectionsPermissionsCheck with Logging {
```
mbarton: See comment below (maybe we can use an `ActionBuilder` here). It would reduce the size of the diff to hopefully just the bits that affect permission checks.

app/util/Acl.scala
Reettaphant: We need to pass in the editorial permissions here because it can be one of two permissions (either just edit or edit and launch). We only have one kind of permission for commercial and training so they don't have to be passed in here but I think the symmetry makes it easier to read.

app/permissions/PermissionsActionCheck.scala
Reettaphant: This is difficult to do because we don't actually know which permissions to check when we hit the endpoint so I think we should leave it as it as discussed.

app/permissions/PermissionsActionCheck.scala
```diff
@@ -27,6 +27,56 @@ trait PermissionActionFilter extends ActionFilter[UserRequest] with Logging {
...
+                                                 isLaunch: Boolean = false)(block: => Future[Result])
+                                                (implicit request: UserRequest[A],
+                                                 executionContext: ExecutionContext): Future[Result] = {
+
+    (testAccess(request.user.email, priorities, isLaunch), testAccess(request.user.email, secondaryPriorities, isLaunch)) match {
+      case (AccessGranted, AccessGranted) => block
+      case _ => {
+        logger.warn(s"User with e-mail ${request.user.email} not authorized to modify collection with priorities $priorities")
```
mbarton: Can we also include the permissions that the user does have? That has helped me debug issues before

PR #379, opened 2018-09-10
Remove google group checker

mbarton:
It's been unused since the move to permissions (#351)


mbarton:
Yeah definitely. I've removed it from the code in this PR (I think?) and will remove it from the file in S3 once deployed

PR #381, opened 2018-09-14
Search articles

Reettaphant:
@jonathonherbert @RichieAHB This is now done and ready for review!

Allows for searching capi by tag and by section. It's missing pagination still but I'm going to add that in a separate pr. The menu icons in the bottom are also missing but those are not urgent and I have a card on trello for adding those later.

It looks a bit different from @Joshua-David's designs, we've had a chat about this and decided that having chosen tags displayed in the input field is not going to work. Josh is going to continue working on the designs for the tags but this should be enough for now to roll it out to central prod. 

screen shot 2018-10-03 at 09 25 21
screen shot 2018-10-03 at 09 25 39

### DIFF
--- a/client-v2/src/components/FrontsCAPIInterface/SearchInput.js
+++ b/client-v2/src/components/FrontsCAPIInterface/SearchInput.js
@@ -10,27 +10,28 @@ import CAPITagInput from '../FrontsCAPIInterface/TagInput';
 
 type FrontsCAPISearchInputProps = {
   children: *,
-  additionalFixedContent?: React.ComponentType<any>
+  additionalFixedContent?: React.ComponentType<any>,
+  displaySearchFilters: boolean,
+  updateDisplaySearchFilters: (value: boolean) => void
 };
 
 type FrontsCAPISearchInputState = {
   q: ?string,
   tags: Array<string>,
   sections: Array<string>,
-  dislaySearchFilters: boolean,
   tagsSearchTerm: '',
   sectionsSearchTerm: ''
 };
 
 const InputContainer = styled('div')`
   margin-bottom: 20px;
+  background: #ffffff;
 `;
 
 const emptyState = {
   q: null,
   tags: [],
   sections: [],
-  displaySearchFilters: true,
   tagsSearchTerm: '',
   sectionsSearchTerm: ''
 };
@@ -46,18 +47,18 @@ class FrontsCAPISearchInput extends React.Component<
       q: null,
       tags: [],
       sections: [],
-      displaySearchFilters: false,
       tagsSearchTerm: '',
       sectionsSearchTerm: ''
     });
+    this.props.updateDisplaySearchFilters(false);
   };
 
   searchInput = () => {
     this.setState({
-      displaySearchFilters: false,
       tagsSearchTerm: '',
       sectionsSearchTerm: ''
     });
+    this.props.updateDisplaySearchFilters(false);
   };
 
   clearIndividualSearchTerm = (searchTerm: string) => {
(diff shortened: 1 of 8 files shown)


client-v2/src/components/FrontsCAPIInterface/TagInput.js
```diff
@@ -4,43 +4,73 @@ import * as React from 'react';
...
+  padding: 2px;
+`;
+
 const CAPITagInput = <T>({ onChange, placeholder }: CAPITagInputProps<T>) => (
   <Downshift
     itemToString={item => (item ? item.id : '')}
-    onChange={onChange}
+    onChange={value => {
+      onChange(value);
+      value = null;
```
LATaylor-guardian: Why do we set `value = null` here?

client-v2/src/components/FrontsCAPIInterface/SearchInput.js
```diff
@@ -6,15 +6,19 @@ import styled from 'styled-components';
...
   children: *,
-  additionalFixedContent?: React.ComponentType<any>
+  additionalFixedContent?: React.ComponentType<any>,
+  displaySearchFilters: boolean
 };
 
 type FrontsCAPISearchInputState = {
   q: ?string,
-  tag: ?string
+  tags: Array<string>,
```
LATaylor-guardian: Is this an array of tag ids?

Reettaphant: Yes. We use this array to query capi for content with these tags.

client-v2/src/components/FrontsCAPIInterface/SearchInput.js
```diff
@@ -42,47 +89,140 @@ class FrontsCAPISearchInput extends React.Component<
     });
   };
 
-  handleTagInput = (item: any) => {
+  handleTagSearchInput = (
+    { currentTarget }: SyntheticEvent<HTMLInputElement>,
+    type: string
+  ) => {
+    const termInState = `${type}SearchTerm`;
```
RichieAHB: Could we change the name spacing in the state to `{ searchTerms: { tags: '', sections: '' } } to make this a bit clearer rather than having to create this string?

client-v2/src/components/FrontsCAPIInterface/SearchInput.js
```diff
@@ -42,47 +89,140 @@ class FrontsCAPISearchInput extends React.Component<
...
     this.setState({
-      tag: item ? item.id : null
+      [termInState]: currentTarget.value
     });
   };
 
+  handleTagInput = (item: any, type: string) => {
+    const searchTerm = `${type}SearchTerm`;
+    const newTags = this.state[type];
```
RichieAHB: `newTags` is being mutated here and when it goes back in the state the array is the same as when it came out, even if new stuff has been added. Perhaps you could use `let newTags` here and `newTags = newTags.concat(...)` below to make sure that if we use `===` to check for equality with this array somewhere, that it'll work as expected.

client-v2/src/components/FrontsCAPIInterface/SearchInput.js
```diff
@@ -42,47 +89,140 @@ class FrontsCAPISearchInput extends React.Component<
...
+      q,
+      tagsSearchTerm,
+      sectionsSearchTerm
+    } = this.state;
+
+    const searchTermsExist = tags.length !== 0 || sections.length !== 0 || !!q;
+    const tagQuery = tags ? tags.join(',') : '';
+    const sectionQuery = sections ? sections.join(',') : '';
+
+    if (!displaySearchFilters) {
```
RichieAHB: Is there any way we could merge these two cases a bit? If the `displaySearchFilters` check is done a few times at a more fine grained level it might mean we don't have to remember to change the props on things like `TextInput` in two places if they need changing in future?

client-v2/src/components/FrontsCAPIInterface/TagInput.js
```diff
@@ -51,57 +82,62 @@ const CAPITagInput = <T>({ onChange, placeholder }: CAPITagInputProps<T>) => (
...
-                /**
-                 * Filter tags based on ids
-                 * think this is only for the test/test tag but sometime tags
-                 * come
-                 */
-                const seenIds = [];
+              /**
+               * Filter tags based on ids
+               * think this is only for the test/test tag but sometime tags
+               * come
```
RichieAHB: I wrote this and it meant to say "come through more than once" - would you mind adding it on to the end of the comment!? Sorry!

RichieAHB:
@Reettaphant also is this still do-not-merge?

client-v2/src/components/FrontsCAPIInterface/SearchInput.js
Reettaphant: This is pretty ugly but I've been struggling to find a nice simple solution to this. We are already saving the actual selected tags and sections as `tags` and `sections` in state but we need to also keep track of the search terms we use to search capi for tags. I guess we could save these in a separate object in the state? 

```
{ 
  tags: [], 
  sections: [], 
  searchTerms: {
    tags: '', 
    sections
  }
};

```

client-v2/src/components/FrontsCAPIInterface/SearchInput.js
Reettaphant: Yes! The second `displaySearchFilters` check was redundant.

client-v2/src/components/FrontsCAPIInterface/SearchInput.js
RichieAHB: I think that would make it a touch easier to read for me - it just seems like the more robust way to namespace it?

client-v2/src/components/FrontsCAPIInterface/TagInput.js
RichieAHB: Thanks!

client-v2/src/components/FrontsCAPIInterface/SearchInput.js
RichieAHB: I missed that you fixed this because I'd commented on a redundant line! Sorrrry.

PR #382, opened 2018-09-17
Add hover actions and delete

RichieAHB:
I thought I'd merged this, just an initial commit for hover states for articles that also allows deleting articles!!!


RichieAHB:
@Reettaphant Do you mean drag and drop articles from the live site?

Reettaphant:
@RichieAHB I meant the button that you can click to copy anything from another part of the fronts tool to the clipboard. It's in the current tool so you can take a look there.

RichieAHB:
Ah ok, yeah that will be in a different PR. After speaking with Mariana, that wasn't deemed critical for CP for rollout.

PR #383, opened 2018-09-19
Move guration into lib

RichieAHB:
So after something that seemed like a good idea I've realised the error of my ways. Keeping this code outside of the lib made it harder to edit as the requirements appeared, made it a black box for the rest of the team, and ultimately made made the feedback loop far too long. While done with the best intentions it seems wise (and overdue) to bring it back in!

All this ultimately does is bring in all of the code, tests and README. The line count is increased quite a lot by the README and tests.


RichieAHB:
Hmmm, I ran the checks this side and they were fine, will investigate.

PR #384, opened 2018-09-19
Dedupe nested

RichieAHB:
~**Review this after the previous PR has been merged as this has branched from there and as such, contains those changes!** #383~

This change allows us to dedupe nested article fragments when dragging them from one front to another. Currently all that is happening here is that nested article fragments of dragged article fragments that already exist on the target front are being removed from the nested data of the drag and left in their original place in the front. This is most likely not ideal and the intention of the user is to probably replace the existing article fragments that are in the front and have them moved to this new position, but this is up for discussion!


client-v2/src/lib/guration/DedupeNode.js
```diff
@@ -0,0 +1,64 @@
...
+  ...ExternalDedupeNodeProps,
+  context: DedupeContextType
+|};
+
+class DedupeNode extends React.Component<DedupeNodeProps> {
+  componentDidMount = () => this.reregister();
+  componentDidUpdate = () => this.reregister();
+  componentWillUnmount = () => this.deregister();
+
+  get dedupeContext() {
```
LATaylor-guardian: What is the `get` keyword doing here?

client-v2/src/lib/guration/README.md
```diff
@@ -0,0 +1,198 @@
+# Guration
```
LATaylor-guardian: Great docs - nice work!

client-v2/src/lib/guration/DedupeNode.js
RichieAHB: It means I can drop the parens when calling this method (as it has arity 0). Someone once told me about the "uniform access principle" and I ate it right up! Also it means I can destructure like below and it runs the getter. I only use them if they're not side effecting.

```js
const { dedupeContext } = this;
```

RichieAHB:
@Reettaphant @jonathonherbert I've just tested this in the original fronts tool and it seems it doesn't care about nested stuff. You can have duplicates in nested articles ... is that a feature or just a limitation of the previous tool?

RichieAHB:
This has been superseded

PR #400, opened 2018-10-05
Fix persisting

RichieAHB:
This changes make sure that _all_ insert actions run the "replace" step of a dedupe action which means that persisting doesn't need to happen on both the remove step and replace step (see `insertAndDedupeSiblings`).

To fix persistence I've also added the ability to set the id to persist in the meta. Previously the `key` prop was being used but, rather than create loads of persistence actions, I just add the meta as part of the insert action, which means we have the `id` here which we can just pass through on the meta more easily.


client-v2/src/actions/ArticleFragments.js
```diff
@@ -105,19 +105,16 @@ const createInsertArticleFragment = (persistTo: 'collection' | 'clipboard') => (
...
   }
 
-  const insert = addPersistMetaToAction(insertAction, {
-    persistTo
-  });
-
   const replace = addPersistMetaToAction(replaceAction, {
-    persistTo
+    persistTo,
+    id
```
RichieAHB: This is now where we get the id from.

PR #402, opened 2018-10-05
Add article meta form

jonathonherbert:
Adds the article meta form - you can now click on articles to edit their meta.

edit-article-meta

Known issues/discussion points:

- This PR includes a bunch of updated types from flow-typed, which makes it seem bigger than it is. Don't worry - the majority of the file changes don't need reviewing! It also includes a few changes to fix Flow issues that have somehow crept into the build.
- Saving should work fine, but updates aren't visible in the collection article yet, because our Article -type doesn't derive data correctly from ArticleFragments ... yet.
- The code to derive data for the initial form values and to transform it outward for persistence is quite verbose, but currently lives alongside the form, making the form file large. I'm not sure whether it's better to colocate here or to break these functions out, perhaps at the expense of immediate legibility.
- The form is somewhat bulky, and I'll be chatting with UX to perhaps slim it down a bit.


client-v2/src/components/FrontsEdit/ArticleFragmentForm.js
```diff
@@ -0,0 +1,354 @@
+// @flow
+
```
jonathonherbert: As mentioned in the PR description, this form file is large; we could look at splitting some of the transform in/out logic out of it.

client-v2/src/components/FrontsEdit/FrontContainer.js
```diff
@@ -173,7 +170,7 @@ const createMapStateToProps = () => {
   });
 };
 
-const mapDispatchToProps = (dispatch: *) => ({
+const mapDispatchToProps = (dispatch: Dispatch) => ({
```
RichieAHB: 😞

PR #405, opened 2018-10-05
Add a way of editing clipboard meta

RichieAHB:
This allows editing meta for clipboard items.

I"m going away 🌴 so I haven't tested this as much as I'd like but I'm pretty sure it all works!


RichieAHB:
Closed by #420

PR #409, opened 2018-10-10
Derive article correctly from CAPI article and ArticleFragment meta

jonathonherbert:
This PR rounds out the Article type and selector result to contain the information necessary to represent itself properly, both in the article meta form and when displaying Articles in collections.

In doing this it was necessary to expand the CAPI definition and import a few utility functions from V1 to correctly derive thumbnails.

There'll be more to do here as we expand the number of things we edit in Fronts.

### DIFF
--- a/client-v2/src/util/CAPIUtils.js
+++ b/client-v2/src/util/CAPIUtils.js
@@ -1,6 +1,5 @@
 // @flow
 
-import get from 'lodash/get';
 import { getArticles } from 'services/faciaApi';
 import type { Element } from 'types/Capi';
 import type { ExternalArticle } from '../shared/types/ExternalArticle';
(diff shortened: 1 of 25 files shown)


client-v2/src/util/CAPIUtils.js
```diff
@@ -40,4 +42,44 @@ const getThumbnailFromElements = (_elements: Element[]) => {
...
-export { getURLInternalPageCode, getThumbnailFromElements };
+function getContributorImage(externalArticle: ExternalArticle) {
+  const contributor =
+    externalArticle.tags &&
+    externalArticle.tags.find(tag => tag.type === 'contributor');
+
+  return contributor && contributor.bylineLargeImageUrl;
+}
+
+function getThumbnail(
```
Reettaphant: Should we write tests for this?

jonathonherbert: Yes! They're now there :)

PR #423, opened 2018-10-18
Only rerender when necessary

RichieAHB:
This will ensure we only render when necessary. As discussed IRL it doesn't fix the slowness completely (due to our tree, which will need a revisit) but definitely helps.


client-v2/src/lib/guration/Root.tsx
```diff
@@ -158,7 +159,7 @@ class Root<T extends object> extends React.Component<RootProps<T>, RootState> {
     path: Path[] | null,
     canDrop: boolean | null,
     // TODO: this is essentially `any` at the moment
-    state = {}
+    state: object | null = null
```
RichieAHB: This is the main part of this change - state was always true so none of the optimizing below worked.

PR #424, opened 2018-10-18
Revert broken TS changes

RichieAHB:
This fixes some broken tag inputs that got broken when I got myself into a loop with TS :(
tags


RichieAHB:
I got myself in a state trying to reproduce Flow's inference in TS, when all I needed to do was specify the properties directly ... couldn't see the wood for the trees 🌳

PR #432, opened 2018-10-19
Search for scheduled content when searching for preview

Reettaphant:
Turns out the old tool does a slightly different capi query for draft content. I've now fixed it so that in draft we see exactly what we do in the old tool and live shows slightly different results as requested by our users.

### DIFF
--- a/client-v2/src/components/CAPI/SearchQuery.tsx
+++ b/client-v2/src/components/CAPI/SearchQuery.tsx
@@ -13,6 +13,7 @@ interface CAPISearchQueryProps {
   children: AsyncChild<SearchReturn>;
   params: object;
   poll?: number;
+  isPreview: boolean;
 }
 
 interface CAPISearchQueryState {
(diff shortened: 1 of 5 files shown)


client-v2/src/components/CAPI/SearchQuery.tsx
```diff
@@ -30,15 +31,15 @@ class SearchQuery extends React.Component<
...
+    { baseURL, fetch, isPreview }: CAPISearchQueryProps,
     prevState: CAPISearchQueryState
   ) {
     if (
       (baseURL && prevState.baseURL !== baseURL) ||
       (fetch && prevState.fetch !== fetch)
     ) {
       return {
-        capi: capiQuery(baseURL, fetch).search,
+        capi: isPreview ? capiQuery(baseURL, fetch).scheduled : capiQuery(baseURL, fetch).search,
```
Reettaphant: Since it makes to always query scheduled content when searching for preview I'm not too worried about having this logic in the search query component. It's annoying though that this also requires changing search params in another component.

PR #436, opened 2018-10-20
DND

RichieAHB:
@Reettaphant @jonathonherbert so basically there is an issue with the old drag and drop as it stands (one of the many ...) where copying an article across won't copy the edited meta data across, so making edits in the clipboard is basically redundant.

I've got this change here which is pretty large and is missing tests (although I've sat and tested it extensively this evening), which fixes every issue I can think of with dragging:
- There is now no delay, even on massive collections, when dragging around
- Far fewer unnecessary CAPI calls are made (and the rest can be easily factored out)
- Meta _is_ copied across
- There are about 1000 lines less code (accounting for ~350 for the test that need to go back in)


RichieAHB:
@Reettaphant @jonathonherbert @vlbee I've now added tests back into this - I'd obviously like to add more to catch as many edge cases as possible but I think most of those edge cases will come at the integration level as discussed IRL.

PR #439, opened 2018-10-23
Remove tree selectors

RichieAHB:
This removes the tree selectors and reverts back to simple child selectors making things a touch snappier and easier to reason about. The test coverage remains the same as we're just reverting back to selectors that already existed and I haven't created any new ones - just removed the tree ones.

Because drops now receive an `ArticleFragment` rather than a `DenormalisedArticleFragment` - due to the fact that the full tree is never built - a fair amount of the work here was to build that tree as needed in order to grab all the meta from the nested article fragments.

I've also tidied some of the `Level` components to make them easier to read as this was making this refactor harder.

I just need to fix a few specs but I'll do this after I've verified that I can do what I want after this PR (working on another branch)!


client-v2/src/shared/selectors/shared.ts
```diff
@@ -165,25 +168,31 @@ const articleFragmentIdSelector = (
...
+const createSupportingArticlesSelector = () =>
+  createSelector(
+    articleFragmentsFromRootStateSelector,
+    articleFragmentIdSelector,
+    (articleFragments, id) =>
+      (articleFragments[id].meta.supporting
+        ? articleFragments[id].meta.supporting!
+        : []
+      ).map((sId: string) => articleFragments[sId])
+  );
```
jonathonherbert: Is there a reason we're exporting both selectors and selector factories here? It seems likely a developer would use the selector rather than the factory by mistake, and miss out on yummy memoisation goodness! Or is is that there are currently components using the selectors, and that we need to refactor further down the line?

PR #442, opened 2018-10-24
Fix article fetching problems

jonathonherbert:
We currently have three problems when fetching articles - 

- Snaplinks are filtered at the fetch level, not the thunk level, and so we aren't filtering all of our snaplinks out of CAPI requests
- Errors aren't raised when the CAPI results length doesn't match the request length, which could lead to orphaned loading states
- The pagesize is set to 10 when our requests can contain up to fifty article ids 🙃 

This PR should fix these issues, and adds regression tests.

### DIFF
--- a/client-v2/src/actions/Collections.ts
+++ b/client-v2/src/actions/Collections.ts
@@ -24,9 +24,10 @@ import { groupsReceived } from 'shared/actions/Groups';
 import { actions as collectionActions } from 'shared/bundles/collectionsBundle';
 import { getCollectionConfig } from 'selectors/frontsSelectors';
 import { State } from 'types/State';
-import { Dispatch, ThunkResult } from 'types/Store';
+import { Dispatch, ThunkResult, GetState } from 'types/Store';
 import { Collection } from 'shared/types/Collection';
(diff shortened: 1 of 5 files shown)


client-v2/src/actions/Collections.ts
```diff
@@ -101,6 +102,37 @@ function updateCollection(collection: Collection): ThunkResult<Promise<void>> {
...
+    const articles = await getArticles(articleIdsWithoutSnaps);
+    dispatch(externalArticleActions.fetchSuccess(articles));
+    const remainingArticles = difference(
+      articleIdsWithoutSnaps,
+      articles.map(_ => _.id)
+    );
+    if (remainingArticles.length) {
+      dispatch(
+        externalArticleActions.fetchError(
+          `The following article ids were in a CAPI query but were not returned by CAPI: ${remainingArticles.join(
```
Reettaphant: This is not for now but when we start looking into error handling in the tool, we need to think about errors which we show to the user and errors which we can see if this happens.

jonathonherbert: Absolutely.

PR #443, opened 2018-10-24
Add test for all of the adding actions

RichieAHB:
There were some missed code paths with testing "adding" actions (the actions that get called when dragging and dropping). This PR adds coverage for all the different ones.


client-v2/src/actions/__tests__/ArticleFragments.spec.ts
```diff
@@ -1,69 +1,180 @@
+jest.mock('uuid/v4', () => jest.fn(() => 'uuid'));
```
RichieAHB: Should have thought of this long ago!

jonathonherbert: *maniacal laugh*

PR #444, opened 2018-10-24
Add snap links and loading states to fill the gaps in collections

jonathonherbert:
This PR adds snap links:

screen shot 2018-10-24 at 19 24 01

screen shot 2018-10-24 at 19 24 35

And loading states for articles (reducing the jank users experience when articles load, as they occupy the same space):

placeholder-anim

... unifying them under a single CollectionItem component, which then switches based on the collection item type via a selector. As a result, together with #442 , we no longer have gaps in our collections.

Sorry, it's quite a big one - happy to go through as needed!


client-v2/src/components/FrontsEdit/Front.tsx
```diff
@@ -204,15 +224,23 @@ const mapStateToProps = (state: State, props: FrontPropsBeforeState) => ({
...
+    dispatch,
+    updateArticleFragmentMeta: (id: string, meta: ArticleFragmentMeta) =>
+      dispatch(updateArticleFragmentMeta(id, meta)),
+    selectArticleFragment: (frontId: string, articleFragmentId: string) =>
+      dispatch(editorSelectArticleFragment(frontId, articleFragmentId)),
+    clearArticleFragmentSelection: (frontId: string) =>
+      dispatch(editorClearArticleFragmentSelection(frontId)),
+    removeCollectionItem: (parentId: string, uuid: string) => {
+      dispatch(removeGroupArticleFragment(parentId, uuid));
+    },
```
jonathonherbert: We now delegate deletion behaviour to whatever it is rendering our collection items, as a single collection item can now represent an item in either a main or supporting position.

client-v2/src/shared/components/CollectionItemMetaContainer.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
```
jonathonherbert: I've broken out a lot of the Article specific styled components into CollectionItem components to share with the new SnapLink component.

client-v2/src/components/FrontsEdit/CollectionComponents/CollectionItem.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import { connect } from 'react-redux';
```
jonathonherbert: Formerly ArticleFragment, this now represents anything we might want to render in a collection.

PR #445, opened 2018-10-25
Draft articles

Reettaphant:
Adds draft labels to clipboard, capi feed and collections. 

Fixes also some inconsistencies in the clipboard kicker presentation.

There is some code duplication here, in the clipboard in particular, but I didn't want to spend too much time in sorting it out as @jonathonherbert and I were discussing this morning a much more comprehensive approach to unifying these different components.


client-v2/src/shared/components/ArticlePolaroidSub.tsx
```diff
@@ -79,9 +80,9 @@ const ArticleComponent = ({
         >
           Delete
         </ButtonDefault>
-        <TonedKicker tone={article.tone}>{article.sectionName}</TonedKicker>
+        <TonedKicker tone={article.tone} isLive={article.isLive}>{ getArticleLabel(article) }</TonedKicker>
         {` ${truncate(article.headline, {
-          length: 40 - article.sectionName.length
+          length: 45 - getArticleLabel(article).length
```
jonathonherbert: good catch!

PR #446, opened 2018-10-26
Section labels and pillar colours

Reettaphant:
This PR corrects the article colours we display on article labels in live content. Article colours should be derived from it's pillar, not tone. On the clipboard and search bar the sectionName is coloured, in collections we colour the kicker but display the section name on the left. 

I'm not sorting out labs content colours in this pr, we can do it before we rollout to commercial.

I will sort out the merge conflicts once we have a version of master with a passing build again.


client-v2/src/shared/util/getPillarColor.ts
```diff
@@ -0,0 +1,24 @@
+const pillarColorMap = {
+  'pillar/news': '#c70000',
+  'pillar/sport': '#0084c6',
+  'pillar/opinion': '#c70000',
+  'pillar/lifestyle': '#bb3b80',
```
jonathonherbert: Sanity returns!

PR #448, opened 2018-10-30
Add hover items wrapper & tooltip component 

vlbee:
After a long gestation, this should be finally ready to merge. Replaces PR #438

- Add hover items wrapper & tooltip component to shared/components
- Applies hover action to shared/components/Article
- Applies hover action to shared/components/SnapLink
- Upgrades getPaths utility function

Preview Gif:
hover wrapper


client-v2/src/shared/components/SnapLink.tsx
```diff
@@ -36,34 +38,39 @@ interface ContainerProps extends SnapLinkProps {
   selectSharedState?: (state: any) => State;
 }
 
-const SnapLink = ({ id, fade, onClick, size = 'default', onDelete, ...rest }: SnapLinkProps) => (
+const SnapLink = ({
+  id,
+  fade,
+  onClick,
+  isSnapLink,
```
jonathonherbert: Is there a reason we have an isSnapLink prop here? (This is a snaplink component - surely we know we're a snap link! ...but I may be missing something 😄 )

client-v2/src/shared/components/input/HoverActionButtonWrapper.tsx
```diff
@@ -0,0 +1,116 @@
...
+    });
+  };
+
+  private hideToolTip = () => {
+    this.setState({
+      isToolTipVisible: false
+    });
+  };
+
+  private setToolTipText = (text: string) =>
```
jonathonherbert: Is it necessary to have this method, or would it be possible to combine it with 'showTooltip', which would receive the 'text' argument?

PR #451, opened 2018-10-30
Add snaplinks to clipboard

jonathonherbert:
Adds snaplinks to the clipboard.

screen shot 2018-10-30 at 14 11 26

In doing this, unifies the idea of a 'CollectionItem' across the app. Now, whenever we render any member of a collection, we render it as a CollectionItem; any complexity in how we render that particular item is hidden from the consuming code.


client-v2/src/shared/components/article/Article.tsx
```diff
@@ -0,0 +1,152 @@
+import React, { ReactNode } from 'react';
+import styled from 'styled-components';
```
jonathonherbert: I've reorganised these components into their own folders to keep the main directory from becoming cluttered and to colocate related things.

client-v2/src/components/Clipboard.tsx
```diff
@@ -72,20 +88,28 @@ class Clipboard extends React.Component<ClipboardProps> {
                 onDrop={this.handleInsert}
               >
                 {(supporting, sProps) => (
-                  <ArticlePolaroidSub
-                    id={supporting.uuid}
+                  <CollectionItem
+                    uuid={supporting.uuid}
```
RichieAHB: Good change 👍

PR #453, opened 2018-10-31
Add hover buttons to the feed items

jonathonherbert:
... yoouuu know it!

screen shot 2018-10-31 at 14 22 18


client-v2/src/components/FeedItem.tsx
```diff
@@ -112,7 +122,10 @@ const FeedItem = ({
         }}
       >
         {isLive && startCase(sectionName)}
-        {!isLive && (firstPublicationDate ? notLiveLabels.takendDown: notLiveLabels.draft)}
+        {!isLive &&
+          (firstPublicationDate
+            ? notLiveLabels.takendDown
+            : notLiveLabels.draft)}
```
RichieAHB: I wonder whether we can get ts-lint to hook into prettier? Quite a few of my PRs have these added in too. (Unless there was some meaning here and I've missed it!)

vlbee: awesome! 👍

jonathonherbert: presuming the parens are the subject here, we do hook tslint into prettier (tslint-config-prettier) - in this case removing the parens would change the behaviour!

jonathonherbert: also yes good re-use thanks to @vlbee!

RichieAHB: It's specifically the fact that when I run prettier on a file it'll make changes that `ts-lint` doesn't catch - will happily take a look given it seems to be me that's seen this the most 👍

PR #454, opened 2018-10-31
Fix CAPIUtils getIDFromURL to enable dragging links from preview

vlbee:
paths

### DIFF
--- a/client-v2/src/fixtures/articleFragment.ts
+++ b/client-v2/src/fixtures/articleFragment.ts
@@ -6,15 +6,15 @@ const articleFragmentWithSlideshowThumbnailMeta = {
     {
       src: 'exampleSrc1',
       thumb: 'exampleThumbnail1',
-      width: 100,
-      height: 100,
+      width: '100',
+      height: '100',
       origin: 'exampleOrigin1'
     },
     {
       src: 'exampleSrc2',
       thumb: 'exampleThumbnail2',
-      width: 100,
-      height: 100,
+      width: '100',
+      height: '100',
       origin: 'exampleOrigin2'
     }
   ]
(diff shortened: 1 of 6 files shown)


client-v2/src/fixtures/articleFragment.ts
```diff
@@ -1,5 +1,3 @@
-// @flow
```
RichieAHB: 😆

client-v2/src/util/CAPIUtils.ts
```diff
@@ -4,7 +4,9 @@ import { ArticleFragment } from '../shared/types/Collection';
 
 const getIdFromURL = (url: string): string | null => {
   const [, id = null] =
-    url.match(/^https:\/\/www.theguardian\.com\/(.*)\??/) || [];
+    url.match(
+      /^https:\/\/(?:www.theguardian\.com|viewer.gutools.co.uk)\/(?:(?:preview|live)\/)?(.*)\??/
```
RichieAHB: Currently this will return the wrong path for any articles that exist at a path such as https://www.theguardian.com/live/myarticle - it will only return `myarticle` for the path as live will match the regex and not be captured.

I just checked and currently https://www.theguardian.com/live redirects to https://www.theguardian.com/tone/minutebyminute so perhaps not an issue but could be future-proofed with:

```suggestion
      /^https:\/\/(?:www.theguardian.com|viewer.gutools.co.uk\/(?:preview|live))\/([^?]*)/
```

If this seems worthwhile then maybe we can add a test for it?

vlbee: I'm not 100% of all the URLs but have tweaked the regex slightly to match all of the below:

```
/^https:\/\/(?:www.theguardian.com\/|viewer.gutools.co.uk(?:\/(?:preview|live))?\/)([^?]*)/
```

image

RichieAHB: Awesome - sorry if mine missed any of your test cases!

RichieAHB: Also I just wanted to try the new github suggested edits 🙄

vlbee: New github edits  👍  If this now covers everything I'll merge the PR after rebasing

PR #458, opened 2018-11-01
Make clipboard drop zone larger (and hover bug fix)

RichieAHB:
Makes the drop zone in the clipboard expand to the size of the remaining space of the clipboard if it's not vertically full.

Also fixes a small bug where the drag and drop above / below a node wasn't changing at 50% vertically.

drop-zone


RichieAHB:
As discussed IRL, this _is_ pretty red ... but I think we need Josh to help here!

PR #460, opened 2018-11-01
What you should know

gtrufitt:
Quick update to make things easier for developers coming to the project.

Initial update of Readme to document technologies and structure. Please add more...

Easier on the eyes:
https://github.com/guardian/facia-tool/blob/ad9c31544ed266808eb06038cd4be336dc62b009/client-v2/README.md


client-v2/README.md
```diff
@@ -60,20 +116,66 @@ And to fix any issues that are automatically resolvable by eslint run:
...
-To check for flow coverage, run `yarn flow-coverage`.
+- Components, Actions, Reducers and Selectors are top level.
+    - The [App component](src/components/App.tsx) is the application entry point
+    - All reducers are combined in the [Root Reducer](src/reducers/rootReducer.ts) as per standard convention
+- `bundles` **what is the concept of a bundle?**
+- `constants` store high-level application constants such as theme styles and image paths
+- `services` contains the modules for requests to APIs such as CAPI and FaciaAPI
+- `lib` contains modules designed to be reusable such as the Drag N' Drop (dnd) module
+- `shared` is a library written using React and Redux and Typescript for typing designed to make it easier to develop tools which need to provide the ability to curate collections of content. [Read more.](src/shared/.README)
+- `types` at the top level store types for ??
```
RichieAHB: This is the top level store for types that have no obvious home: e.g. `Action` which is a union of things that are split across a few files. I wonder whether some of those (perhaps `State` / `Store`) could be move to places like the root reducer etc? Any thoughts @Reettaphant , @jonathonherbert , @vlbee

client-v2/README.md
```diff
@@ -60,20 +116,66 @@ And to fix any issues that are automatically resolvable by eslint run:
...
+- `types` and `__tests__` are co-located with their modules at the folder level
```
RichieAHB: The other `types` folder is for the `shared` sub directory, which is meant to be conceptually separate from the rest of the domain logic in `src`. @jonathonherbert and I discussed moving this, (along with some other parts) into the `lib` directory and exporting only the necessary exports for `shared` (similar to `dnd`) from an `index.js` file in that directory to help decoupling some of these concepts. Any thoughts on this too @vlbee @Reettaphant

client-v2/README.md
```diff
@@ -60,20 +116,66 @@ And to fix any issues that are automatically resolvable by eslint run:
...
+## Typescript
+We are using Typescript for typing in Fronts V2. 
 
-If you would like to overwrite types that are previously installed to newer
-versions that have been updated in the project pass the `--overwrite` flag to
-the above command.
+## File Structure
 
-If you want to run flow to check for type errors, run `yarn flow`.
+```
```
RichieAHB: Again, as discussed with @jonathonherbert , we could probably do with a bit of a think / tidy in this area. I think some of these directories have grown organically but probably aren't that clear on their separation. Specifically `containers` and `components` seemed to be interleaved and (at least for me) this has proven to be a useful mental separation to have.

client-v2/README.md
```diff
@@ -19,8 +37,46 @@ To start developing just run:
...
+
 ## Testing
 
+| Uses        | For         | Config |
+| ------------|------------- |--- |
+| [Jest](https://jestjs.io/docs/en/getting-started.html)      | Testing library | [jest.config.js](jest.config.js)|
+| [Enzyme](https://github.com/airbnb/enzyme)      | JavaScript Testing utilities for React  ||
+| [React Test Renderer](https://reactjs.org/docs/test-renderer.html)      | Render Components to JSON (e.g for Jest Snapshotting) ||
+| [Test Cafe](https://testcafe.devexpress.com/)      | Integration testing - see below ||
+
```
RichieAHB: As mentioned here we could throw in `react-testing-libaray` here. @vlbee is that something you can just add into this(with GitHub's new suggested change!?) maybe unless @gtrufitt is already on it?!

client-v2/README.md
```diff
@@ -60,20 +116,66 @@ And to fix any issues that are automatically resolvable by eslint run:
...
+├── types
+└── util
+    └── sharedStyles
+```
 
-To check for flow coverage, run `yarn flow-coverage`.
+- Components, Actions, Reducers and Selectors are top level.
+    - The [App component](src/components/App.tsx) is the application entry point
+    - All reducers are combined in the [Root Reducer](src/reducers/rootReducer.ts) as per standard convention
+- `bundles` **what is the concept of a bundle?**
```
RichieAHB: @jonathonherbert do you reckon you could add a suggested change here just to give a simple overview of the key concepts?

client-v2/README.md
Reettaphant: The shared library has already become too tightly coupled with the fronts tool - there is a card on trello for fixing this (here)[https://trello.com/c/SYgZ62dV/281-remove-reverse-type-dependencies-from-fronts-tool-to-the-shared-library-eg-capiarticle-type-defined-outside-of-shared]. Maybe we could update that card with this and fix this when we have time?

client-v2/README.md
Reettaphant: I really like having all types in the types folder, I find it confusing when we define types in other files but that's just something I'm used to.

client-v2/README.md
jonathonherbert:
```suggestion
- `bundles`
    - A bundle exports a reducer and all of the related things a reducer needs to function in an app - selectors, actions and action names. It's a bit like an index.ts for a single redux module. This is especially useful when you're generating the actions, reducer and selectors rather than writing them manually, for example with the `createAsyncResourceBundle` utility in shared/util.
```

client-v2/README.md
```diff
@@ -19,8 +37,46 @@ To start developing just run:
...
+| [Yarn](https://yarnpkg.com/en/)      | Yarning | [package.json](package.json) |
+| [Babel](https://babeljs.io/)      | Compile Typescript etc | [babel.config.json](babel.config.json) |
+| [Webpack](https://webpack.js.org/)      | Bundle your assests | [webpack.config.common.js](config/webpack.config.common.js), [webpack.config.prod.js](config/webpack.config.prod.js), [webpack.config.dev.js](config/webpack.config.dev.js) |
+
 ## Testing
 
+| Uses        | For         | Config |
+| ------------|------------- |--- |
+| [Jest](https://jestjs.io/docs/en/getting-started.html)      | Testing library | [jest.config.js](jest.config.js)|
+| [Enzyme](https://github.com/airbnb/enzyme)      | JavaScript Testing utilities for React  ||
```
vlbee:
```suggestion
| [Enzyme](https://github.com/airbnb/enzyme)      | JavaScript Testing utilities for React  ||
| [react-testing-library](https://github.com/kentcdodds/react-testing-library)     | JavaScript Testing utilities for React Components |
```

client-v2/README.md
vlbee:
```suggestion
| [Enzyme](https://github.com/airbnb/enzyme)      | JavaScript Testing utilities for React  ||
| [react-testing-library](https://github.com/kentcdodds/react-testing-library)     | JavaScript Testing utilities for React Components |
```

vlbee: I've added the suggestion further down the conversation.

client-v2/README.md
RichieAHB: Thank you! 👍

client-v2/README.md
RichieAHB: Ok I'm happy with that too - I guess we just need to be consistent either way for newcomers to the project. Potentially we can extract some types from places then?

client-v2/README.md
RichieAHB: Yeah sounds great! 👍

client-v2/README.md
RichieAHB: Excellenté - thank @jonathonherbert

client-v2/README.md
```diff
@@ -60,20 +118,26 @@ And to fix any issues that are automatically resolvable by eslint run:
...
+    - All reducers are combined in the [Root Reducer](src/reducers/rootReducer.ts) as per standard convention
+- `bundles`
+    - A bundle exports a reducer and all of the related things a reducer needs to function in an app - selectors, actions and action names. It's a bit like an index.ts for a single redux module. This is especially useful when you're generating the actions, reducer and selectors rather than writing them manually, for example with the `createAsyncResourceBundle` utility in shared/util.
+- `constants` store high-level application constants such as theme styles and image paths
+- `services` contains the modules for requests to APIs such as CAPI and FaciaAPI
+- `lib` contains modules designed to be reusable such as the Drag N' Drop (dnd) module
+- `shared` is a library written using React and Redux and Typescript for typing designed to make it easier to develop tools which need to provide the ability to curate collections of content. [Read more.](src/shared/.README)
+- `types` at the top level contains types that have no obvious home: e.g. Action which is a union of things that are split across a few files.
```
gtrufitt: Took your explanation @RichieAHB - Feel free to suggest change

gtrufitt:
> Actually, I take it back! After discussing with people here we feel that perhaps the folder structure could come out given that it's not annotated with any info, it just mirrors (until it doesn't) the structure!

I knew it'd be contentious  - for me it's useful as part of the .README but it'll never be up to date unless it's automated and updated with a hook so pulling out for now

client-v2/README.md
```diff
@@ -19,8 +37,48 @@ To start developing just run:
...
+| [Babel](https://babeljs.io/)      | Compile Typescript etc | [babel.config.json](babel.config.json) |
+| [Webpack](https://webpack.js.org/)      | Bundle your assests | [webpack.config.common.js](config/webpack.config.common.js), [webpack.config.prod.js](config/webpack.config.prod.js), [webpack.config.dev.js](config/webpack.config.dev.js) |
+
 ## Testing
 
+| Uses        | For         | Config |
+| ------------|------------- |--- |
+| [Jest](https://jestjs.io/docs/en/getting-started.html)      | Testing library | [jest.config.js](jest.config.js)|
+| [Enzyme](https://github.com/airbnb/enzyme)      | JavaScript Testing utilities for React  ||
+| [react-testing-library](https://github.com/kentcdodds/react-testing-library)     | JavaScript Testing utilities for React Components |
```
vlbee:
```suggestion
```
Removing duplicate table row that was showing up for react-testing-lib

PR #464, opened 2018-11-08
Duplicate platform specific containers are now cleaned up when added ...

jonathonherbert:
... as beforehand, they remained in the front and there were ways around validation, allowing fronts to be persisted in invalid states.

fronts-probs

Tested on CODE. There aren't any tests for this yet but I'd be happy to pair with somebody to cover this case - I couldn't find any that covered similar cases.

### DIFF
--- a/public/src/js/modules/grid.js
+++ b/public/src/js/modules/grid.js
@@ -4,7 +4,7 @@ import {CONST} from 'modules/vars';
 export function recordUsage(mediaId, frontId) {
     const usageData = {
         mediaId: mediaId,
-        front: frontId,
+        front: frontId
     };
 
     return request({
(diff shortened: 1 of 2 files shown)


public/src/js/modules/grid.js
```diff
@@ -4,7 +4,7 @@ import {CONST} from 'modules/vars';
 export function recordUsage(mediaId, frontId) {
     const usageData = {
         mediaId: mediaId,
-        front: frontId,
```
jonathonherbert: ... did our linting get stricter, or... ?

Reettaphant: How strange, I don't think we've changed anything in the old app.

PR #467, opened 2018-11-14
Snap links now droppable

jonathonherbert:
We can now drag snap links into article positions. Snaps can come from Guardian tag or section pages,  or just external web pages.

If a tag or section is dropped, there's an option to create a 'latest' snap. This is currently a horrible dialog box, but once we've merged a modal solution we can move to a more user-friendly implementation.

front-snaps

...yes, it's underwhelming 😁 This PR does not
- display snaplink content, which means that it's basically useless to users as it stands - however, this won't take much work!
- allow the user to edit snaplink content, which will require some forms work.


client-v2/src/services/__tests__/faciaApi.spec.ts
```diff
@@ -75,22 +76,43 @@ describe('faciaApi', () => {
...
+      chunkedArticleIds.map(_ =>
+        fetchMock.once(
+          getCapiUriForContentIds(_),
+          '{"response":{ "results": [] }}'
+        )
+      );
+      await getArticlesBatched(articleIds);
+    });
+  });
+  describe('getArticles', () => {
```
Reettaphant: rename `getArtcles` function to singular and maybe instead of articles get content since it is no longer getting articles?

jonathonherbert: done and done 👍

client-v2/src/shared/actions/ArticleFragments.ts
```diff
@@ -112,23 +115,62 @@ const replaceGroupArticleFragments = (
...
+ *  - an external link.
+ */
+function createArticleFragment(
+  resourceId: string
+): ThunkResult<Promise<ArticleFragment | undefined>> {
+  return async (dispatch: Dispatch) => {
+    const isURL = isValidURL(resourceId);
+    const id = isURL ? getIdFromURL(resourceId) : resourceId;
+    let fragment: ArticleFragment;
+    if (isURL && !id) {
```
Reettaphant: I think we may be able to remove some of the if else's here. We could do something like this: 

```
if (!id && !isURL) {
  return;
} 

if (isUrl && !id) {
...

``` 

It's not a big deal but might make the code more readable?

client-v2/src/shared/actions/ArticleFragments.ts
```diff
@@ -112,23 +115,62 @@ const replaceGroupArticleFragments = (
...
+      // If we have a URL from an external site,
+      // create a snap of type 'link'.
+      fragment = await createLinkSnap(resourceId);
```
Reettaphant: Are we not returning anything in this instance? Only asking because I am interested in getting rid of the elses!

client-v2/src/shared/util/url.ts
```diff
@@ -13,4 +18,56 @@ function getAbsolutePath(url: string, keepQueryString?: boolean) {
   return keepQueryString ? path + a.search : path;
 }
 
-export { getAbsolutePath };
+/**
+ * Get the complete viewable URL for an article.
+ */
+export function getViewUrl(article: DerivedArticle) {
```
Reettaphant: Could we get rid of the elses in the function by returning the `url` ince we know it's value?

jonathonherbert: We can get rid of this function completely, because it's not used! 🙃

client-v2/src/shared/actions/ArticleFragments.ts
jonathonherbert: ... yeah, I've refactored this to be a bit less smelly.

RichieAHB:
Just want to say this looks really really good 🎉

PR #468, opened 2018-11-14
Style snap links

jonathonherbert:
A first stab at styling snap links so they mean something to the user.

screen shot 2018-11-14 at 21 39 36

For god's sake, don't merge this before #467.


client-v2/src/shared/components/snapLink/SnapLink.tsx
```diff
@@ -51,43 +50,69 @@ const SnapLink = ({
...
-        ) : (
-          <>
-            <strong>Snap link</strong>
-            {snapLinkPlaceholder}
-          </>
+}: SnapLinkProps) => {
+  const headline =
+    articleFragment.meta.headline ||
+    (articleFragment.meta.customKicker
+      ? `{ ${articleFragment.meta.customKicker} }`
```
jonathonherbert: It's what the old tool does? 🤷‍♂️

PR #469, opened 2018-11-15
Fix integration tests

jonathonherbert:
... because my previous PR broke them! For two reasons:

- window.location.(hostname|name) are not the same thing, and the integration tests caught this subtle distinction 🎉 
- the CAPI response the integration tests are modelling changes when we request individual pieces of content.

### DIFF
--- a/client-v2/integration/server/server.js
+++ b/client-v2/integration/server/server.js
@@ -20,7 +20,7 @@ module.exports = async () =>
       res.sendFile(path.join(__dirname, './index.html'))
     );
 
-    // send all api requests to here
+    // Endpoint for api search requests here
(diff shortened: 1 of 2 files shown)


client-v2/integration/server/server.js
```diff
@@ -53,13 +53,48 @@ module.exports = async () =>
...
+          content: article,
+          status: 'ok',
+          total: 1
+        }
+      };
+      return res.json(result);
+    };
+    // Attempts at a capture group:
+    // /api/(preview|live)/*
+    // /api/(?!preview|live)/*
```
vlbee:
```suggestion
    // /api/(?:preview|live)/*
```

PR #470, opened 2018-11-15
Rv kicker confusions

Reettaphant:
Previously if we had no kicker on the article we'd display the pillar name in place of the kicker. On the clipboard we'd never display the kicker here, we'd only ever display the section name or draft or taken down label if the article wasn't live. Jessie pointed out that this is confusing as it was hard to tell which articles had kickers and which did not. Now we are only displaying kickers expect when an article is draft on the clipboard in which case we display the draft label. So we still have a problem on displaying kickers on the clipboard for articles which are not live. Any ideas for what to do about this?
screen shot 2018-11-15 at 15 35 21

### DIFF
--- a/client-v2/src/shared/components/article/ArticleBodyPolaroid.tsx
+++ b/client-v2/src/shared/components/article/ArticleBodyPolaroid.tsx
@@ -34,11 +34,11 @@ const ArticlePolaroidComponent = ({
   displayPlaceholders,
   pillarId,
   thumbnail,
-  sectionName,
+  kicker,
   isLive
 }: ArticleBodyProps) => {
   const articleLabel =
-    getArticleLabel(firstPublicationDate, sectionName, isLive) || '';
+    getArticleLabel(firstPublicationDate, kicker, isLive) || '';
   return (
     <>
       {size === 'default' &&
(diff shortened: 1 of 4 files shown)


client-v2/src/shared/components/article/ArticleBodyPolaroid.tsx
```diff
@@ -34,11 +34,11 @@ const ArticlePolaroidComponent = ({
   displayPlaceholders,
   pillarId,
   thumbnail,
-  sectionName,
+  kicker,
   isLive
 }: ArticleBodyProps) => {
   const articleLabel =
-    getArticleLabel(firstPublicationDate, sectionName, isLive) || '';
+    getArticleLabel(firstPublicationDate, kicker, isLive) || '';
```
jonathonherbert: 👍

jonathonherbert:
Ace! Just to double check - is there an extra space in front of the article headlines?

Re: your question - what are the reasons why the clipboard behaviour can't be consistent with the collection behaviour?

harpreetpurewal:
Looks good

Reettaphant:
@jonathonherbert There is no space in the clipboard for both. In collections we display the draft label in the place of the section name which does not appear on the clipboard. I'll double check with central prod to make sure this is what they want. Have also removed the extra space - good catch.

PR #474, opened 2018-11-16
Visible indicators

Reettaphant:
Mobile and Desktop visibility indicators on collection items! Jon already told me that there is no way I'm getting `properStages` as a variable name through peer review so I welcome any better suggestions :).
screen shot 2018-11-16 at 12 22 39

### DIFF
--- a/client-v2/src/actions/Collections.ts
+++ b/client-v2/src/actions/Collections.ts
@@ -6,6 +6,7 @@ import {
   updateCollection as updateCollectionFromApi,
   fetchVisibleStories
 } from 'services/faciaApi';
+import { VisibleStoriesResponse } from 'types/FaciaApi';
 import {
   selectUserEmail,
   selectFirstName,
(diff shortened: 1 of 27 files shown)


client-v2/src/shared/types/Collection.ts
```diff
@@ -7,6 +7,7 @@ interface Group {
   articleFragments: string[];
 }
 
+type ProperStages = 'draft' | 'live';
```
jonathonherbert: Suggestions from the team - CurrentStages? ActiveStages?

Or, we could name ProperStages as Stages and rename Stages to CollectionItemSets or similar?

client-v2/src/actions/Collections.ts
```diff
@@ -3,77 +3,97 @@ import { batchActions } from 'redux-batched-actions';
...
-          batchActions([
-            collectionActions.fetchSuccess(collection),
-            articleFragmentsReceived(articleFragments),
-            recordUnpublishedChanges(collectionId, hasUnpublishedChanges),
-            groupsReceived(groups)
-          ])
-        );
+      dispatch(
+        batchActions([
+          recordVisibleStories(collection.id, liveVisibleStories, frontStages.live),
```
RichieAHB: Is there a way we can create this action with a shape that means we only have to call it once with a subset of either `live` or `draft` in the payload? Maybe that isn't desirable anyway?

client-v2/src/util/frontsUtils.ts
```diff
@@ -43,9 +47,26 @@ const isFrontStale = (lastUpdated?: number, lastPressed?: number) => {
   return false;
 };
 
+const getVisibilityStoryDetails = (groupsWithStories: ArticleFragment[][]) => groupsWithStories.reduce((stories, storiesInGroup, index) => {
```
RichieAHB: I think I need to understand stories a bit more - not quite sure what this is doing ...

Reettaphant: Fear not, stories are just badly named articles. This has been fixed now.

PR #477, opened 2018-11-19
Parse section titles correctly when adding snap links

jonathonherbert:
... as before, we didn't look for them! Hopefully it's clear in the code what's going on.

### DIFF
--- a/client-v2/src/services/__tests__/faciaApi.spec.ts
+++ b/client-v2/src/services/__tests__/faciaApi.spec.ts
@@ -120,10 +120,7 @@ describe('faciaApi', () => {
           response: {
             status: 'ok',
             results: [],
-            tag: {
-              id: 'sectionId',
-              type: 'section',
-              webUrl: 'exampleUrl',
+            section: {
               webTitle: 'Example section title'
             }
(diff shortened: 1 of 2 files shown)


client-v2/src/services/__tests__/faciaApi.spec.ts
```diff
@@ -98,6 +99,38 @@ describe('faciaApi', () => {
...
+            }
+          }
+        })
+      ).toEqual('Example tag title');
+      expect(
+        getTagOrSectionTitle({
+          response: {
+            status: 'ok',
+            results: [],
+            tag: {
```
RichieAHB: I thiiiink this is meant to be a `section` key but I may be wrong?

RichieAHB: Hopefully I'm not wrong because that means I understand the rest of it and will approve it accordingly!

jonathonherbert: hah. ohh boy. Yes, yes it is. Thanks Rich! 👍 the system wooorks

PR #478, opened 2018-11-19
Change feed display to first published

RichieAHB:
This actually fixes a bug that we thought we'd fixed were we order by first published - not by last updated!

screenshot 2018-11-19 at 13 51 53
screenshot 2018-11-19 at 14 02 39


vlbee:
@Reettaphant  Believe this is fixed now.

Draft is ordered to match the old tool: 
```
 'order-by': 'oldest',
 'from-date': [TODAY'S DATE] 
````

PR #480, opened 2018-11-21
Add snap link drag n drop integration tests

vlbee:
**Adds 3 tests**  

1. Drag and Drop renders a Guardian Snap Link correctly
2. Drag and Drop renders a Latest Guardian Snap Link correctly
3. Drag and Drop renders an external page Snap Link correctly

**Notes** 

- for dnd, I had to mock html anchor tags in index.html
- TestCafe bug - sometimes headless tests don't pass b/c it can't find DOM nodes unless you `t.maximizeWindow()` 
- Express server endpoints are specific to the fixtures used. 

The last Express `.get` endpoint is throwing an error _I think_ because at some point it's interpreting the front id `test/front` as an endpoint and trying to serve this static file: 
`Error: ENOENT: no such file or directory, stat '/client-v2/dist/front'`

This does not seem to be affecting test behavior, and I think it's due to the implementation of the Express server endpoints. Any thoughts if needs fixing and how?

### DIFF
--- a/client-v2/integration/fixtures/snap-tag-page.js
+++ b/client-v2/integration/fixtures/snap-tag-page.js
@@ -1,0 +1,49 @@
+export default `
+<!DOCTYPE html>
+<html id="js-context" class="js-off is-not-modern id--signed-out" lang="en" data-page-path="/tone/recipes">
+<head>
+<title>Recipes | The Guardian</title>
+<meta charset="utf-8">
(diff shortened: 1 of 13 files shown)


vlbee:
If we want to remove `data-testid` from production we can use this babel plugin: https://www.npmjs.com/package/babel-plugin-react-remove-properties

PR #483, opened 2018-11-26
Refactor insertion

RichieAHB:
This moves all the insertion / deduplication logic to the reducer and swaps out various actions. There are quite a few linting changes annoyingly here which I missed when committing.

Unlike the method John and I discussed. I've reverted back to having an action for each insert type (`SHARED/INSERT_GROUP_ARTICLE_FRAGMENT` etc.) as this will allows us to customise behaviour in future for different types - currently only groups counts affect collection caps. However, there is an action creator `insertArticleFragmentWithCreate` that will compose these action creators. So if there is some behaviour that is required for all cases then it can go here.


client-v2/src/components/FrontsEdit/Front.tsx
```diff
@@ -254,10 +264,12 @@ const mapDispatchToProps = (dispatch: Dispatch) => {
     clearArticleFragmentSelection: (frontId: string) =>
       dispatch(editorClearArticleFragmentSelection(frontId)),
     removeCollectionItem: (parentId: string, uuid: string) => {
-      dispatch(removeGroupArticleFragment(parentId, uuid));
+      dispatch(removeArticleFragment('collection')('group', parentId, uuid));
```
RichieAHB: I should probably update this to be un-curried.

RichieAHB:
To go into a bit more depth on the changes, all reducers that care about article insertion / removal now listen for the same action `SHARED/[INSERT/REMOVE]_ARTICLE_FRAGMENT`. These actions have a `type` key in their payload (*not* the action `type` key at the top level of the action), that describes which type of entity is being inserted or removed so that the reducer knows if it's a removal of that type of entity and it can either remove it or do nothing.

So when an action such as:

```js
{
  type: 'INSERT_ARTICLE_FRAGMENT`,
  payload: { type: 'articleFragment', /* ... */ }
}
```

gets fired off, the reducers (`group`, `clipboard` and `articleFragment`) can all listen for that action but the only reducer to update state from that action would be the `articleFragment` reducer.

This pattern makes composing together moves, from say a `group` to a `clipboard` much easier, and means we only have one action to care about for all inserts. And _this_ means that when implementing things like collection cap modals, we can do it directly in that action for *every* time we insert an article fragment in whatever context.

The other thing I've done here, as discussed with @jonathonherbert is provide the shared state to a few reducers as a third argument, so that deduplication logic can happen in exactly the same places as the insertion logic (inside the reducer). Previously this was happening in a thunk action and multiple actions were having to get fired to tidy stuff up.

I've also added quite a few more tests!

client-v2/src/shared/selectors/shared.ts
```diff
@@ -219,6 +219,61 @@ const createDemornalisedArticleFragment = (
       }
     : { ...articleFragments[articleFragmentId] };
 
+
+// this creates a map between a group id and it's parent collection id
+// { [groupId: string]: string /* collectionId */ }
+const groupCollectionMapSelector = createSelector(
```
jonathonherbert: V. handy!

PR #490, opened 2018-12-06
Save kickers

Reettaphant:
Kickers are now saved the same way in the old tool and in the new tool and will no longer cause confusions when switching between the two tools. 

Do the selected button colours make sense in the image below? It seems to me that the selected property in the button describes more the unselected kicker state?

screen shot 2018-12-06 at 09 13 26


client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
```diff
@@ -199,20 +206,50 @@ const formComponent: React.StatelessComponent<Props> = ({
...
           useHeadlineFont
-        />
-        {kickerOptions.map(kickerOption => (
-          <Button
-            key={kickerOption}
-            type="button"
-            pill
-            onClick={(e: React.FormEvent<HTMLButtonElement>) =>
-              change('customKicker', e.currentTarget.value)
+          format={value => {
```
jonathonherbert: glad this worked as expected 👍

franziskas:
@Joshua-David

jonathonherbert:
there he is! 👋

PR #491, opened 2018-12-06
Add collection caps

RichieAHB:
This implements collection caps (and the required modal) - it relies on #483 so not going to merge before that has been reviewed. As with the old tool this only implements collection caps on collections (as the name suggests) but not on the supporting articles, and not on the clipboard.

We use plain old `redux-thunk` for this, it's perhaps a touch complex but I think it's less complex than bringing in another library. Thoughts welcome 👍 

ezgif-1-03055cc7e644

Despite the GIF the modal is in the middle, I just cropped the screen!


RichieAHB:
Closed in favour of #492

PR #493, opened 2018-12-06
Collection cap tests

RichieAHB:
Somehow these tests didn't make it through on the original PR. These do the actual testing of the group collection caps (and use the _actual_ insert action creator used at the application level). Much safe 🚑 ...

Also, because who _doesn't_ want to see a large passing test suite:

kapture 2018-12-06 at 15 19 15

### DIFF
--- a/client-v2/src/actions/__tests__/ArticleFragments.spec.ts
+++ b/client-v2/src/actions/__tests__/ArticleFragments.spec.ts
@@ -23,8 +23,12 @@ import {
   reducer as collectionsReducer,
   initialState as collectionsState
 } from 'shared/bundles/collectionsBundle';
+import confirmModal from 'reducers/confirmModalReducer';
+import { endConfirmModal } from 'actions/ConfirmModal';
+import config from 'reducers/configReducer';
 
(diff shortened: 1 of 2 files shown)


client-v2/src/actions/__tests__/ArticleFragments.spec.ts
```diff
@@ -77,25 +85,36 @@ const buildStore = (added: ArticleFragmentSpec) => {
   return createStore(root, state as any, applyMiddleware(thunk));
 };
 
-const insert = (
+const insert = async (
   [uuid, id]: [string, string],
```
jonathonherbert: what do the uuid and id represent here?

RichieAHB: They are used to build a stub article fragment seen as in these test we're just testing the way `uuid`s and `id`s affect things like deduplication / collection caps etc. It's basically a short hand for creating article fragments.

RichieAHB: Ah I get what you mean, I'll add a variable name and destructure inside the function.

PR #495, opened 2018-12-06
Add conditional forms

jonathonherbert:
Add conditional form fields. This also adds a form for snap links 🎉 

front-forms

A ConditionalField component helps to keep the form markup clean, and all of the field logic is encapsulated in a selector. I've moved some of the form functions out into a utils to keep that file manageable, too, it was getting kinda large!


client-v2/overrides.d.ts
```diff
@@ -0,0 +1,25 @@
...
+  Parser,
+  Validator,
+  GenericFieldHTMLAttributes
+} from 'redux-form';
+import { ComponentType, Component } from 'react';
+
+declare module 'redux-form' {
+  interface BaseFieldProps<P = {}> extends Partial<CommonFieldProps> {
+    name: string;
+    // Any-typed to avoid a known issue, see https://github.com/DefinitelyTyped/DefinitelyTyped/issues/26253
```
jonathonherbert: Sadly 😢 Any suggestions welcome - I didn't spend too much time on this but maybe with a pair we could do better here

client-v2/package.json
```diff
@@ -103,6 +103,7 @@
     "redux-thunk": "^2.2.0",
     "reselect": "^3.0.1",
     "styled-components": "^3.2.5",
+    "ts-optchain": "^0.1.2",
```
jonathonherbert: Couldn't resist! Any objections let me know 😄

client-v2/modules.d.ts
```diff
@@ -30,7 +30,8 @@ declare module 'panda-session' {
 }
 
 declare module 'normalise-with-fields' {
-  export const createType: any
+  export const createType: any;
```
jonathonherbert: prettier 🤷‍♀️

client-v2/package.json
Reettaphant: 👍

PR #496, opened 2018-12-06
Minimum viable collection overview

RichieAHB:
Here's a straw man PR for a MVP of the collection overview, there's a lot more to do but I worked on this. All feedback welcome!

kapture 2018-12-06 at 17 28 12

### DIFF
--- a/client-v2/src/components/FrontsEdit/CollectionOverview.tsx
+++ b/client-v2/src/components/FrontsEdit/CollectionOverview.tsx
@@ -1,0 +1,94 @@
+import React from 'react';
+import { Collection, CollectionItemSets } from 'shared/types/Collection';
+import { State } from 'types/State';
+import {
+  createCollectionSelector,
+  selectSharedState,
+  createArticlesInCollectionSelector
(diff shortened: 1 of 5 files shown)


jonathonherbert:
Oh nice, yes please do this! I was considering some state management in a parent component to handle the scrollTo behaviour, but I think the simplicity is good here. Only two points:

- Possibly we could derive the component #id in the same way each time? (sure you've thought of this already!)
- How does behavior: smooth for the scrollTo feel? Possibly it's a bit much but might make it easier for the user to orient themselves with regard to the new selection.

RichieAHB:
@jonathonherbert Point 1 - yes! I thought about having a helper but it got to the end of the day. I'll add that in for sure 👍 

The smooth scroll stuff is a bit slow - it scrolls left then down, but we can look at it for sure. Also, for larger screens that's probably not an issue 🤔

RichieAHB:
kapture 2018-12-07 at 10 10 18

@jonathonherbert Added smooth scrolling, a little hover styled and made a small helper for generating the id for the collection component.

PR #501, opened 2018-12-13
Display Collection config data in component with styling

vlbee:
Shows platform and metatag config with styling - 
("Locked" will appear in the same row, but that's another PR)

image
image

### DIFF
--- a/client-v2/src/util/frontsUtils.ts
+++ b/client-v2/src/util/frontsUtils.ts
@@ -32,15 +32,19 @@ const getFrontCollections = (
 const combineCollectionWithConfig = (
   collectionConfig: CollectionConfig,
   collection: CollectionWithNestedArticles
-): CollectionWithNestedArticles =>
-  Object.assign({}, collection, {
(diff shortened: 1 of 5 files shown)


client-v2/src/util/frontsUtils.ts
```diff
@@ -32,15 +32,18 @@ const getFrontCollections = (
 const combineCollectionWithConfig = (
   collectionConfig: CollectionConfig,
   collection: CollectionWithNestedArticles
-): CollectionWithNestedArticles =>
-  Object.assign({}, collection, {
+): CollectionWithNestedArticles => {
+  return Object.assign({}, collection, {
```
jonathonherbert: just in case you didn't know and fancied being super-concise in future, this might be expressed more briefly with object spreading and parenthesised function bodies as => ({ ...collection, { <etc> } }) 😄

vlbee: I tried to solve this but got type mismatches when the CollectionConfig is merged with Collection so will leave as is for now.

PR #506, opened 2018-12-19
Search date filters

Reettaphant:
Search filters to capi search 
screen shot 2018-12-19 at 11 45 30


client-v2/src/components/App.tsx
```diff
@@ -22,6 +22,7 @@ import FrontsEdit from './FrontsEdit/Edit';
 import Home from './Home';
 import NotFound from './NotFound';
 import ConfirmModal from './ConfirmModal';
+import '../globalStyles';
```
Reettaphant: Are we happy with this??

RichieAHB: Is there any way we could just import the css from react-dates? Just so that if we update we don't have to manually update the styles? Or are they not exported?

Reettaphant: I've added a commit doing this, let me know what you think.

PR #508, opened 2018-12-21
Remove duplicate tags correctly

Reettaphant:
Fixes a bug where adding duplicate search tag search terms removed all the tags AND simplifies the code a bit at the same time.


client-v2/src/components/FrontsCAPIInterface/SearchInput.tsx
```diff
@@ -161,12 +158,9 @@ class FrontsCAPISearchInput extends React.Component<
...
 
   public handleTypeInput = (item: any, type: SearchTypes) => {
-    let newTags = [] as string[];
-    const oldTags = this.state.selected[type];
+    const oldTags = this.state.selected[type].filter(tag => tag !== item.id);
 
-    if (item && oldTags.indexOf(item.id) === -1) {
-      newTags = oldTags.concat([item.id]);
-    }
+    const newTags = oldTags.concat([item.id]);
```
akash1810: really very minor: should we spread here, to be consistent with the object spread on line 164?

PR #510, opened 2019-01-02
Improve fronts search

jonathonherbert:
Two quick wins to improve the fronts search - 

- the text highlighting only searches for matches of two characters or more, which removes a little input lag
- the entire list item is now selectable, not just the button, which is a very small target - often I've seen users try to click the list item, and only navigate to the button once that fails to work.

front-search-improvement

### DIFF
--- a/client-v2/src/components/FrontList.tsx
+++ b/client-v2/src/components/FrontList.tsx
@@ -11,13 +11,21 @@ interface Props {
   searchString: string;
 }
 
-const ListItem = styled('li')`
+const ListItem = styled('li')<{ isActive?: boolean }>`
   position: relative;
-  padding: 10px 0;
+  padding: 10px 5px;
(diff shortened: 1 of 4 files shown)


client-v2/src/components/FrontList.tsx
```diff
@@ -53,15 +61,19 @@ const FrontList = ({ fronts, onSelect, searchString }: Props) => {
...
+        >
+          <ListLabel isActive={!front.isOpen}>
             <TextHighlighter
               originalString={front.id}
               searchString={searchString}
             />
           </ListLabel>
-          {front.isOpen && (
-            <ButtonAdd onClick={() => onSelect(front.id)}>
+          {!front.isOpen && (
```
Reettaphant: This line makes so much more sense now!

PR #514, opened 2019-01-04
Grid iframe

Reettaphant:
screen shot 2019-01-04 at 10 47 10

### DIFF
--- a/client-v2/src/shared/components/input/InputImage.tsx
+++ b/client-v2/src/shared/components/input/InputImage.tsx
@@ -169,7 +169,13 @@ class InputImage extends React.Component<Props, State> {
         >
           <ButtonDelete type="button" priority="primary">
             {this.props.input.value ? (
-              <IconDelete src={deleteIcon} onClick={this.clearField} />
+              <IconDelete
+                src={deleteIcon}
+                onClick={(event) => {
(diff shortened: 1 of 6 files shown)


client-v2/src/shared/util/validateImageSrc.ts
```diff
@@ -252,38 +242,38 @@ function getData(
     : null;
 }
 
+function validateMediaItem(crop: Crop, imageOrigin: string, criteria?: Criteria): Promise<ValidationResponse | Error> {
```
jonathonherbert: More readable 👍

client-v2/src/shared/components/input/InputImage.tsx
```diff
@@ -83,10 +85,68 @@ class InputImage extends React.Component<Props, State> {
...
+  public onMessage = (event: MessageEvent) => {
+    if (event.origin !== this.props.gridUrl) {
+      // Log: did not come from the grid
+      return;
+    }
+
+    const data: GridData = event.data;
+
+    if (!data) {
+      // TODO Log did not get data
```
jonathonherbert: For future, but we have quite a few @todos across the codebase that relate to logging -- perhaps we can send errors directly to Sentry in these places

PR #518, opened 2019-01-07
Add Uneditable/Locked Collection feature

vlbee:
This PR locks a Collection in a Front if it's marked as "no curation" in the config.  

Mouse Events/DragNDrop are disabled and collection is greyed out.

**Tests** - one added for isCollectionUneditable Selector.  I think disabled functionality would be best tested in an integration test? 

Builds on PR #501 which adds the "Special/Canonical/Breaking/Branded" meta tags and should be merged first. 

screen shot 2019-01-07 at 10 22 262


client-v2/src/shared/components/Collection.tsx
```diff
@@ -77,26 +89,58 @@ const ItemCountMeta = CollectionMetaBase.extend`
...
 `;
 
-class CollectionDetail extends React.Component<Props, { isOpen: boolean }> {
+class CollectionDisplay extends React.Component<Props, { isOpen: boolean }> {
+  public static defaultProps = {
+    isUneditable: false
+  };
   public state = {
-    isOpen: true
+    isOpen: !this.props.isUneditable
```
RichieAHB: Question: do we ever want a Collection to be opened when it's uneditable?

client-v2/src/shared/components/collectionItem/CollectionItemContainer.tsx
```diff
@@ -1,3 +1,6 @@
 import styled from 'styled-components';
 
-export default styled('div')``;
+export default styled('div')<{ isUneditable?: boolean }>`
+  pointer-events: ${({ isUneditable }) => (isUneditable ? 'none' : 'auto')};
```
RichieAHB: What does this pointer-events do here - do you remember? I think ideally we'd ensure that if somethings was `uneditable` then we would remove all event handlers etc. such that `pointer-events` would not be required. What would change if this CSS property was removed.

client-v2/src/shared/components/Collection.tsx
jonathonherbert: @RichieAHB I think we briefly discussed and came down on the side of yes -- the case being that even if users can't edit collections it's useful for them to see what's in them to make curation decisions -- does this make sense?

vlbee: CP said they wanted the ability to see what was inside a locked collection. A locked Collection is collapsed by default when it loads.

jonathonherbert: 👍

client-v2/src/shared/components/collectionItem/CollectionItemContainer.tsx
vlbee: This has zero effect that I can see so have removed it. Leftover from earlier commit - thanks for catching it!

vlbee:
Caught a bug where an unlaunched but locked collection was still showing the Launch button - it will now only only display "Locked" whether a not the collection has been launched.
image

PR #519, opened 2019-01-07
Collection caps now only care about live articles

jonathonherbert:
... because that's the way the old tool works; collection caps don't apply to draft content.

Don't merge -- the auto-removal doesn't take into account the draft article exemption yet!


jonathonherbert:
Closed -- managing removal of scheduled content complicates this logic to the extent that it fails on cost/benefit!

PR #522, opened 2019-01-09
only sudo to create the config dir

akash1810:
favour changing ownership of the config dir to the current user to avoid needing to use sudo with the aws cli to download files from s3


akash1810:
> What will the mkdir command do here if it exists if user is not the owner? Is it going to overwrite the existing directory?

`mkdir` will only create a directory if it doesn't exist and fail with a useful error `mkdir: /etc/gu: File exists`.

Reettaphant:
So if the directory exists but user is not the owner, then how can the files be added to it without using sudo or am I missing something here?

akash1810:
Yep, great question. Its really subtle.

`if [[ -w ${CONFIG_DIR} ]]; then` will check if the directory is writable (the `-w` part) by current user. If it is, great, copy files. If not, then `mkdir` and make current user the owner, then copy files.

akash1810:
Update!

`fetch-config.sh` will now try to copy files from S3 to `/etc/gu`. If it fails, either because `/etc/gu` doesn't exist or the current user doesn't have write access to it, then we check if the `--allow-creation` flag was used when running this script. If it was, create `/etc/gu` and `chown` it, else ⛔️ .

PR #525, opened 2019-01-09
Update README.md

akash1810:
link to all the Janus credentials needed

### DIFF
--- a/client-v2/README.md
+++ b/client-v2/README.md
@@ -28,7 +28,7 @@ View [setup.sh](scripts/setup.sh).
 
 ## Developing
 
-To start developing, first get [Janus credentials for `cmsFronts`](https://janus.gutools.co.uk/credentials?permissionId=cmsFronts-dev&tzOffset=0) then run:
+To start developing, first get [Janus credentials](https://janus.gutools.co.uk/multi-credentials?&permissionIds=cmsFronts-dev,capi-api-gateway,frontend-frontend-local-dev&tzOffset=0) then run:
 
 ```bash
 ./scripts/client-dev.sh
--- a/client-v2/README.md
+++ b/client-v2/README.md
@@ -28,7 +28,7 @@ View [setup.sh](scripts/setup.sh).
 
 ## Developing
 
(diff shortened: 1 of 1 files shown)


vlbee:
This new link gives me invalid permissions on Janus. This is what I've been using: `https://janus.gutools.co.uk/multi-credentials?&permissionIds=cmsFronts-dev,capi-api-gateway,media-service-dev&tzOffset=0` 

Just double checkin?

akash1810:
I'm not sure if `media-service-dev` is needed by this project. That is, I don't think there are any resources in the `media-service` account for this project.

From [this readme](https://github.com/guardian/facia-tool/blob/master/README.md#credentials) `frontend` credentials are needed to press fronts.

You're getting an error with this Janus link because our [access is different](https://github.com/guardian/janus/blob/master/app/data/Access.scala#L135-L149). I'll raise a PR to unify our access in Janus now.

akash1810:
Janus PR here - https://github.com/guardian/janus/pull/1642

RichieAHB:
@akash1810 is this good to go?

akash1810:
This is blocked by the Janus PR atm.

client-v2/README.md
```diff
@@ -28,7 +28,7 @@ View [setup.sh](scripts/setup.sh).
 
 ## Developing
 
-To start developing, first get [Janus credentials for `cmsFronts`](https://janus.gutools.co.uk/credentials?permissionId=cmsFronts-dev&tzOffset=0) then run:
+To start developing, first get [Janus credentials](https://janus.gutools.co.uk/multi-credentials?&permissionIds=cmsFronts-dev,capi-api-gateway,frontend-frontend-local-dev&tzOffset=0) then run:
```
adamnfish: The timezone UTC offset being included in the URL is going to be a problem in summer. This is a design problem with that Janus endpoint rather than a problem with this link though. It uses the timezone so that the credentials expire at 19:00. At the moment it uses the time zone offset to power this.

Leave it like this, but this has alerted me to the fact that this is an issue with Janus. I have a few ideas for improving this behaviour so keep in mind that we should revisit this link in the future...

PR #528, opened 2019-01-09
Make a single request for multiple collections

jonathonherbert:
... which makes fetching large fronts slightly quicker -- in our (very thorough and scientific) tests, opening the tool with CODE/UK loaded saw an 8% speed boost from page load to idle.

It also makes the behaviour slightly more predictable for the user -- there's a longer wait for the initial collections, but less jank in between as the view isn't constantly thrashing as new collections arrive.

### DIFF
--- a/app/controllers/FaciaToolController.scala
+++ b/app/controllers/FaciaToolController.scala
@@ -41,25 +41,40 @@ class FaciaToolController(
         Ok(Json.toJson(configJson)).as("application/json")}}}
 
   def getCollection(collectionId: String) = AccessAPIAuthAction.async { implicit request =>
-
     val collectionPriorities = configAgent.getFrontsPermissionsPriorityByCollectionId(collectionId)
 
     withModifyGroupPermissionForCollections(collectionPriorities, Set()) {
+      val collections = fetchCollections(List(collectionId))
 
-      FaciaToolMetrics.ApiUsageCount.increment()
-      val collection = frontsApi.amazonClient.collection(collectionId).flatMap { collectionJson =>
-        collectionJson.map(json => mediaServiceClient.addThumbnailsToCollection(json, collectionId)) match {
-          case Some(f) => f.map(Some(_))
-          case None => Future.successful(None)
-
-        }
-      }
-      collection.map(c => NoCache {
+      collections.head.map(c => NoCache {
         Ok(Json.toJson(c)).as("application/json")
       })
-
     }
   }
+
+  def getCollections(collectionIdsString: String) = AccessAPIAuthAction.async { implicit request =>
+    val collectionIds = collectionIdsString.split(",").toList
+    val collectionPriorities = collectionIds.flatMap(configAgent.getFrontsPermissionsPriorityByCollectionId(_)).toSet
+
+    withModifyGroupPermissionForCollections(collectionPriorities, Set()) {
+      val collections = fetchCollections(collectionIds)
+
+      Future.sequence(collections).map(c => NoCache {
+        Ok(Json.toJson(c)).as("application/json")
+      })
+    }
+  }
+
+  def fetchCollections(collectionIds: List[String]) = {
+    FaciaToolMetrics.ApiUsageCount.increment()
+    collectionIds.map(collectionId => frontsApi.amazonClient.collection(collectionId).flatMap { collectionJson =>
+      collectionJson.map(json => mediaServiceClient.addThumbnailsToCollection(json, collectionId)) match {
+        case Some(f) => f.map(Some(_))
+        case None => Future.successful(None)
+      }
+    })
+  }
+
 
   def publishCollection(collectionId: String) = AccessAPIAuthAction.async { implicit request =>
 
--- a/app/controllers/FaciaToolController.scala
+++ b/app/controllers/FaciaToolController.scala
@@ -1,7 +1,7 @@
 package controllers
 
 import _root_.util.Acl
-import com.gu.facia.api.models.TrainingPriority
+import com.gu.facia.client.models.{CollectionJson}
 import com.gu.facia.api.models.faciapress.{Draft, FrontPath, Live, PressJob}
 import com.gu.facia.client.models.Metadata
 import com.gu.pandomainauth.action.UserRequest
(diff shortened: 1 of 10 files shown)


app/controllers/FaciaToolController.scala
```diff
@@ -41,26 +41,41 @@ class FaciaToolController(
...
-      FaciaToolMetrics.ApiUsageCount.increment()
-      val collection = frontsApi.amazonClient.collection(collectionId).flatMap { collectionJson =>
-        collectionJson.map(json => mediaServiceClient.addThumbnailsToCollection(json, collectionId)) match {
-          case Some(f) => f.map(Some(_))
-          case None => Future.successful(None)
-
-        }
-      }
-      collection.map(c => NoCache {
+      collections.head.map(c => NoCache {
```
LATaylor-guardian: `.head` here is potentially unsafe if the List is empty. Can we use `.headOption` and handle accordingly?

app/controllers/FaciaToolController.scala
```diff
@@ -41,26 +41,41 @@ class FaciaToolController(
...
         Ok(Json.toJson(c)).as("application/json")
       })
+    }
+  }
+
+  def getCollections(collectionIdsString: String) = AccessAPIAuthAction.async { implicit request =>
```
LATaylor-guardian: Can this simply be called `collectionIds`?

LATaylor-guardian: Public functions should have return types ideally.

app/controllers/FaciaToolController.scala
jonathonherbert: thanks for catching this, Luke!

app/controllers/FaciaToolController.scala
```diff
@@ -41,26 +41,41 @@ class FaciaToolController(
...
+    withModifyGroupPermissionForCollections(collectionPriorities, Set()) {
+      val collections = fetchCollections(collectionIds)
 
+      Future.sequence(collections).map(c => NoCache {
+        Ok(Json.toJson(c)).as("application/json")
+      })
     }
   }
 
+  def fetchCollections(collectionIds: List[String]) = {
```
LATaylor-guardian: Return type needed here also.

app/controllers/FaciaToolController.scala
```diff
@@ -41,26 +41,41 @@ class FaciaToolController(
...
+    FaciaToolMetrics.ApiUsageCount.increment()
+    collectionIds.map(collectionId => frontsApi.amazonClient.collection(collectionId).flatMap { collectionJson =>
+      collectionJson.map(json => mediaServiceClient.addThumbnailsToCollection(json, collectionId)) match {
+        case Some(f) => f.map(Some(_))
```
LATaylor-guardian: What is `f` here?

app/controllers/FaciaToolController.scala
jonathonherbert: it's named to disambiguate it from the collectionIds List that follows on line 56, so it needs to be called something else, but perhaps there's a better name?

Ideally we'd be acting on a List[String] to start with -- we discussed binding query params in the route but weren't sure if this was possible, any help welcomed :)

jonathonherbert: 👍 will do!

app/controllers/FaciaToolController.scala
```diff
@@ -41,26 +41,41 @@ class FaciaToolController(
...
 
+      Future.sequence(collections).map(c => NoCache {
+        Ok(Json.toJson(c)).as("application/json")
+      })
     }
   }
 
+  def fetchCollections(collectionIds: List[String]) = {
+    FaciaToolMetrics.ApiUsageCount.increment()
+    collectionIds.map(collectionId => frontsApi.amazonClient.collection(collectionId).flatMap { collectionJson =>
```
LATaylor-guardian: `amazonClient` here is still making multiple requests. Can this make a bulk request? This may give you a similar optimisation as you intended with the call to your own server.

app/controllers/FaciaToolController.scala
```diff
@@ -41,26 +41,41 @@ class FaciaToolController(
...
+      collectionJson.map(json => mediaServiceClient.addThumbnailsToCollection(json, collectionId)) match {
```
LATaylor-guardian: As above.

app/controllers/FaciaToolController.scala
LATaylor-guardian: Ah I see - this link looks useful: https://stackoverflow.com/questions/36562330/query-parameter-as-list

But may be overkill for what you're trying to do. Worth exploring.

LATaylor-guardian: Repeating the query param gives you a simple solution given you control the client request.

jonathonherbert: Yes, much prefer to handle query param -- thanks for this.

app/controllers/FaciaToolController.scala
jonathonherbert: I'd love toI did some searching before opening the PR and every answer to the question 'can S3 handle requests for multiple files' was met with a negative. Admittedly most of the replies are old. I've asked around and can't find an answer to the contrary, but if there was a way I'd be delighted to use it!

app/controllers/FaciaToolController.scala
jonathonherbert: 👍

app/controllers/FaciaToolController.scala
jonathonherbert: Yes, it's badly named -- I've made it clearer in an incoming commit.

app/controllers/FaciaToolController.scala
jonathonherbert: This code seems to be somewhat vestigial in that I can't find a situation in which it makes a HTTP call at runtime. You're right, though -- it's a nested call, and batching would be preferable. I'll see if I can figure out if it's still necessary.

client-v2/src/actions/Collections.ts
```diff
@@ -2,7 +2,7 @@ import uniq from 'lodash/uniq';
 import { batchActions } from 'redux-batched-actions';
 import {
   getArticlesBatched,
-  getCollection as fetchCollection,
+  getCollections as fetchCollection,
```
RichieAHB: Tiny tiny thing but maybe just rename this to `fetchCollections` it confused me for longer than it should have 😊

jonathonherbert: Definitley -- I think there's another PR to write that renames all of our 'gets' to 'fetches'!

PR #529, opened 2019-01-09
Move search params state to redux

RichieAHB:
This PR moves the search state from a rather odd, un-idiomatic way of managing state into redux.

# Why do this?
This was required because, in order to refresh the feed using a button _outside_ the component, the state of the search fields required for the CAPI params needs to be lifted to at least this level in the component tree. There will be a subsequent PR to add the refresh button.

I could have lifted the state further up the tree _without_ using redux but given the search will potentially undergo a bit of change (including UX) it seemed more sensible to just have it outside the tree.

# Problems with this?
The initial aim of the search was to have it be a drop-in component that could be used elsewhere. Having it depend on redux seems unnecessary. However, until we have something useful then optimising for this case doesn't seem like a priority at the moment. Additionally, if we wanted to remove the external dependency on redux, bringing in the state to the top level component - even using redux internally - would be relatively trivial.

# Redux boilerplate
I've used `createAsyncResourceBundle` for boilerplating the redux stuff, it's not a *perfect* fit but I think it saves time. Might be worth catching up with @jonathonherbert on my usage here! (I've also added a small change in order to allow setting / removing `loadingIds` manually for bundles that aren't indexed by id).

# Preview / Live feeds
As it stands, `preview` and `live` requests are being made at the same time - regardless of what is on screen. I could change this but it actually required more of a refactor, be interested to hear whether people think this is confusing / undesirable.

# Tests
There's quite a bit of threading about here, for what is essentially some basic state updating. I've not added tests as it would probably take quite a lot of work (again largely down to the size of the components) but happy to discuss this and find places to test here.

# Other changes
I've touched a tiny bit more than I needed to in this PR as these components are quite big so I broke them out in a couple of places to help some of my reasoning. I'm not sure how much more we want to touch search this quarter but there's still more that can be done to make this a slightly less daunting place in the future!

**Edit**
I also have missed adding back in the error UI for the CAPI feed which I'll do before I merge.


client-v2/src/components/FrontsCAPIInterface/SearchInput.tsx
```diff
@@ -1,124 +1,130 @@
...
+  if (!date) {
+    return 'Not selected';
+  }
+  return date.format('DD/MM/YYYY');
 };
 
-const emptyState = {
-  q: undefined,
-  isResource: false,
-  searchTerms: emptySearchTerms,
```
RichieAHB: I took search terms out from the parent and left them in their respective child components just to thin this component out a bit. Potentially could be reverted to how it was but they weren't currently being used so it was just another thing to handle here.

client-v2/src/components/FrontsCAPIInterface/SearchInput.tsx
```diff
@@ -131,256 +137,211 @@ class FrontsCAPISearchInput extends React.Component<
...
+      this.debouncedRunSearchAndRestartPolling();
+    });
+  }
 }
 
-export { AsyncState, CAPISearchQueryReponse };
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  fetchFeed: (query: string, liveParams: object, previewParams: object) => {
+    const maybeArticleId = getIdFromURL(query);
+    dispatch(fetchLive(liveParams, !!maybeArticleId));
```
RichieAHB: This is where we fetch both at the same time. If there were a sub-component that handled some of this stuff then this wouldn't be necessary and we could handle each feed separately but again as mentioned in the PR. Happy to go with whatever people think makes sense.

client-v2/src/components/FrontsCAPIInterface/FieldFilter.tsx
```diff
@@ -65,17 +62,16 @@ const FilterContainer = styled('div')`
 
 // The extension here is the result of JSX ambiguity -
 // see https://github.com/Microsoft/TypeScript/issues/4922.
-const CAPIFieldFilter = <T extends any>({
+const CAPIFieldFilter = <T extends { label: string; id: string }>({
```
jonathonherbert: 💯

client-v2/src/lib/createAsyncResourceBundle/index.ts
```diff
@@ -272,12 +277,14 @@ function createAsyncResourceBundle<Resource>(
...
-              ? removeStatusIds(
-                  state.loadingIds,
-                  getStatusIdsFromData(action.payload.data)
-                )
-              : []
+            loadingIds: removeStatusIds(
+              state.loadingIds,
+              indexById
+                ? getStatusIdsFromData(action.payload.data)
+                : action.payload.loadingId
```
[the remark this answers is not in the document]
jonathonherbert: yes, I see what you mean! I wonder whether it'd be better to decouple the reducer from the rest of the code here to account for different resource models, but in the mean time this is good 👍

client-v2/src/index.tsx
```diff
@@ -11,11 +11,19 @@ import { configReceived } from 'actions/Config';
...
 import { storeClipboardContent } from 'actions/Clipboard';
 import { Dispatch } from 'types/Store';
+import capiQuery from 'services/capiQuery';
+import CAPIServiceContext from './components/CAPIServiceContext';
 import Modal from 'react-modal';
 
-const store = configureStore();
 const config = extractConfigFromPage();
 
+const capiServices = {
```
jonathonherbert: To ask a possibly silly question - is this context still necessary? I can't remember if capiQuery carries state or not, but if not -- might this be better expressed as a statically imported service?

client-v2/src/lib/createAsyncResourceBundle/index.ts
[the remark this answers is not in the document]
RichieAHB: I was wondering the same thing - I don't think it'd be too hard to rewind this if we think of a better way. It's definitely not perfect but I think for now it should be ok.

client-v2/src/index.tsx
RichieAHB: As discussed IRL - will rip this out and have it static and ignore config option 👍

client-v2/src/lib/createAsyncResourceBundle/index.ts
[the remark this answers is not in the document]
RichieAHB: I've reverted this change now and added a bundle for each feed - much cleaner 🚿

PR #530, opened 2019-01-10
Revert "Move search params state to redux (#529)"

RichieAHB:
This reverts commit ae270d73bd39935424be64618e1bdc75bd97d5f5.

I hard coded our CAPI urls but overlooked that these need to be added in on a per-stage basis. I can revert this commit, undo that change and raise a new PR.


RichieAHB:
I can actually fix this quickly so will not revet.

PR #533, opened 2019-01-10
Open top <n> collections on front open

jonathonherbert:
For performance reasons, only open the top <n> collections when a front is opened.

Clicking a collection in the right-hand overview also opens a collection, so navigation along the happy path (via overview) remains snappy.

Even without conditionally loading content (the total number of requests to the server remains the same), this makes a serious difference to the performance of startup for fronts with large numbers of collections, e.g. the UK front -- it loads on average almost three times faster on CODE, from ~15 seconds to ~6secs from hard refresh.

### DIFF
--- a/client-v2/src/bundles/__tests__/frontsUIBundle.spec.ts
+++ b/client-v2/src/bundles/__tests__/frontsUIBundle.spec.ts
@@ -2,7 +2,9 @@ import reducer, {
   editorOpenFront,
   editorCloseFront,
   editorClearOpenFronts,
-  editorSetOpenFronts
+  editorSetOpenFronts,
+  editorOpenCollections,
+  editorCloseCollections
 } from '../frontsUIBundle';
 
 describe('frontsUIBundle', () => {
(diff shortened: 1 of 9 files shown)


client-v2/src/bundles/__tests__/frontsUIBundle.spec.ts
```diff
@@ -32,5 +34,28 @@ describe('frontsUIBundle', () => {
       );
       expect(state.frontIds).toEqual(['front1', 'front3']);
     });
+    it('should add a collection to the open editor collections', () => {
```
jonathonherbert: Opening with some tests, bam 💥

client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
```diff
@@ -87,16 +98,17 @@ const createMapStateToProps = () => {
...
-  return {
-    publishCollection: (id: string, frontId: string) =>
-      dispatch(publishCollection(id, frontId))
-  };
-};
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  publishCollection: (id: string, frontId: string) =>
+    dispatch(publishCollection(id, frontId)),
+  onChangeOpenState: (id: string, isOpen: boolean) =>
+    dispatch(isOpen ? editorOpenCollections(id) : editorCloseCollections(id))
```
jonathonherbert: Bit of a funny one, but the API matches the fronts UI and is quite useful elsewhere!

PR #534, opened 2019-01-10
Loading placeholders for containers

jonathonherbert:
'cos right now we are a bit gross!

Before --
screen shot 2019-01-10 at 15 32 37

After -- 
screen shot 2019-01-10 at 15 34 26


client-v2/src/lib/createAsyncResourceBundle/index.ts
```diff
@@ -187,7 +187,7 @@ function createAsyncResourceBundle<Resource>(
   const selectIsLoadingById = (state: RootState, id: string) =>
     selectLocalState(state).loadingIds.indexOf(id) !== -1;
 
-  const selectById = (state: RootState, id: string): Resource =>
+  const selectById = (state: RootState, id: string): Resource | undefined =>
```
jonathonherbert: This was a bit of an elision before, so I've corrected it here.

PR #537, opened 2019-01-11
Convert Hex to style-components colour theme in Shared/Components

vlbee:
## Spot the difference: 
colors

## What has changed?

All hard coded Hex values within Shared/Components have been replaced with a styled-components theme. 

The relevant file: `client-v2/src/shared/constants/theme.ts`
I decided to code colour names so things are easier to reason about for those not good with Hex values.  

Base colors applies to things like Collection items. 
Inputs -> form fields and toggle inputs
Buttons -> Buttons
 
I will roll this out to the non-shared components in another PR.


vlbee:
Just a note:  the idea with the Theme was to have a an easy to understand Color Pallette using real color names, but then apply colors by function such as Text, Border etc. 

One can easily use the BASE colors to retheme everything quckly.

The INPUT and BUTTON themes are seperated out so that it's easier to apply colors when creating components. 

I think perhaps also having separate a COLLECTION theme would be good, so that you don't really need to go to the BASE often when applying colors by component type, but this could all be over kill!   Please let me know what you thing would be simpler or better.

client-v2/src/shared/components/BasePlaceholder.tsx
```diff
@@ -13,12 +13,12 @@ injectGlobal`
 `;
 
 export default styled('div')`
```
jonathonherbert: I think this, and possibly other files, are still using the base `styled` function -- difficult to spot as there's no red ink, but this will mean that .props.theme is `any` typed! If that's right, it'd be worth replacing anything that uses the base `styled` function before merging.

client-v2/src/shared/constants/theme.ts
```diff
@@ -1,23 +1,63 @@
...
+import baseStyled, { ThemedStyledInterface } from 'styled-components';
+
+const colors = {
+  blackDark: '#121212', // darkest
+  blackLight: '#333',
+  greyDark: '#515151',
+  greyMedium: '#999999',
+  greyLight: '#A9A9A9',
+  greyLightPinkish: '#C9C9C9',
+  whiteFour: '#DCDCDC',
```
jonathonherbert: Ahh, the eternal struggle 😄I agree there are lots of whites, and I'm having a pointless struggle with myself over whether it's worth having a `lightest`, `lighter`, `light`, which might be easier for developers to use without checking the hexes. I'd be interested to hear @Reettaphant and @RichieAHB's thoughts!

client-v2/src/shared/components/input/InputImage.tsx
```diff
@@ -19,7 +22,10 @@ const ImageContainer = styled('div')<{
   width: 100%;
   max-width: ${props => (props.size === 'small' ? '100px' : '180px')};
   height: ${props => (props.size === 'small' ? '60px' : '115px')};
-  background-color: ${props => (props.isHovering ? '#bbb' : '#ccc')};
+  background-color: ${props =>
+    props.isHovering
+      ? props.theme.base.colos.placeholderLight
```
RichieAHB: There looks like a typo here, which suggests that some types aren't getting through maybe??

client-v2/src/shared/constants/theme.ts
RichieAHB: I guess the only answer here is what makes sense to a reader most. For me I think I'd prefer `light[er]` but then given we're just a command-click away maybe it's not an issue.

client-v2/src/shared/components/BasePlaceholder.tsx
vlbee: Good catch! This has been fixed in the shared components. It might be something to flag in the readme? I have also added it as a note in the Theme.ts file.

client-v2/src/shared/constants/theme.ts
vlbee: I've changed the numbers in the whites to light, medium, dark.  
I was also thinking if might be useful to add a styleguide somewhere to the documentation?

 
screen shot 2019-01-16 at 11 24 55
screen shot 2019-01-16 at 11 27 19
screen shot 2019-01-16 at 11 28 16
screen shot 2019-01-16 at 11 29 52

client-v2/src/shared/components/input/InputImage.tsx
vlbee: This has been fixed with the strongly typed theme swapped in throughout.

client-v2/src/shared/components/BasePlaceholder.tsx
jonathonherbert: Ace! `flag in the readme` -- great idea.

client-v2/src/shared/components/input/InputImage.tsx
jonathonherbert: this should be resolved by the changes @vlbee has made to imports (see conversation around BasePlaceholder.tsx), which will provide the right types 😄

jonathonherbert: got there just before me! 👍

PR #539, opened 2019-01-14
Only fetch collections when they're open

jonathonherbert:
Does pretty much what it says on the tin -- only fetch collections when they're opened. This might be when the front loads, or it might be when a user opens a previously unopened collection.

We always fetch a new collection on open, as its contents may have changed (polling will solve the problem of long-opened collections). We could have a staleness check or debounce if this causes problems, for example if a user rapidly toggles a collection open and closed.

The large file count is partly due to the additional files for fixtures -- I got a bit carried away scaffolding testing the 'initialiseFront' thunk! Sorry all.


client-v2/src/util/configureStore.ts
```diff
@@ -15,25 +11,26 @@ import {
   persistClipboardOnEdit,
   persistOpenFrontsOnEdit
 } from './storeMiddleware';
+import { State } from 'types/State';
 
-export default function configureStore() {
+export default function configureStore(initialState: State) {
```
jonathonherbert: in tests, this lets us inject fixture state into the store as it's used in the application, for end-to-end thunk testing. We've used mock-store previously, but I think I'm beginning to prefer this approach as it's as close to the in-app setup as possible. Interested in any thoughts 😄

client-v2/src/types/Config.ts
```diff
@@ -24,8 +24,7 @@ interface Config {
   sentryPublicDSN: string,
   mediaBaseUrl: string,
   apiBaseUrl: string,
-  stage: string,
-  switches: unknown,
+  switches: {[key: string]: boolean},
```
jonathonherbert: The type here was previously `unknown`, which was breaking store initialisation with state -- this is closer to the type as it stands IRL.

client-v2/src/actions/Collections.ts
```diff
@@ -35,69 +35,88 @@ import { getCollectionConfig } from 'selectors/frontsSelectors';
...
-        const {
-          normalisedCollection,
-          articleFragments,
-          groups
-        } = normaliseCollectionWithNestedArticles(
-          collectionWithDraftArticles,
-          collectionConfig
-        );
+      const collections = await fetchCollection(collectionIds);
+      await Promise.all(
```
RichieAHB: Is the only change in this area the removal of `flatten` 😨

jonathonherbert:
Thanks @RichieAHB, yes, as discussed something to enforce a `prettier` pass would be ideal 👍

RichieAHB:
@jonathonherbert just realised that this also trriggers an update on any Overview click - hadn't appreciated that at the time and perhaps you had but I thought it was worth calling out.

PR #540, opened 2019-01-14
Don't hydrate collection api from Grid

akash1810:
As I understand it, the `GET collection` will hydrate the `collection.json` with Grid image URLs from a URL in the json blob. I think a better solution to doing this would be to ensure the URL in the `collection.json` blob is correct at time of writing as this would stop us calling Grid each time we GET a collection. Write once, read many.

I'm not quite sure how/where we write an image url to the json blob... I need to double check we're doing a sensible thing here before merging this in.


akash1810:
It looks like we're [filtering out small crops already](https://github.com/guardian/facia-tool/blob/master/public/src/js/constants/article-meta-fields.js#L42-L65), this would be choosing the 500px wide crop.

PR #545, opened 2019-01-15
Revert back to default trail, headline and byline when they are erased

Reettaphant:
Fixes the problem described here: https://trello.com/c/yT3Ffhup/390-hi-had-a-big-scary-one-today-on-the-old-front-when-youd-backspace-completely-on-a-headline-it-would-autofill-it-back-in-so-today

Also adds in the correct placeholders from capi if the whole field gets removed.


Reettaphant:
Yes!

PR #547, opened 2019-01-16
Actually init analytics

RichieAHB:
So two boo-boos here.

1. It seems the tracking needs to be set up in the head - not quite sure why but the network requests weren't going out until I did this.
2. I didn't actually call `init` anyway 🙄 

I can see the network requests going out in dev now but still can't test in GA because I _still_ haven't found anyone who can give me enough access to setup a dev account in GA.

Hopefully this fixes the analytics 🤞

### DIFF
--- a/app/views/V2App/app.scala.html
+++ b/app/views/V2App/app.scala.html
@@ -1,26 +1,18 @@
 @(title: String, jsLocation: String, faviconLocation: String, clientConfigJson:
-String) <!DOCTYPE html>
+String)
+<!DOCTYPE html>
 <html lang="en">
-    <head>
-        <meta charset="UTF-8" />
-        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-        <meta http-equiv="X-UA-Compatible" content="ie=edge" />
-        <link
-            rel="icon"
-            type="image/png"
-            sizes="32x32"
-            href="@{faviconLocation}favicon-32x32.png"
-        />
-        <link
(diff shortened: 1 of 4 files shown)


app/views/V2App/app.scala.html
```diff
@@ -35,17 +27,25 @@
                 width: 100%;
             }
         </style>
-        <!-- Global site tag (gtag.js) - Google Analytics -->
-        <script async src="https://www.googletagmanager.com/gtag/js?id=UA-78290349-2"></script>
-    </head>
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-78290349-2"></script>
```
RichieAHB: Lines 30-37 are the only meaningful change in this file

client-v2/src/index.tsx
```diff
@@ -12,6 +12,9 @@ import { editorSetOpenFronts } from 'bundles/frontsUIBundle';
 import { storeClipboardContent } from 'actions/Clipboard';
 import { Dispatch } from 'types/Store';
 import Modal from 'react-modal';
+import { init as initGA } from 'services/GA';
+
+initGA();
```
jonathonherbert: 🎉

RichieAHB:
I've made some changes here to fix tests breaking because of missing `gtag` on the global object in node. I think this actually a better solution on the whole tbh.

PR #550, opened 2019-01-17
Add integration tests to travis

Reettaphant:
This adds about 1 minute to our build times but I think it's worth it.


RichieAHB:
> This adds about 1 minute to our build times

I wish you'd just stopped here as if this was the benefit

PR #551, opened 2019-01-17
Better messaging for CAPI errors

jonathonherbert:
At the moment, there are quite a few different errors that requests to CAPI can produce, and although we catch them, the messaging isn't as clear as it might be.

This PR adds what I hope is clearer messaging for the different scenarios, and some tests to catch regressions.

We now repeat this logic in two places -- once for the feed, and another time for the article fetches via the faciaApi service. For the sake of velocity I haven't made an effort to DRY up this code, but I've added a trello card and we can choose to prioritise this as we'd like it.

### DIFF
--- a/client-v2/src/bundles/capiFeedBundle.ts
+++ b/client-v2/src/bundles/capiFeedBundle.ts
@@ -43,7 +43,7 @@ export const fetchLive = (
   try {
     results = await fetchResourceOrResults(liveCapi, params, isResource);
   } catch (e) {
-    dispatch(liveActions.fetchError(e.message));
+    return dispatch(liveActions.fetchError(e.message));
   }
 
   if (results) {
@@ -62,7 +62,6 @@ export const fetchPreview = (
   } catch (e) {
     dispatch(previewActions.fetchError(e.message));
   }
-
   if (results) {
     dispatch(previewActions.fetchSuccess(results));
   }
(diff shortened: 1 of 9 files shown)


client-v2/src/bundles/capiFeedBundle.ts
```diff
@@ -43,7 +43,7 @@ export const fetchLive = (
   try {
     results = await fetchResourceOrResults(liveCapi, params, isResource);
   } catch (e) {
-    dispatch(liveActions.fetchError(e.message));
+    return dispatch(liveActions.fetchError(e.message));
```
RichieAHB: Good catch 👏

client-v2/src/services/capiQuery.ts
```diff
@@ -31,9 +31,56 @@ interface CAPIOptions {
...
+  TCAPIResponse extends CAPISearchQueryReponse | CAPITagQueryReponse
+>(
+  fetch: Fetch = window.fetch,
+  request: string
+) => {
+  let response: Response;
+  try {
+    response = await fetch(request);
+  } catch (e) {
+    if (e.status && e.statusText) {
```
RichieAHB: This looks like it expects a thrown `Response` object - is that a thing?

jonathonherbert: This is a really interesting question, because that's what I _get_ -- but now that you'd said this, it's not what the spec seems to say (https://fetch.spec.whatwg.org/#fetch-method). All of the rejections seem to be passed descendants of an Error type.

I'm ... a bit confused. I'd be surprised if whatwg-fetch was behaving in a non-spec way, so it's probably something I'm not doing right in the tests -- I'll take a look.

jonathonherbert: -- turns out we're using pandaFetch here, which is _not_ spec equivalent to window.fetch 🙃

PR #552, opened 2019-01-17
add sentry to v2

akash1810:
Also add SSA tags to sentry config so we can use one sentry property for both apps. This is mainly because I don't have permission to create new properties in Sentry, but it works quite well as the config is simple and we can use saved searches in Sentry to filter by stage.

This relies on local config having `stage` set to `DEV` (run a `./fetch-config.sh` to update your local config).

### DIFF
--- a/client-v2/src/index.tsx
+++ b/client-v2/src/index.tsx
@@ -20,8 +20,16 @@ const store = configureStore();
 const config = extractConfigFromPage();
 
 // publish uncaught errors to sentry.io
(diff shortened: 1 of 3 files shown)


akash1810:
> This looks great, one question - why have we changed `stage` to `env`?

We (incorrectly imo) call it `env` on the server - https://github.com/guardian/facia-tool/blob/master/app/controllers/DefaultsController.scala#L16. Probably want to change that, but in its own PR.

PR #553, opened 2019-01-17
Re-add linting

RichieAHB:
Mainly config (this can be seen in [this commit](https://github.com/guardian/facia-tool/pull/553/commits/456e382aea990be145b92038c272a70bee3fb4b8)) - there are some odd lint rules that don't seem idiomatic to me. Will flag any I see and we can discuss, but not too much discussion - prettier is our master!


client-v2/src/components/DropZone.tsx
```diff
@@ -37,8 +37,8 @@ class DropZone extends React.Component<
     return !!this.props.disabled
       ? false
       : typeof this.props.override === 'boolean'
-        ? this.props.override
-        : this.state.isHoveredOver;
+      ? this.props.override
```
RichieAHB: This one seems less readable to me - surprised to see it here! Thoughts?

RichieAHB: This is actually the only one I question ...

RichieAHB: Maybe I just shouldn't use nested ternary operators ...

PR #554, opened 2019-01-17
Show Collection status and flags in Collection Overview

vlbee:
### What does this change?
Collection metadata are now displayed in the overview - 

- Locked Flag
- Warning Flag -> Unlaunched or Unsaved Article changes (💡tooltips can be used to specify)
- Text info -> time since last update and if Backfilled

❗️TODO

- Liberties were taken with the UX/UI so needs UX to review
- Missing Unit Tests for  `isCollectionBackfilledSelector` and `hasUnsavedArticleEdits` selector and will add before merging.

screen shot 2019-01-17 at 17 14 46


harpreetpurewal:
Can we get a cute lil padlock where the 'L' is @akemitakagi or Josh should be able to supply.

Also what is the case use for unsaved? Closing the edit panel should discard changes, so not sure when something unsaved would be flagged.

client-v2/src/components/FrontsEdit/CollectionOverview.tsx
```diff
@@ -69,19 +90,76 @@ const TextRight = styled.div`
...
+  isUneditable,
+  isBackfilled,
+  lastUpdatedTimeStamp
+}: CollectionStatusProps) => (
+  <>
+    <LastUpdated>{`${distanceFromNow(
+      lastUpdatedTimeStamp as number
+    )} ago`}</LastUpdated>
+
+    {!!isBackfilled ? <LastUpdated> | Backfill</LastUpdated> : null}
```
RichieAHB: As discussed IRL here you don't need the `!!` throughout here (not that it will change anything but just worthy of note!)

vlbee:
> Can we get a cute lil padlock where the 'L' is @akemitakagi or Josh should be able to supply.

Padlock has been changed: 
screen shot 2019-01-22 at 23 39 53

PR #555, opened 2019-01-18
Exclude labs and hosted content

Reettaphant:
Talked to Harpreet about this and instead of adding a commercial label to commercial content we should just remove it from the capi feed altogether and allow it to be added elsewhere in the tool if someone so wishes for whatever reason. 

The implementation below is not ideal - the best solution would be to filter commercial content out in capi. Such capability does exist but it filters out hosted content, labs content and supported content. We don't want to filter out supported content so we can't use this option. However, I think we can get the filter we want added to capi later so this pr would only be an interim measure.


Reettaphant:
Yes I thought about that but then we'd need new capi endpoints for capi search, in case someone explicitly adds commercial content to a collection we want to display it and get a response from capi. Probably less work to add same code to v1 client if we deem it's worth it, before last week I hadn't heard of anyone adding commercial content to fronts.

PR #557, opened 2019-01-18
Live and previously are uneditable

Reettaphant:
Somehow the code which prevented people from editing live and previously stages of articles got lost somewhere causing some confusion on the opinion desk as we aren't launching live changes. This pr adds the logic back in. The styling here at the moment is the same as it is for collections which are locked - how do people feel about this? Does it make it harder to view what the live content on a collections is.
screen shot 2019-01-18 at 15 47 02

### DIFF
--- a/client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
+++ b/client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
@@ -7,7 +7,7 @@ import Button from 'shared/components/input/ButtonDefault';
 import { AlsoOnDetail } from 'types/Collection';
 import { publishCollection } from 'actions/Fronts';
 import { hasUnpublishedChangesSelector } from 'selectors/frontsSelectors';
-import { isCollectionUneditableSelector } from 'selectors/collectionSelectors';
+import { isCollectionLockedSelector } from 'selectors/collectionSelectors';
 import { State } from 'types/State';
 import { CollectionItemSets, Group } from 'shared/types/Collection';
 import {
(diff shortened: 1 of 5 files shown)


jonathonherbert:
I've approved this because we may want to ship it quickly, but I think we should be distinguishing between 'locked' and 'uneditable', which mean different things! So I thought I'd mention that here. It's purely a messaging issue but conflating the two concepts might be confusing to users, who might ask `why is my 'unlocked' collection marked as 'locked' in live? Can I still edit it?`

client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
```diff
@@ -54,37 +55,43 @@ const Collection = ({
...
+  const isUneditable =
+    isCollectionUneditable || browsingStage !== collectionItemSets.draft;
+
+  return (
+    <CollectionDisplay
+      frontId={frontId}
+      id={id}
+      browsingStage={browsingStage}
+      isUneditable={isUneditable}
+      isLocked={isCollectionUneditable}
```
akash1810: The naming here is a bit confusing. The prop is `locked` but the value is `uneditable`, there is also `isUneditable` above which makes it slightly more confusing...

Not sure what a better naming looks like though...

Reettaphant: I need to run this by Akemi before merging to make sure this looks ok but if it does, re naming of variables, I'd be inclined to just call `isCollectionUneditable` `isCollectionLocked` instead and `isUneditable` could be called `isStageUneditable`. There are quite a few collection info types of things we display on collections in addition to this, coming from the fronts config so I'd be worried that just referring to `infoText` is too generic.

RichieAHB:
@Reettaphant you and I seem to be suffering with the same failing test as on #565 - any clues?

Reettaphant:
@RichieAHB That failing test should be fixed on master now, I hope you haven't had any more problems!

PR #560, opened 2019-01-21
Open a closed Collection by dragging and hovering an ArticleFragment over it

vlbee:
@RichieAHB and I paired on this feature.

### What does this change? 

The user can now drag Article Fragments over closed Collections to open them automatically after a 300ms delay.
kapture 2019-01-21 at 17 26 23

A discrete component call `DragIntentContainer` listens for drag events and invokes a function passed in as `onIntentConfirm` prop. It manages drag events fired within nested child nodes to determine the user's intention. 

We use this component to wrap the entire `CollectionMetaContainer` styled component in Collection to give the user a  much larger drag and drop target and create a better UX experience. 

### Tests
Unit tests have been added for the `DragIntentContainer`

### DIFF
--- a/client-v2/src/shared/components/DragIntentContainer.tsx
+++ b/client-v2/src/shared/components/DragIntentContainer.tsx
@@ -1,0 +1,77 @@
+import React from 'react';
+
+type Props = {
+  onIntentConfirm: () => void;
+  onDragIntentStart: () => void;
+  onDragIntentEnd: () => void;
+  active: boolean;
+} & React.HTMLProps<HTMLDivElement>;
+
+class DragIntentContainer extends React.Component<Props> {
+  private dragTimer: number | null = null;
+
+  // Keeps track of dragEvents over the Collection's toggleVisibility button
+  // to determine enter/leave events
+  private dragHoverDepth: number = 0;
+
+  public handleToggleButtonDragEnter = () => {
+    if (this.dragHoverDepth === 0 && this.props.active) {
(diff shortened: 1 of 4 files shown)


client-v2/src/shared/components/__tests__/DragIntentContainer.spec.tsx
```diff
@@ -0,0 +1,118 @@
+import React from 'react';
+import { render, cleanup, fireEvent } from 'react-testing-library';
+import DragIntentContainer from '../DragIntentContainer';
+
+afterEach(cleanup);
+jest.useFakeTimers();
+
+describe('DragIntentContainer', () => {
+  it('should fire onIntentConfirm callback 300ms after Drag has entered', async () => {
+    const callback1 = jest.fn();
```
akash1810: minor: could this have a more descriptive name, just to make it obvious what's happening further down the test?

client-v2/src/shared/components/DragIntentContainer.tsx
```diff
@@ -0,0 +1,77 @@
...
+  };
+
+  public handleToggleButtonDrop = () => {
+    this.dragHoverDepth = 0;
+    this.deregisterDragIntent();
+  };
+
+  public registerDragIntent = () => {
+    this.dragTimer = window.setTimeout(() => {
+      if (this.props.active) {
```
akash1810: do we want to make this check earlier, so we only `setTimeout` when necessary?

PR #561, opened 2019-01-21
Catch errors from panda session

RichieAHB:
This solves a problem where timeouts in `panda-session` where not getting caught when making a request. Specifically (seemingly in dev) `panda-session`'s `reEstablishSession` may hit the `5000` timeout and then throw an error.

This moves that call intro the `try {} catch (e) {}` and adds some tests around this and other peculiarities around `pandaFetch`. Notable that it _throws_ non-200 responses!

### DIFF
--- a/client-v2/src/services/__tests__/pandaFetch.spec.ts
+++ b/client-v2/src/services/__tests__/pandaFetch.spec.ts
@@ -1,0 +1,45 @@
+import pandaFetchImp from '../pandaFetch';
+import fetchMock from 'fetch-mock';
+
+afterEach(fetchMock.restore);
+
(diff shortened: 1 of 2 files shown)


RichieAHB:
@akash1810 - @jonathonherbert and I had a chat about this. We decided that we'd actually programmed quite a few parts of the codebase to this interface (catch anything that's not a 200) so it would be quite a big change.

The other nice(?) thing about this is that we know that we have a valid domain object inside `promise.then(domainObject => ...)` so we don't have to check the response status every time. Not sure how nice this ultimately is but we came down on the side of leaving it as it is?

PR #562, opened 2019-01-22
Update colours and spacing for accessibility

RichieAHB:
# What does this change?
screenshot 2019-01-22 at 12 01 03

The "Launch" button is now orange 🍊 and all orange buttons now have black text for improved accessibility!

## Before 👎 
screenshot 2019-01-22 at 12 04 08

## After 👍 
screenshot 2019-01-22 at 12 04 25

Shout out to Chrome DevTools for the accessibility contrast ratio info 🎨 

Additionally, large buttons now have a slight letter spacing for improved legibility.

### DIFF
--- a/client-v2/src/shared/components/input/ButtonDefault.ts
+++ b/client-v2/src/shared/components/input/ButtonDefault.ts
@@ -54,7 +54,7 @@ const colorMap = {
   },
   deselected: {
     default: theme.colors.white,
-    primary: theme.colors.white,
+    primary: theme.colors.blackDark,
     muted: theme.colors.blackLight
   }
 };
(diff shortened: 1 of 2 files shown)


client-v2/src/shared/components/input/ButtonDefault.ts
```diff
@@ -54,7 +54,7 @@ const colorMap = {
   },
   deselected: {
     default: theme.colors.white,
-    primary: theme.colors.white,
+    primary: theme.colors.blackDark,
```
akash1810: This is nothing to do with this PR, but... I'm not really a fan of theme names referencing the colour they represent. That is, prefer `primary` to `red`, for example. This allows the theme to be changed more easily as we just change the value of `primary` without the fear of the name not representing the value.

RichieAHB: I sort of agree but also when you get down to having lots of different shades of white the names become harder and harder to represent. Maybe we could talk about that with respect to our current colour constants and whether there's a better way? It's definitely been discussed a bit already.

PR #563, opened 2019-01-22
Batch stories-visible information in the collections endpoint response

jonathonherbert:
This reduces the number of gets we make when we're asking for collections:

Previously -
- GET for n collectionIds returns collections
- (GET for collection story count returns stories visible) for each collection (possibly more, given the numbers below for the UK CODE front? Bit confused!)
- screen shot 2019-01-22 at 13 31 07
(NB: this includes a few visual assets to render the collections as well as the collection request)

Now -
- GET for n collectionIds returns collections + stories visible
- screen shot 2019-01-22 at 13 30 26

Tested on CODE. We'd really appreciate comments about the idiomatic-ness (idiomaticity?) and readability of the Scala!


app/controllers/FaciaToolController.scala
```diff
@@ -55,25 +56,6 @@ class FaciaToolController(
     }
   }
```
jonathonherbert: To make the endpoint usage across tool versions explicit, we've moved getCollections to the V2 controller.

app/controllers/StoriesVisibleController.scala
```diff
@@ -11,57 +12,11 @@ case class StoriesVisibleRequest(
...
-  implicit val jsonFormat = Json.format[StoriesVisibleResponse]
-}
-
-case class StoriesVisibleResponse(
-  desktop: Option[Int],
-  mobile: Option[Int]
-)
-
-class StoriesVisibleController(val containers: Containers, val deps: BaseFaciaControllerComponents) extends BaseFaciaController(deps) {
+class StoriesVisibleController(val containerService: ContainerService, val deps: BaseFaciaControllerComponents) extends BaseFaciaController(deps) {
```
jonathonherbert: The logic here now lives in a service, and the controller is nice and slim as a result.

client-v2/src/actions/Collections.ts
```diff
@@ -48,13 +48,34 @@ function getCollections(collectionIds: string[]): ThunkResult<Promise<void>> {
...
       await Promise.all(
-        collections.map(async collection => {
+        collectionResponses.map(async (collectionResponse, index) => {
+          const collectionId = collectionIds[index];
           const collectionConfig = getCollectionConfig(
             getState(),
-            collection.id
+            collectionId
           );
+          if (!collectionResponse) {
```
jonathonherbert: There are a few additional cases to manage here, as any result in an array of returned collections can be null.

test/services/CollectionServiceTest.scala
```diff
@@ -0,0 +1,74 @@
+package services
+
+import com.gu.facia.client.models.{CollectionConfigJson, CollectionJson, ConfigJson, FrontJson, Trail}
+import conf.ApplicationConfiguration
+import org.joda.time.DateTime
+import org.scalatest.{DoNotDiscover, FreeSpec, Matchers}
+import play.api.{Configuration}
+import slices.{Containers, FixedContainers, Story}
+
+@DoNotDiscover class CollectionServiceTest extends FreeSpec with Matchers {
```
akash1810: why the `DoNotDiscover` annotation?

test/services/CollectionServiceTest.scala
```diff
@@ -0,0 +1,74 @@
...
+    val draft = Trail("newId", 0, Some(""), None) :: live
+    CollectionJson(live, Some(draft), None, new DateTime(0), "oldUpdatedBy", "oldUpdatedEmail", None, None, None)
+  }
+
+  private def configJson: ConfigJson = {
+    ConfigJson(Map[String, FrontJson](), Map[String, CollectionConfigJson]("id" -> collectionConfig))
+  }
+
+  private def collectionConfig: CollectionConfigJson = {
+    CollectionConfigJson(
```
akash1810: do we do this often? that is, construct a largely empty `CollectionConfigJson`? If so, we could push its creation to the companion object to make things a bit DRYer.

test/services/CollectionServiceTest.scala
jonathonherbert: Because it was `ctrl-c`'ed from another test class! Whoops! We can remove it.

PR #565, opened 2019-01-22
Stole trash icon from workflow and made button larger

RichieAHB:
# What changed?
See title _Stole trash icon from workflow and made button larger_!

screenshot 2019-01-22 at 15 19 00

Stole the trash from here:
screenshot 2019-01-22 at 15 20 11

Currently the trash icon is showing when there is not an image 🤔 will add a card for this.

### DIFF
--- a/client-v2/src/shared/components/input/InputImage.tsx
+++ b/client-v2/src/shared/components/input/InputImage.tsx
@@ -2,7 +2,8 @@ import React from 'react';
 import { styled } from 'shared/constants/theme';
 import { connect } from 'react-redux';
 import { WrappedFieldProps } from 'redux-form';
-import deleteIcon from '../../images/icons/delete-copy.svg';
+import deleteIcon from '../../images/icons/trash.svg';
+import crossIcon from '../../images/icons/delete-copy.svg';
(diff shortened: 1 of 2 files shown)


jonathonherbert:
👍 I wonder whether it's worth making the click target square, either the button itself or a click target underneath it? It's quite a small button and users accidentally hitting the top right corner of the image by accident will be rewarded with a Grid modal 😢

RichieAHB:
@jonathonherbert I was actually going to bring this up tomorrow. I feel like that's probably what happened with Mariana's "view button opening the edit pane instead of the article". I wondered whether we could focus on this more broadly? Whadyathink?

jonathonherbert:
yep, sounds good!

RichieAHB:
Additionaly our hover style may be too subtle, which probably is a tiny bit more of an aid when you can see your hovering!

akash1810:
re hover area, I'd advocate for copying Composer's design for this by putting action buttons beneath the element

image

RichieAHB:
@akash1810 I don't necessarily disagree but I'm no designer! Happy to refer your suggestion to @akemitakagi 😁

PR #567, opened 2019-01-23
Extend button target area for toggling Collection open/close visibility

vlbee:
📣 Shout out to @laurenemms for her help with this PR. 

### What does this change?

Users now have a much larger target to hit the Collection toggle visibility button. We made the area the width of the article image, so that it did not overlap with any Warning buttons or other, non-clickable text in the meta data header. 

kapture 2019-01-23 at 14 07 31

### DIFF
--- a/client-v2/src/components/CollectionNotification.tsx
+++ b/client-v2/src/components/CollectionNotification.tsx
@@ -61,9 +61,12 @@ class CollectionNotification extends React.Component<
           &nbsp;
           <Button
             size="s"
-            onClick={() =>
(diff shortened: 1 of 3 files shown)


vlbee:
The height of the row with meta data varies - it's much taller when there are warnings, so the button might move around and look different if it's not top aligned.

RichieAHB:
As discussed IRL @vlbee - potentially the "show more" button shouldn't open the collection but not sure of the UX around this.

vlbee:
Fixed the button event propagation - now the Warning and Toggle button don't interfere with each other:
kapture 2019-01-24 at 17 40 55

PR #568, opened 2019-01-24
Add supporting articles to collection articles selectors

jonathonherbert:
Previously, selectors for articles in collection groups, and entire collections, ignored supporting articles. This meant that CAPI articles for supporting articles were not being picked up when Fronts loaded.

Do not merge -- this change may have consequences for other parts of the app that we haven't thought of yet!


RichieAHB:
Discussed IRL and this makes sense given what we discussed before you implemented this 👍 also nice work for fixing the other place where this was happening!

jonathonherbert:
I've altered the selector to allow callers to opt out of supporting articles -- the two places I can see this mattering are collection caps (which don't use this selector) and the overview item count (which does), and I've made alterations accordingly. I can't see any more usages, so this should cover all cases.

PR #570, opened 2019-01-25
exclude labs and hosted content on v1

akash1810:
https://github.com/guardian/facia-tool/pull/555 but on v1


Reettaphant:
Do we still want to get this merged @akash1810 ?

Reettaphant:
@akash1810 This was really stupid and my fault. The boolean logic was wrong but in two different places so in most cases we got the correct result. We got the wrong result in the edge case where an article didn't have any tags yet and the mocks in the tests didn't have them so caught the error.

PR #572, opened 2019-01-25
Scheduled display

Reettaphant:
This pr fixes the issue where draft capi feed was displaying the wrong results. This was because we were using the wrong query. This pr fixes the issue by using the correct query and displaying the scheduled launch date in the capi feed so the ordering of the article makes sense. I did a few other changes to this as well
- Change the way we display dates, the old way was too long as we now need more dates for draft content
- Only poll for live/draft content if that toggle has been selected
- Nicer colour for scheduled dates in collections

I know @jonathonherbert wanted tests but not sure which tests to add as the problem here was just an incorrect query to capi and bad date displays.

screen shot 2019-01-25 at 15 31 05
screen shot 2019-01-25 at 15 30 59

### DIFF
--- a/client-v2/src/bundles/__tests__/capiFeedBundle.spec.ts
+++ b/client-v2/src/bundles/__tests__/capiFeedBundle.spec.ts
@@ -10,11 +10,17 @@ import {
 import { capiArticle } from 'shared/fixtures/shared';
 
 const resources = [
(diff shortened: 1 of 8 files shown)


akemitakagi:
From UX/UI, looks good!

PR #577, opened 2019-01-31
Only add dirtied fields to the article meta on save.

jonathonherbert:
## What's changed?

Previously, the form would persist every available field when the user hit 'save'. Now, it just persists what's been dirtied. This makes it less likely that users will stomp on fields concurrently edited by other users, which paves the way for the upcoming collection polling and presence features.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/util/form.ts
+++ b/client-v2/src/util/form.ts
@@ -105,7 +105,8 @@ export const getInitialValuesForArticleFragmentForm = (
           src: article.imageCutoutSrc,
           width: strToInt(article.imageCutoutSrcWidth),
           height: strToInt(article.imageCutoutSrcHeight),
-          origin: article.imageCutoutSrcOrigin
+          origin: article.imageCutoutSrcOrigin,
+          thumb: article.imageCutoutSrc
         },
         slideshow: slideshow.concat(slideshowBackfill)
       }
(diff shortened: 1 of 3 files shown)


client-v2/src/util/form.ts
```diff
@@ -101,14 +105,30 @@ export const getInitialValuesForArticleFragmentForm = (
           src: article.imageCutoutSrc,
           width: strToInt(article.imageCutoutSrcWidth),
           height: strToInt(article.imageCutoutSrcHeight),
-          origin: article.imageCutoutSrcOrigin
+          origin: article.imageCutoutSrcOrigin,
+          thumb: article.imageCutoutSrc
```
jonathonherbert: @akash1810 This thumbnail is the original image -- as far as I can see from the modelling and old tool, there isn't a thumbnail available in the article meta and the old tool simply uses the entire image here, but I thought I'd run this by you!

client-v2/src/util/form.ts
```diff
@@ -101,14 +105,30 @@ export const getInitialValuesForArticleFragmentForm = (
...
         },
         slideshow: slideshow.concat(slideshowBackfill)
       }
     : undefined;
 };
 
+const formToMetaFieldMap: { [fieldName: string]: string } = {
```
jonathonherbert: This is nice and quick to map, but there's some duplication. We could model it the other way round `{[field: string]: Array<string>} `if anyone feels that the grossness factor is too high!

PR #582, opened 2019-02-04
Add styled-components theme to replace all hex colors

vlbee:
## What's changed?

This PR completes removal of all remaining hex codes in unshared components. The tool otherwise looks and performs the same. 

## Implementation notes

There are two 'theme' files: `shared/constants/theme` and `constants/theme` which incorporates the shared theme typings.  Please refer to these and expand on them as needed when adding colors going forward. 

Some `rgb` colors with alpha channels have not been changed. 

### General
- N/A 🤖 Relevant tests added 
- [x] 🔍 Checked on CODE

### Client

⚠️  I am getting the error below in local dev tools but it is not on Code. Any help understanding why would be appreciated as this doesn't seem related to any code I changed :) 
image

- N/A 📷 Screenshots / GIFs of relevant UI changes included

### DIFF
--- a/client-v2/src/components/App.tsx
+++ b/client-v2/src/components/App.tsx
@@ -3,7 +3,7 @@ import { injectGlobal, ThemeProvider } from 'styled-components';
 import { Switch, Route } from 'react-router-dom';
 import DropDisabler from './util/DropDisabler';
 
-import { theme, styled } from 'shared/constants/theme';
+import { theme, styled } from 'constants/theme';
 import { priorities } from 'constants/priorities';
 import SectionHeaderWithLogo from './layout/SectionHeaderWithLogo';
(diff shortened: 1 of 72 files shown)


vlbee:
@RichieAHB  I will fix the checkbox red ink in a separate PR.  

Regarding `withTheme` - I didn't try it because I did not know about it! 

I've since tried to use `withTheme` on components that require access, but decided against it at this stage. At the moment without `withTheme` we are importing the theme directly as a dependency. 
Ideally, using the `withTheme` HOC component should allow us to forgoe this hardcoded dependency. However, because we are using TS and a strongly-typed theme, there is no way for `withTheme` to know about the theme typings without importing it as a dependency anyways, so it seems like more complex code without clear benefits at the moment.

PR #583, opened 2019-02-04
Drop crop

RichieAHB:
## What's changed?

This allows dropping of grid crops on to Collection Items

## Implementation notes

I've extended `DragIntentContainer` to tell me when things are hovered over immediately as it works quite nicely and the logic is already baked in there. This may be a touch confusing (especially given the name of the component)!

I've not done any of the checks yet but thought I'd get this out for people to look at.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client
  - This interacts a lot with article dropping but I've tested that a lot and seems to work well

### DIFF
--- a/client-v2/src/components/FrontsEdit/Front.tsx
+++ b/client-v2/src/components/FrontsEdit/Front.tsx
@@ -7,7 +7,8 @@ import { State } from 'types/State';
 import { Dispatch } from 'types/Store';
 import {
   removeArticleFragment,
-  moveArticleFragment
+  moveArticleFragment,
+  updateArticleFragmentMeta
 } from 'actions/ArticleFragments';
 import { insertArticleFragmentFromDropEvent } from 'util/collectionUtils';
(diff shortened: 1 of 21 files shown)


RichieAHB:
@harpreetpurewal is this blocked until we can figure some method of undoing this behaviour - I don't remember 🤔

client-v2/src/components/FrontsEdit/Front.tsx
```diff
@@ -134,6 +136,16 @@ class FrontComponent extends React.Component<FrontProps, FrontState> {
     }
   }
 
+  public addImageToArticleFragment(
```
Reettaphant: Is this a duplicate of the one function which exists in the clipboard component? Could we add to some utilities file perhaps with the `getImageMetaFromValidationResponse` function?

RichieAHB: Done 🆗

PR #586, opened 2019-02-08
Remove redundant config usage

jonathonherbert:
## What's changed?

This is a minor refactor of the collection and stories getter logic, to remove some redundant code.

## Implementation notes

We now only need the collection config to determine the collection `type`, and this dependency could be removed if we passed that in some other way. This might matter because the collection config is _large_ -- around 200kb at the time of writing post-gzip -- and every time we read from the collections endpoint we're parsing this file for a tiny subset of the data it contains. It'd be nice to avoid that if possible.

### General
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/services/CollectionService.scala
+++ b/app/services/CollectionService.scala
@@ -50,7 +50,7 @@ class CollectionService(frontsApi: FrontsApi, containerService: ContainerService
 
 object CollectionService {
   def getStoriesVisibleByStage(collectionId: String, collection: CollectionJson, config: ConfigJson, containerService: ContainerService): Option[StoriesVisibleByStage] = {
-    val stages = CollectionService.getStoriesForCollectionStages(collectionId, collection, config)
+    val stages = CollectionService.getStoriesForCollectionStages(collectionId, collection)
     config.collections.get(collectionId).flatMap(_.`type`) match {
       case Some(cType) =>
         Some(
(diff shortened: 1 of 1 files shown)


app/services/CollectionService.scala
```diff
@@ -63,34 +63,28 @@ object CollectionService {
...
-          group <- meta.group
-        } yield {
-          group.toInt
-        }
-        val group = maybeGroup.getOrElse(0)
-        Story(
-          group,
-          article.meta.flatMap(_.isBoosted).getOrElse(false)
-        )
+  def getStoriesForStage(stage: List[Trail]): Seq[Story] = {
```
Reettaphant: Should the `stage` variable be called something like `storiesInStage`?

app/services/CollectionService.scala
```diff
@@ -63,34 +63,28 @@ object CollectionService {
...
-      case Some(collectionConfig) => collectionConfig.groups.map(_.length).getOrElse(0)
-      case None => 0
-    }
-
-    val liveStages = Some(collection.live).map { getStoriesForStage(_, numberOfGroups) }.getOrElse(Seq())
+  def getStoriesForCollectionStages(collectionId: String, collection: CollectionJson): (Seq[Story], Seq[Story]) = {
+    val liveStages = Some(collection.live).map(getStoriesForStage).getOrElse(Seq())
 
     (
       liveStages,
```
Reettaphant: we do we don't get the liveStages inline like we do draftStages?

app/services/CollectionService.scala
jonathonherbert: yes, the naming was bad and I didn't notice! Thanks 👍

app/services/CollectionService.scala
jonathonherbert: Yeah, maybe it's a bit gross -- I've refactored.

PR #589, opened 2019-02-13
Add Collapse All button to Collection Overview

vlbee:
## What's changed?

Users can now close all open collections at the click of a button form the Overview panel.
 Clicking it again with no open Collections does nothing. (ignore pink highlighting on headline - that is just highlighting captured by mistake)

kapture 2019-02-08 at 17 20 24

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/FrontCollectionsOverview.tsx
+++ b/client-v2/src/components/FrontsEdit/FrontCollectionsOverview.tsx
@@ -1,11 +1,16 @@
 import React from 'react';
+import { connect } from 'react-redux';
+import { styled } from 'constants/theme';
 import { State } from 'types/State';
+import { Dispatch } from 'types/Store';
 import { getFront } from 'selectors/frontsSelectors';
+import { closeCollections } from 'actions/Collections';
 import { FrontConfig } from 'types/FaciaApi';
 import CollectionOverview from './CollectionOverview';
-import { connect } from 'react-redux';
 import { CollectionItemSets } from 'shared/types/Collection';
-import { styled } from 'constants/theme';
+import ButtonCircularCaret, {
+  ButtonCircularWithTransition
+} from 'shared/components/input/ButtonCircularCaret';
 import ContainerHeadingPinline from 'shared/components/typography/ContainerHeadingPinline';
 import ContentContainer from 'shared/components/layout/ContentContainer';
 
@@ -16,7 +21,12 @@ interface FrontContainerProps {
 
 type FrontCollectionOverviewProps = FrontContainerProps & {
   front: FrontConfig;
+  closeAllCollections: (collections: string[]) => void;
 };
+interface CollectionOverviewCollapseAllButtonProps {
+  front: FrontConfig;
+  closeAllCollections: (collections: string[]) => void;
+}
 
 const Container = styled(ContentContainer)`
   display: flex;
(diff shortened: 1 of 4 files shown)


vlbee:
I was having weird CI build failures on #587  so killed the PR and made a new one. As far as I can tell, using `git commit --amend` and force pushing changes can sometimes cause Travis build errors.

client-v2/src/components/FrontsEdit/FrontCollectionsOverview.tsx
```diff
@@ -16,7 +21,12 @@ interface FrontContainerProps {
 
 type FrontCollectionOverviewProps = FrontContainerProps & {
   front: FrontConfig;
+  closeAllCollections: (collections: string[]) => void;
 };
+interface CollectionOverviewCollapseAllButtonProps {
+  front: FrontConfig;
+  closeAllCollections: (collections: string[]) => void;
+}
```
vlbee: Not sure what is the best DRY way to write this. 
`type FrontCollectionOverviewProps = FrontContainerProps & CollectionOverviewCollapseAllButtonProps` doesn't feel quite right.

client-v2/src/shared/components/input/ButtonCircularCaret.tsx
```diff
@@ -6,34 +6,45 @@ import ButtonCircular from './ButtonCircular';
 
 export const ButtonCircularWithTransition = ButtonCircular.extend<{
   highlight?: boolean;
+  small?: boolean;
 }>`
```
vlbee: I wanted to try to reuse the size types `'s' | 'm' | 'l'` from the very original ButtonDefault component but wasn't able to get it working. Open to suggestions.

client-v2/src/components/FrontsEdit/FrontCollectionsOverview.tsx
RichieAHB: This should be fine, I think I'd have defaulted to the other but now you mention it I can see why it would feel odd.

client-v2/src/components/FrontsEdit/FrontCollectionsOverview.tsx
```diff
@@ -26,13 +36,58 @@ const Container = styled(ContentContainer)`
...
+  front,
+  closeAllCollections
+}: CollectionOverviewCollapseAllButtonProps) => (
+  <CollapseAllDiv
+    onClick={e => {
+      e.preventDefault();
+      closeAllCollections(front.collections);
+    }}
+  >
+    <ButtonCircularCaret small={true} active={true} preActive={false} />{' '}
```
RichieAHB: Just a style note here you can omit `={true}` here for the true ones boolean props.

client-v2/src/components/FrontsEdit/FrontCollectionsOverview.tsx
```diff
@@ -26,13 +36,58 @@ const Container = styled(ContentContainer)`
...
+`;
+
+const CollapseAllLabel = styled('div')`
+  display: inline-block;
+`;
+
+const CollapseAllButton = ({
+  front,
+  closeAllCollections
+}: CollectionOverviewCollapseAllButtonProps) => (
```
RichieAHB: I'd be tempted here just to have the `CollapseAll` button be a bit more generic and just take an `onClick` handler (or even `React.HTMLProps<HTMLDivElement>` and just spread them on to the `CollapseAllDiv`).  The consumer can just pass the whole handler in and we still get all of the styles to reuse in another case.

client-v2/src/shared/components/input/ButtonCircularCaret.tsx
RichieAHB: The problem here is that `ButtonCircular` overrides `height`, `width` and `padding` that are the things modified by the `size` prop in `ButtonDefault`. You could modify the implementation of `ButtonCircular` (if you wanted) to set the _props_ for `ButtonDefault` to achieve the required styles rather than just overriding their styles. It's up to you though.

vlbee: I think I'm going to leave as is for now because the caret image also needs to be resized and would rather leave the logic in one component for now. I think there is probably a separate PR here re-looking at making this component more resusable alongside other use cases, like the hover buttons and the large circular button on the Fronts list overlay...

client-v2/src/components/FrontsEdit/FrontCollectionsOverview.tsx
vlbee: Good point. I will refactor.

RichieAHB: No need to refactor if you don't have time - it was more just an idea. Everything looks really good though!

PR #590, opened 2019-02-14
Rahb/close panels

RichieAHB:
## What's changed?

This allows collapsing of both the clipboard or an overview through either toggle buttons or key commands.

kapture 2019-02-14 at 17 15 28

## Implementation notes

Currently they key commands are arbitrarily picked and have no discoverability. I feel like we should address this in a following PR in conversation with @akemitakagi .

As @akash1810 reminded me, we have a general way of highlighting keyboard shortcuts in tools: `shift-?`, which brings up a modal showing them. Maybe we can do something like this?

I've implemented some defaults as you can see in the code:

`command-j` - close all overviews
`command-k` - open all overviews
`command-u` - toggle clipboard

I have no idea whether these are good defaults but they are up for debate and they aren't used by macOS / Chrome!

Also as @paperboyo notes, there may be a bit of wasted space here when things are closed which, if necessary we could address but perhaps this would be good to test before we worry too much.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


RichieAHB:
Also I've used this for key mapping https://github.com/ccampbell/mousetrap

akemitakagi:
Looks great. For the space when they are closed, we will review this once the user started to use this function. The space is useful/necessary for 'Clipboard' as the user can still drag an item into the closed clipboard.

paperboyo:
> The space is useful/necessary for 'Clipboard' as the user can still drag an item into the closed clipboard.

Exactly! So let’s show clipped content there in collapsed state. Collapsed state of an overview could also have, maybe, first letters of container names and also be functional?

RichieAHB:
Thanks for nudging me @akemitakagi - I've just made it so that when you hover over the clipboard when it's closed it will open it 😁 🎉

RichieAHB:
As noted by @jonathonherbert the fronts weren't collapsing properly due to the header. As per @akemitakagi 's actual design that I missed, the _last refreshed_ time is now under the header:
 
screenshot 2019-02-18 at 10 12 34

client-v2/src/keyboard-shortcuts/init.ts
```diff
@@ -0,0 +1,29 @@
...
+  editorCloseAllOverviews,
+  editorOpenAllOverviews,
+  selectIsClipboardOpen,
+  editorCloseClipboard,
+  editorOpenClipboard
+} from 'bundles/frontsUIBundle';
+import { trap } from './trap';
+
+const createTrapMap = (store: Store) => ({
+  'command+j': () => {
```
akash1810: how does this map on non macs?

RichieAHB:
@akash1810 generic at which level? You mean a generic UI for panels?

client-v2/src/keyboard-shortcuts/init.ts
RichieAHB: Good point. It doesn't, although this key binding is _secret_. It's there to test the implementation, I'm not even sure if that's a good binding even on mac. I think `command` is synonymous with `meta` which I _think_ is the windows key on windows for example.

PR #596, opened 2019-02-22
Collection polling

RichieAHB:
## What's changed?

This change implements polling for collections! Our application will poll all collections in the loaded fronts foreeeevvveeer. When the application hits `/collections` with a `lastUpdated` in a `CollectionSpec` then that collection will only be returned _if_ it has been updated since that time.

This _could_ be done on the server but we made the decision to unburdern the client of this, although it has led to a somewhat unusual `/collections` endpoint (noted below).

**Note:** this has borked our integration tests, which will need changing to this new `/collections` endpoint before merging.

### Get collections endpoint

This endpoint is a bit odd (and un-semantic) in that it is a POST endpoint for something that should really be a GET. This is because GETs don't traditionally support bodies - although it seems there's nothing in the spec the rules this out and I _think_ Play supports it. It still feels a bit weird without putting it in the query params though, but doing this would be in either JSON or perhaps some weird usage of query param syntax ... we may also decide that it simply doesn't matter 🤷‍♂️ 

### Race conditions

Aside from the obvious race conditions between two clients that happen with polling, Vanessa and had a think about whether _one_ client could ever poll and update at the same time and stomp on it's own update. We don't think it can but it'd be good to have a few other people thinking about that.

## Tests

We've tested a good amount of this logic, both on the server and client but it'd be good if people could read the specs and see whether they think we're missing anything.

### `type`

When @jonathonherbert and @vlbee started this branch a `type` key was added to what is now a `CollectionSpec`. One potentially large oversight here is that _we don't use it anywhere_ and Vanessa can't remember why it was there. So @jonathonherbert if you can let us know what it was meant for before we merge this, just in case we've missed anything major 😨 

Obviously Vanessa and I are happy to talk IRL about this PR too 👍 

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/actions/__tests__/Collections.spec.ts
+++ b/client-v2/src/actions/__tests__/Collections.spec.ts
@@ -66,7 +66,7 @@ describe('Collection actions', () => {
         }
       });
       expect(actions[1]).toEqual(
-        collectionActions.updateSuccess('exampleCollection', collection)
+        collectionActions.updateSuccess('exampleCollection')
       );
     });
   });
@@ -80,11 +80,7 @@ describe('Collection actions', () => {
 
     it('should add fetched Collections to the store', async () => {
       const collectionIds = ['testCollection1', 'testCollection2'];
-      fetchMock.post('/collections', getCollectionsThunkFaciaApiResponse, {
-        headers: {
-          'Content-Type': 'application/json'
-        }
-      });
+      fetchMock.post('/collections', getCollectionsThunkFaciaApiResponse);
       await store.dispatch(getCollections(collectionIds) as any);
       expect(store.getState().shared.collections.data).toEqual({
         exampleCollection: {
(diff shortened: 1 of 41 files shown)


jonathonherbert:
Re: `type` -- I think this was a _potential_ addition that could be used to remove our dependency on the collection config every time we fetch collections. Sorry, that might not have been clear from the whiteboarding exercises! If that's the case, we shouldn't need it at the moment 👍

client-v2/src/actions/__tests__/Collections.spec.ts
```diff
@@ -62,11 +66,95 @@ describe('Collection actions', () => {
...
+      );
+      await store.dispatch(getCollections(collectionIds, true) as any);
+      const result = request.lastOptions().body;
+      expect(JSON.parse(result as string)).toEqual([
+        { id: 'testCollection1', type: 'type', lastUpdated: 1547479667115 },
+        { id: 'testCollection2', type: 'type', lastUpdated: 1547479667115 }
+      ]);
+    });
+  });
+
```
jonathonherbert: Testing testing testing day and night 👍

jonathonherbert:
Me and @RichieAHB have been over this and it looks 👍, with one suggestion -- leaving the user's form intact when articles change underneath them might be less disruptive to them.

PR #598, opened 2019-02-25
Avoid rerendering the collection tree when selecting articles for edit

jonathonherbert:
## What's changed?

At the moment, we rerender the entire collection tree when collection items are selected for edit. This is expensive, and you can feel the jank when the form opens on large fronts. This PR should fix the issue -- we now only render the articles that are changing state.

## Implementation notes

This was a nice candidate for a refactor, as it DRYed up code across the Fronts and Clipboard components. I've also const-ified the relevant action names where necessary to allow us to depend on them outside of their reducer context, which is probably a good thing in any case!

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/Clipboard.tsx
+++ b/client-v2/src/components/Clipboard.tsx
@@ -10,7 +10,6 @@ import {
 } from 'actions/ArticleFragments';
 import {
   editorSelectArticleFragment,
-  selectEditorArticleFragment,
   editorClearArticleFragmentSelection,
   selectIsClipboardOpen,
   editorOpenClipboard,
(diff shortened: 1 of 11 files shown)


client-v2/src/components/Clipboard.tsx
```diff
@@ -80,25 +78,6 @@ class Clipboard extends React.Component<ClipboardProps> {
     this.props.dispatch(insertArticleFragmentFromDropEvent(e, to, 'clipboard'));
   };
 
-  public removeCollectionItem = (id: string) => {
```
jonathonherbert: This is nice to remove -- we repeat ourselves in the Front and Clipboard components, and in any case this logic is probably better off living in a reducer, as it's business-y.

client-v2/src/selectors/frontsSelectors.ts
```diff
@@ -267,11 +267,13 @@ const visibleArticlesSelector = createSelector(
...
 );
 
+const defaultVisibleFrontArticles = {};
+
 const visibleFrontArticlesSelector = createSelector(
   [collectionVisibilitiesSelector, collectionSetSelector],
   (collectionVisibilities, collectionSet) => {
     if (collectionSet === 'previously') {
-      return {};
+      return defaultVisibleFrontArticles;
```
jonathonherbert: This will prevent rerenders, as an object literal will have a different identity each time we return it.

PR #600, opened 2019-02-27
Avoid rerendering the collection tree when selecting articles for edit (take 2)

jonathonherbert:
## What's changed?
At the moment, we rerender the entire collection tree when collection items are selected for edit. This is expensive, and you can feel the jank when the form opens on large fronts. This PR should fix the issue -- we now only render the articles that are changing state.

_NB:_ This PR corrects an issue with form saving present in the original PR for this feature.

## Implementation notes
This was a nice candidate for a refactor, as it DRYed up code across the Fronts and Clipboard components. I've also const-ified the relevant action names where necessary to allow us to depend on them outside of their reducer context, which is probably a good thing in any case!

### General
* [x]  🤖 Relevant tests added
* [x]  ✅ CI checks / tests run locally
* [x]  🔍 Checked on CODE

### Client
* [x]  🚫 No obvious console errors on the client (i.e. React dev mode errors)
* [x]  🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
* [ ]  📷 Screenshots / GIFs of relevant UI changes included (N/A)

### DIFF
--- a/client-v2/src/components/FrontsEdit/FrontDetailView.tsx
+++ b/client-v2/src/components/FrontsEdit/FrontDetailView.tsx
@@ -1,0 +1,65 @@
+import React from 'react';
+import ArticleFragmentForm from './ArticleFragmentForm';
+import FrontCollectionsOverview from './FrontCollectionsOverview';
+import { connect } from 'react-redux';
+import {
+  ArticleFragmentMeta,
+  CollectionItemSets
+} from 'shared/types/Collection';
+import { updateArticleFragmentMeta } from 'actions/ArticleFragments';
+import {
+  editorClearArticleFragmentSelection,
+  selectEditorArticleFragment
+} from 'bundles/frontsUIBundle';
+import { Dispatch } from 'types/Store';
+import { State } from 'types/State';
+
+interface ContainerProps {
+  id: string;
+  browsingStage: CollectionItemSets;
+}
+
+interface ComponentProps extends ContainerProps {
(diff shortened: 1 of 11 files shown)


client-v2/src/components/FrontsEdit/FrontDetailView.tsx
```diff
@@ -0,0 +1,66 @@
...
+  updateArticleFragmentMeta: (id: string, meta: ArticleFragmentMeta) => void;
+  clearArticleFragmentSelection: (id: string) => void;
+}
+
+const FrontsDetailView = ({
+  selectedArticleFragment,
+  id,
+  browsingStage,
+  clearArticleFragmentSelection,
+  updateArticleFragmentMeta
```
jonathonherbert: I'd made a mistake here in the previous PR -- I hadn't included updateArticleFragmentMeta as a destructured prop, and as a result the component was calling the action creator, which has the same name. This meant that nothing was dispatched when onSave was called 😢

RichieAHB: Oh wow that's fiddly ...

RichieAHB: Is the signature exactly the same too?

jonathonherbert: yuh-huh 👎 Potential way of spotting these things in future -- adopt naming convention for actions, e.g. actionUpdateArticleFragmentMeta

PR #601, opened 2019-03-04
Add escaping to config

RichieAHB:
## What's changed?

Before this change `</script>` tags left in anything that made it to the `config` would close the config script tag and allow arbitrary scripts to be added to the body on page load.

This PR leaves the config JSON escaped as we print it to the DOM to stop this from happening.

## Implementation notes

When reading from an DOM attribute the DOM parser will unescape the escaped JSON and gives us our config JSON with the HTML tags in it as expected. We solve any XSS possibility in the app using #549.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/views/V2App/app.scala.html
+++ b/app/views/V2App/app.scala.html
@@ -40,11 +40,7 @@ String)
 
 <body>
     <div class="wrap" id="react-mount"></div>
-
-    <script type="application/json" id="config">
-            @Html(clientConfigJson)
-        </script>
-
+    <div id="config" data-value="@clientConfigJson"></div>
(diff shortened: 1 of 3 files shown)


RichieAHB:
@akash1810 if you write one in in the clipboard then - voila - you have a script tag in config 👌

akash1810:
> @akash1810 if you write one in in the clipboard then - voila - you have a script tag in config 👌

Ah! Is this desired behaviour? (feels a little dangerous to me). Perhaps we should ensure we're writing the correct data to config too?

PR #603, opened 2019-03-04
Fix dedupe when flattening a drops supporting articles to sibling in the targets supporting articles

RichieAHB:
## What's changed?

Supporting articles get flattened and we end up inserting multiple article fragments on one level occasionally as a result.

E.g. when dragging an article with supporting articles into a supporting position, the dragged parent article is added as a supporting article of the target article (as is normal) but also the supporting articles from the _dragged_ article are also added as supporting articles, following after their parent. (This may need reading a few times to get!). I've called this "flattening".

Prior to this PR, if any of those supporting articles (the ones that had been to be adjacent to their sibling) were duplicates in their new position, both them and their duplicate counterparts that existed as a supporting article in the target _prior_ to the drop would be removed, instead of keeping one of them (specifically the newer ones)

This changes that and adds a couple of tests around here.

## Implementation notes

N/A

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/util/insertAndDedupeSiblings.ts
+++ b/client-v2/src/shared/util/insertAndDedupeSiblings.ts
@@ -22,7 +22,7 @@ export const insertAndDedupeSiblings = (
   );
 
   const isAnInsertedItem = (i: number) =>
-    i >= index && i < (index + insertionUUIDs.length);
+    i >= index && i < index + insertionUUIDs.length;
 
   // the filter alone should be enough here but just in case any of the
   // insertions were duplicates then run `uniqBy` over and dedupe again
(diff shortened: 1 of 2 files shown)


akash1810:
You'll need to rebase master to get CI in teamcity to pass.

client-v2/src/shared/util/insertAndDedupeSiblings.ts
```diff
@@ -29,7 +32,7 @@ export const insertAndDedupeSiblings = (
         // keep anything that doesn't match on id or is the item we just
         // inserted
         !insertionIDs.includes(siblingArticleFragment.id) ||
-        (isInsertionGroup && i === index)
+        (isInsertionGroup && isAnInsertedItem(i))
```
RichieAHB: Previously we were only checking for one drop article to ignore when deduping - the article that was dropped. However, when articles are "flattened" we can actually end up having more than one article being added in this position. So instead of checking `index` to see if we were the dropped _article_, we check index plus the amount of articles we've actually inserted to see if we were one of the dropped _articles_.

jonathonherbert: Thanks for the thorough explanation!

PR #604, opened 2019-03-05
Add fork-ts plugin

jonathonherbert:
## What's changed?

At the moment, our build step during development takes a while - between 5 and 7 seconds. To speed things up, this PR adds https://github.com/Realytics/fork-ts-checker-webpack-plugin and disables typechecking during transpilation. Typechecking happens via this plugin in a separate thread.

This speeds both parts of the process up -- our builds now run in under a second, and we get typechecking results slightly after. An error during typechecking still fails a build, so we're still typesafe  in CI 👍 

Before (build time):
screen shot 2019-03-05 at 09 19 43

After (build time, then typechecking time):
screen shot 2019-03-05 at 08 59 56

### General
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/config/webpack.config.common.js
+++ b/client-v2/config/webpack.config.common.js
@@ -3,7 +3,7 @@ const TSConfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
 // https://github.com/Igorbek/typescript-plugin-styled-components
 const createStyledComponentsTransformer = require('typescript-plugin-styled-components').default;
 const styledComponentsTransformer = createStyledComponentsTransformer();
-
+var ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
 
 module.exports = {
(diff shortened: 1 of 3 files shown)


client-v2/config/webpack.config.common.js
```diff
@@ -3,14 +3,15 @@ const TSConfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
 // https://github.com/Igorbek/typescript-plugin-styled-components
 const createStyledComponentsTransformer = require('typescript-plugin-styled-components').default;
 const styledComponentsTransformer = createStyledComponentsTransformer();
-
+var ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
```
akash1810: `const`?

jonathonherbert: Oh, dumb, yes const 👍

PR #605, opened 2019-03-05
Move and adjust UI of Collapse All button

vlbee:
## What's changed?

Per the new UI/UX, the position and look & feel of the Collapse All but has moved.

image
image

## Implementation notes

Adjusted some height attributes in the Clipboard and Collection overview as well which were causing those components to flex-shrink. 

Not sure the best way to change fill colors in an image imported SVG. I've used a filter hack, but if anyone has a better idea of how to adjust fill color let me know. Other options are:
1. duplicate SVG file with a dark fill
2. inline the SVG and adjust with CSS to change the fill color as used

What do people prefer?

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


akash1810:
> Not sure the best way to change fill colors in an image imported SVG. I've used a filter hack, but if anyone has a better idea of how to adjust fill color let me know. Other options are:
> 1. duplicate SVG file with a dark fill
> 2. inline the SVG and adjust with CSS to change the fill color as used

Prefer the second option, but don't think we'd need to inline. We should be able to adjust styles with classes - https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_and_CSS.

Reettaphant:
@akash1810 @vlbee I did this in a recent PR to change colours on svgs as per @jonathonherbert's advice: client-v2/src/shared/components/collectionItem/CollectionItemQuote.tsx

vlbee:
Waiting on #612 to make this change and merge.

client-v2/src/actions/__tests__/ArticleFragments.spec.ts
```diff
@@ -296,7 +296,7 @@ describe('ArticleFragments actions', () => {
   });
 
   describe('insert image', () => {
-    fit('adds the correct image data', () => {
+    it('adds the correct image data', () => {
```
RichieAHB: 🙄 sorry

PR #606, opened 2019-03-06
Edited metadata

Reettaphant:
## What's changed?

Show how metadata has been edited in articles inside the collection. 
See underneath an example of: 
- an article with a breaking breaking news toggle selected
- an article with show byline and quote headline selected
- an article with hide image set to true
- an article with a slideshow 

screen shot 2019-03-06 at 13 44 17

## Implementation notes

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/components/article/ArticleBodyDefault.tsx
+++ b/client-v2/src/shared/components/article/ArticleBodyDefault.tsx
@@ -56,6 +56,17 @@ const ArticleHeadingContainerSmall = styled('div')`
   text-overflow: ellipsis;
 `;
 
+const ArticleBodyByline = styled('div')`
+  font-style: italic;
+  padding-top: 5px;
+`;
+
+const ArticleBodyQuote = styled('span')`
+  font-family: GHGuardianHeadline;
+  font-size: 20px;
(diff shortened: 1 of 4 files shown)


client-v2/src/shared/components/article/ArticleBodyDefault.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { styled } from 'shared/constants/theme';
 import startCase from 'lodash/startCase';
 import distanceInWordsStrict from 'date-fns/distance_in_words_strict';
+import { MdCollections } from 'react-icons/md';
```
vlbee: I believe a number of the icons we are using are from Material Design. Is is worth importing them all via react-icons like this rather than hosting our own svg files?  The library allows you to easily change color via context which is something I had to hack in a current PR #605 and it did not feel great.

Reettaphant: Yes, we should be consistent with how we import svgs. If everyone is happy with using this library I will make another changing this pr once this one is merged.

PR #609, opened 2019-03-08
Feed trail image

Reettaphant:
## What's changed?

Displays article trail pics in capi feed, reduces the clipboard size to make way for a slightly wider feed and adjusts some padding for the same purpose as well as per @akemitakagi's instructions.

## Implementation notes
Screenshot 2019-03-11 at 15 19 16

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsCAPIInterface/Feed.tsx
+++ b/client-v2/src/components/FrontsCAPIInterface/Feed.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { styled } from 'constants/theme';
 import FeedItem from './FeedItem';
 import { CapiArticle } from 'types/Capi';
+import { getThumbnail } from 'util/CAPIUtils';
 
 const getId = (internalPageCode: string | number | undefined) =>
   `internal-code/page/${internalPageCode}`;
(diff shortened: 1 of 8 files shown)


client-v2/src/components/FrontsCAPIInterface/Feed.tsx
```diff
@@ -28,31 +29,24 @@ const Feed = ({ articles = [], error }: FeedProps) => (
...
+            publicationDate={article.webPublicationDate}
+            sectionName={article.sectionName}
+            pillarId={article.pillarId}
+            internalPageCode={
+              article.fields && getId(article.fields.internalPageCode)
+            }
+            firstPublicationDate={article.fields.firstPublicationDate}
+            isLive={!article.fields.isLive || article.fields.isLive === 'true'}
+            scheduledPublicationDate={article.fields.scheduledPublicationDate}
+            thumbnail={getThumbnail(article, article.frontsMeta.defaults)}
```
Reettaphant: Is there a way to restructure an object and still have access to it in the fuction? I couldn't figure out the syntax if it exists.

client-v2/src/shared/components/Thumbnail.tsx
```diff
@@ -1,9 +1,20 @@
...
+const ThumbnailBase = styled('div')`
   background-size: cover;
   background-color: ${({ theme }) =>
     theme.shared.base.colors.backgroundColorFocused};
 `;
+
+const ThumbnailSmall = styled(ThumbnailBase)`
+  width: 83px;
+  height: 50px;
+  margin: 3px 5px 3px 5px;
```
Reettaphant: We should probably make an image component which can have different widths but preserves correct aspect rations and handles loading states consistently but that is perhaps for an another pr.

jonathonherbert: 👍 great idea.

client-v2/src/components/FrontsCAPIInterface/Feed.tsx
jonathonherbert: not in the function params, I think -- you'd have to pass it in whole and destructure within the function body.

jonathonherbert:
@Reettaphant looks good, it'll be nice to see the trail pics in there!

PR #610, opened 2019-03-08
(Re)add discard button

jonathonherbert:
## What's changed?

The discard button is back. That's pretty good.

form

We get a small grey notification icon in the overview when unsaved changes are present in a collection. This is pretty low-key -- would it be worth thinking of ways to make the messaging clearer here, @harpreetpurewal @akemitakagi ?

### General
- [x] 🔍 Checked on CODE

### Client


akemitakagi:
Yes, the message needs to be clear. I wonder if we could do something similar with the composer 'Publish' warning button. Show only one orange ! icon in the overview, when you hover over it you will see a note says " Saved but unlaunched changes" or "Unsaved change" etc.

jonathonherbert:
@akemitakagi that makes sense! I've just updated it to a single notification -- the hover text is a combination of whatever validation text applies. It'd be nice to have a larger tooltip, but perhaps this is a start?

jonathonherbert:
Closed as this was -- _the wrong discard button!_ Users were referring to discarding unlaunched changes, not unsaved form changes.

PR #612, opened 2019-03-10
Convert icons images to inline SVG components

vlbee:
## What's changed?

No changes to the UX/UI.  SVG icons are now inlined as React components rather than imported in image tags.

## Implementation notes

SVG icons were previously being imported as source files via html image tags. This meant it was not possible to  access and manipulate SVG attributes such as `fill` either directly or via CSS. To facilitate resizing and changing the color of SVG icons in the shared component library so they are more re-usable, all icons have now been converted to React components whose attributes can be changed via `props`
```
interface IconProps {
  fill?: string; // fill color
  size?: number | string; // numbers are converted to pixels automatically. For other units, use strings
  title?: string | null; // SVG titles function the same as alt text for images
}
```
- ColoredQuote Icon as been kept seperate as it doesn't operate semantically like the other icons. 
- Hover Icons on Article Fragments have been kept in their own module as they will always appear and function together as a group, whereas the other Icons I feel need to be more flexible in usage. 

Needs to be merged before #605 

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/shared/components/input/InputImage.tsx
```diff
@@ -52,9 +52,9 @@ const IconDelete = styled('img')`
   left: 9px;
 `;
 
-const IconAdd = styled(IconDelete)`
-  transform: rotate(45deg);
-`;
+// const IconAdd = styled(IconDelete)`
```
akash1810: do we need these lines?

client-v2/src/components/FrontsCAPIInterface/FilterItem.tsx
```diff
@@ -25,7 +25,7 @@ const FilterItem = ({ children, onClear }: FilterItemProps) => (
   <SearchTermItem>
     {children}
     <SmallRoundButton onClick={() => onClear()} title="Clear search">
-      <ClearButtonIcon src={moreImage} alt="" height="22px" width="22px" />
+      <ClearIcon size={22} />
```
akash1810: minor: I wonder if `size` could be `small`, `normal`, `big` to reduce the amount of possible variance?

vlbee: There are 6 different Icon sizes in use but I can do something like the map below? I think this is because the icons come from different sources they scale differently. 
```
const SizeMap = {
  xs: 10,
  s: 12,
  m: 14,
  l: 22,
  xl: 30,
  xxl: 40
};
```

client-v2/src/shared/components/input/InputImage.tsx
vlbee: Removed.

vlbee:
> Great work!
> 
> The directory structure of `icons` is a little confusing with some icons having their own file and others being grouped together in a single one. Could we follow one or the other?

I thought about this but specifically decided to split them into Icons and Hover Icons - 
The general icons are used individually in different places and can mean different things. This means they will usually be imported one at a time and it is easiest to import them by being able to type a named export. 
The Hover Icons on the other hand, will always appear together in a group when hovering over Article Fragments. They need to retain the same shape, size, etc for consistency, but I think this can be clear with appropriate naming. 

I've now made:
- All Hover Icons are their own named export but in a separate file.
- A size map has been added to the regular icons so that there are _only_ 7 options, like 👕

PR #618, opened 2019-03-13
Rearrange fronts

jonathonherbert:
## What's changed?

Adds the ability to rearrange and scroll to fronts via the fronts icon in the top left corner. You can also click on fronts to scroll directly to them.

frnotlist

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


client-v2/package.json
```diff
@@ -113,6 +116,7 @@
     "react-router": "^4.2.0",
     "react-router-dom": "^4.2.2",
     "react-router-redux": "^4.0.8",
+    "react-transition-group": "^2.6.0",
```
jonathonherbert: Can I have it? Can I have it? Pleease? (I'm surprised we don't have it already -- v. handy and also tiny! V. useful when you need to keep components around for animations as they unmount)

client-v2/src/bundles/frontsUIBundle.ts
```diff
@@ -26,21 +30,26 @@ import {
...
 
-const EDITOR_OPEN_FRONT = 'EDITOR_OPEN_FRONT';
-const EDITOR_CLOSE_FRONT = 'EDITOR_CLOSE_FRONT';
-const EDITOR_CLEAR_OPEN_FRONTS = 'EDITOR_CLEAR_OPEN_FRONTS';
-const EDITOR_SET_OPEN_FRONTS = 'EDITOR_SET_OPEN_FRONTS';
-const EDITOR_OPEN_COLLECTION = 'EDITOR_OPEN_COLLECTION';
-const EDITOR_CLOSE_COLLECTION = 'EDITOR_CLOSE_COLLECTION';
-const EDITOR_SELECT_ARTICLE_FRAGMENT = 'EDITOR_SELECT_ARTICLE_FRAGMENT';
-const EDITOR_CLEAR_ARTICLE_FRAGMENT_SELECTION =
+export const EDITOR_OPEN_CURRENT_FRONTS_MENU =
```
jonathonherbert: Constants are better here, they make the dependency explicit

client-v2/src/bundles/frontsUIBundle.ts
```diff
@@ -213,12 +256,43 @@ const clearArticleFragmentSelection = (state: State, frontId: string) => ({
 
 const reducer = (state: State = defaultState, action: Action): State => {
   switch (action.type) {
+    case EDITOR_OPEN_CURRENT_FRONTS_MENU: {
```
jonathonherbert: We had to lift this state out of React, as the fronts components need to know about it in order to get out of the way of our lovely menu

client-v2/package.json
```diff
@@ -105,6 +107,7 @@
     "raven": "^2.5.0",
     "raven-js": "^3.24.1",
     "react": "^16.3.0",
+    "react-beautiful-dnd": "^10.1.0",
```
jonathonherbert: This takes care of our drag/drop behaviour. Dare I say it -- it just works?

client-v2/src/components/transitions/FadeTransition.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import { CSSTransition } from 'react-transition-group';
+import { styled } from 'constants/theme';
+
+type TransitionDirections = 'up' | 'right' | 'down' | 'left';
+
```
jonathonherbert: A handy container for fade transitions. Hopefully it's useful to others outside of the context of this PR.

client-v2/src/shared/components/icons/Icons.tsx
```diff
@@ -7,17 +7,18 @@ interface IconProps {
...
 const IconSizeMap = {
   xxs: 10,
   xs: 12,
   s: 14,
   m: 18,
   l: 22,
   xl: 30,
-  xxl: 40
+  xxl: 40,
+  fill: '100%'
```
jonathonherbert: Sometimes you just want your icon to fill the parent!

client-v2/package.json
RichieAHB: I just had a flash back to `react-addons` - yes we can have this back!

client-v2/src/bundles/frontsUIBundle.ts
```diff
@@ -1,7 +1,11 @@
 import without from 'lodash/without';
+import compact from 'lodash/compact';
 import {
   Action,
+  EditorShowOpenFrontsMenu as EditorOpenCurrentFrontsMenu,
```
akash1810: any reason to not export this as `EditorOpenCurrentFrontsMenu` rather than alias on import?

client-v2/src/bundles/frontsUIBundle.ts
```diff
@@ -213,12 +256,43 @@ const clearArticleFragmentSelection = (state: State, frontId: string) => ({
...
     case EDITOR_OPEN_FRONT: {
       return {
         ...state,
         frontIds: state.frontIds.concat(action.payload.frontId)
       };
     }
+    case EDITOR_MOVE_FRONT: {
+      const maxIndex = state.frontIds.length - 1;
+      if (
+        state.frontIds.indexOf(action.payload.frontId) === -1 ||
```
akash1810: minor: would this be more readable if we extract them to local variables?

client-v2/src/bundles/frontsUIBundle.ts
```diff
@@ -213,12 +256,43 @@ const clearArticleFragmentSelection = (state: State, frontId: string) => ({
...
+        action.payload.fromIndex > maxIndex ||
```
Reettaphant: Not sure when this would be the case and why we are checking for it?

client-v2/src/components/FeedSectionHeader.tsx
```diff
@@ -0,0 +1,165 @@
...
+  margin-left: auto;
+  align-self: center;
+  padding-right: 10px;
+  line-height: 60px;
+  height: 60px;
+`.withComponent('a');
+
+const SectionHeaderContent = styled('div')`
+  position: relative;
+  // overflow: hidden;
```
akash1810: do we need this comment?

client-v2/src/bundles/frontsUIBundle.ts
Reettaphant: Ok, I understand. I was just worried I had missed something!

client-v2/src/bundles/frontsUIBundle.ts
jonathonherbert: ah, a hangover from a VSCode F2 refactor, thanks for spotting 👍

client-v2/src/components/FeedSectionHeader.tsx
jonathonherbert: .. no! 👀

client-v2/src/bundles/frontsUIBundle.ts
jonathonherbert: good point, I've just done this 👍

PR #620, opened 2019-03-13
Expand clipboard drop zone to fill area

jonathonherbert:
## What's changed?

Restores the clipboard target area, which has shrunk as a result of an element above the flex area not having height: 100%.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


jonathonherbert:
> Am sure we (in fact you) already fixed this once?

Yes -- the height disappeared from StyledDragIntentContainer again 😢

PR #625, opened 2019-03-15
Provide path for migration of user data

jonathonherbert:
## What's changed?

This PR exposes a temporary /userdata/migrate route to allow us to trigger the migration of user data from one format to another. It'll be necessary to migrate our list of frontIds to a new format that keys them by priority.

It also readies the endpoint that writes this data to accept both formats for the new client.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/UserDataController.scala
+++ b/app/controllers/UserDataController.scala
@@ -31,16 +31,21 @@ class UserDataController(frontsApi: FrontsApi, dynamo: Dynamo, val deps: BaseFac
   }
 
   def putFrontIds() = APIAuthAction { request =>
-    // We handle both cases here, because of the migration from a list to a map by priority
     val maybeFrontIds: Option[List[String]] = request.body.asJson.flatMap(
       _.asOpt[List[String]])
-    val maybeFrontIdsByPriority: Option[Map[String, List[String]]] = request.body.asJson.flatMap(
-      _.asOpt[Map[String, List[String]]])
-    (maybeFrontIds, maybeFrontIdsByPriority) match {
-      case (Some(frontIds), _) =>
+    maybeFrontIds match {
+      case Some(frontIds) =>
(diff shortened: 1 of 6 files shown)


app/controllers/UserDataController.scala
```diff
@@ -28,18 +31,48 @@ class UserDataController(dynamo: Dynamo, val deps: BaseFaciaControllerComponents
...
-      jsValue.asOpt[List[String]])
-
-    frontIds match {
-      case Some(ids) => {
-        val userEmail = request.user.email
-        Scanamo.exec(dynamo.client)(userDataTable.update('email -> request.user.email, set('frontIds -> ids)))
+    // We handle both cases here, because of the migration from a list to a map by priority
+    val maybeFrontIds: Option[List[String]] = request.body.asJson.flatMap(
+      _.asOpt[List[String]])
+    val maybeFrontIdsByPriority: Option[Map[String, List[String]]] = request.body.asJson.flatMap(
```
akash1810: It might be clearer to split this into two functions? One for the old path and one for the new?

app/controllers/UserDataController.scala
```diff
@@ -28,18 +31,48 @@ class UserDataController(dynamo: Dynamo, val deps: BaseFaciaControllerComponents
...
+      _.asOpt[Map[String, List[String]]])
+    Logger.info(maybeFrontIds.toString)
```
RichieAHB: Do we want to keep these in?

jonathonherbert: 'N' to the 'O'

app/controllers/UserDataController.scala
jonathonherbert: Yeah, I do prefer this. Good point, thanks 👍

jonathonherbert:
> how does V1 cope with this change?

Good question! V1 doesn't persist front ids on the server -- I think it uses local storage.

PR #626, opened 2019-03-15
Key front ids by priority on client

jonathonherbert:
## What's changed?

This PR alters the way the client models open fronts, keying them by priority. This fixes a bug in the current front list implementation.

## Implementation notes

This should only be merged once #625 is merged, and user data has been migrated.

Because we're touching a widely used selector here, there are lots of file changes, but most of them simply relate to adding priority to fronts selection where it was missing before (sometimes in error, e.g. when we're polling we only care about the current priority)
 
### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/UserDataController.scala
+++ b/app/controllers/UserDataController.scala
@@ -3,10 +3,13 @@ package controllers
 import com.gu.facia.client.models.Trail
 import com.gu.scanamo._
 import com.gu.scanamo.syntax._
-import services.Dynamo
+import services.{Dynamo, FrontsApi}
 import model.UserData
+import play.api.Logger
 
-class UserDataController(dynamo: Dynamo, val deps: BaseFaciaControllerComponents) extends BaseFaciaController(deps) {
+import scala.concurrent.{ExecutionContext, Future}
+
+class UserDataController(frontsApi: FrontsApi, dynamo: Dynamo, val deps: BaseFaciaControllerComponents)(implicit ec: ExecutionContext) extends BaseFaciaController(deps) {
   import model.UserData._
 
   private lazy val userDataTable = Table[UserData](config.faciatool.userDataTable)
@@ -28,18 +31,48 @@ class UserDataController(dynamo: Dynamo, val deps: BaseFaciaControllerComponents
   }
 
   def putFrontIds() = APIAuthAction { request =>
-
-    val frontIds: Option[List[String]] = request.body.asJson.flatMap(jsValue =>
-      jsValue.asOpt[List[String]])
-
-    frontIds match {
-      case Some(ids) => {
-        val userEmail = request.user.email
-        Scanamo.exec(dynamo.client)(userDataTable.update('email -> request.user.email, set('frontIds -> ids)))
+    // We handle both cases here, because of the migration from a list to a map by priority
+    val maybeFrontIds: Option[List[String]] = request.body.asJson.flatMap(
+      _.asOpt[List[String]])
+    val maybeFrontIdsByPriority: Option[Map[String, List[String]]] = request.body.asJson.flatMap(
+      _.asOpt[Map[String, List[String]]])
+    Logger.info(maybeFrontIds.toString)
+    Logger.info(maybeFrontIdsByPriority.toString)
+    (maybeFrontIds, maybeFrontIdsByPriority) match {
+      case (Some(frontIds), _) =>
+        Scanamo.exec(dynamo.client)(userDataTable.update('email -> request.user.email, set('frontIds -> frontIds)))
         Ok
-      }
-      case None => BadRequest
+      case (_, Some(frontIdsByPriority)) =>
+        Scanamo.exec(dynamo.client)(userDataTable.update('email -> request.user.email, set('frontIdsByPriority -> frontIdsByPriority)))
+        Ok
+      case _ => BadRequest
     }
+  }
+
+  def migrateUserData() = AccessAPIAuthAction.async {
+    val result = frontsApi.amazonClient.config.flatMap { config =>
+      val maybeUserData = Scanamo.exec(dynamo.client)(userDataTable.scan)
+      Future.successful(maybeUserData.filter(_.isRight).map {
+        case Right(userData) =>
+          val frontIds = userData.frontIds.getOrElse(List.empty[String])
+          val frontIdsByPriority = frontIds.foldLeft(Map.empty[String, List[String]])((acc, frontId) => {
+            val maybeAcc = for {
+              front <- config.fronts.get(frontId)
+              priority <- front.priority
+            } yield {
+              val frontIdsByCurrentPriority = acc.getOrElse(priority, List.empty[String])
+              acc + (priority -> (frontIdsByCurrentPriority :+ frontId))
+            }
+            maybeAcc.getOrElse(acc)
+          })
+          Scanamo.exec(dynamo.client)(userDataTable.update('email -> userData.email, set('frontIdsByPriority -> frontIdsByPriority)))
+          Map(userData.email -> frontIdsByPriority)
+      })
+    }
+    result.map(data => {
+      Logger.info(data.toString)
+      Ok
+    })
   }
 }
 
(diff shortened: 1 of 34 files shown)


client-v2/src/actions/Collections.ts
```diff
@@ -286,12 +304,100 @@ function getVisibleArticles(
   return fetchVisibleArticles(collectionType, articleDetails);
 }
 
+/**
```
jonathonherbert: These have been moved to prevent a circular dependency -- arguably as they deal with collections in the state, not fronts, they're better placed here.

client-v2/src/bundles/frontsUIBundle.ts
```diff
@@ -80,11 +79,11 @@ const editorCloseCurrentFrontsMenu = (): EditorCloseCurrentFrontsMenu => ({
  * involved. On balance we're going for it ...
  */
 
-const editorOpenFront = (frontId: string): EditorAddFront => {
+const editorOpenFront = (frontId: string, priority: string): EditorAddFront => {
   events.addFront(frontId);
   return {
     type: EDITOR_OPEN_FRONT,
-    payload: { frontId },
+    payload: { frontId, priority },
```
jonathonherbert: The new priority param, which you'll see threaded through innumerable places.

client-v2/src/bundles/frontsUIBundle.ts
```diff
@@ -220,17 +219,37 @@ const selectPriority = (
...
+    getFronts,
+    selectEditorFrontIds,
+    selectPriority,
+    (fronts, frontIdsByPriority, priority) => {
+      const openFrontIds = frontIdsByPriority[priority] || [];
+      return compact(openFrontIds.map(frontId => fronts[frontId]));
+    }
+  );
+
+const createSelectFrontIdAndOpenStateByPriority = () => {
```
jonathonherbert: A new selector for the FrontsList, which needed a refactor anyway because of the priority work.

client-v2/src/components/App.tsx
```diff
@@ -101,6 +101,10 @@ const BackgroundHeader = styled('div')`
   width: 100%;
 `;
 
+export const frontsEditPath = `/:priority(${Object.keys(priorities).join(
```
jonathonherbert: We need to derive priority from this route elsewhere, so it lives here and is exported.

client-v2/src/components/CurrentFrontsList.tsx
```diff
@@ -34,7 +38,7 @@ const FrontTabList = styled('div')<{ isDraggingOver?: boolean }>`
   width: calc(100vw - 120px);
   transition: background-color 0.15s;
   white-space: nowrap;
-  overflow: hidden;
+  overflow-x: scroll;
```
jonathonherbert: A scrolling container for the current front menu, @akemitakagi 👍

client-v2/src/components/FeedSectionHeader.tsx
```diff
@@ -136,10 +141,15 @@ const FeedSectionHeader = ({
   </SectionHeaderWithLogo>
 );
 
-const mapStateToProps = (state: State) => ({
-  isCurrentFrontsMenuOpen: selectIsCurrentFrontsMenuOpen(state),
-  frontCount: selectEditorFrontIds(state).length
-});
+const mapStateToProps = () => {
+  const selectEditorFrontsByPriority = createSelectEditorFrontsByPriority();
+  return (state: State, props: ContainerProps) => ({
```
jonathonherbert: We need to thread the current priority through this component to the CurrentFrontList component --  it didn't feel right for such a small component to depend upon the current route.

client-v2/src/components/FrontsEdit/Front.tsx
```diff
@@ -262,7 +262,7 @@ const mapDispatchToProps = (
   return {
     dispatch,
     initialiseFront: () =>
-      dispatch(initialiseFront(props.id, props.browsingStage)),
+      dispatch(initialiseCollectionsForFront(props.id, props.browsingStage)),
```
jonathonherbert: Just a rename to make the intent of this action clearer in its new context.

client-v2/src/util/pollingConfig.ts
```diff
@@ -1,14 +1,23 @@
...
  * TODO: do we want to check if there are any collectionUpdates going out here
  * could there be a race condition where we try to update a collection and poll
  * at the same time and the poll overwrites the update
  */
 export default (store: Store) =>
-  setInterval(
-    () => (store.dispatch as Dispatch)(fetchStaleOpenCollections()),
-    10000
-  );
+  setInterval(() => {
```
jonathonherbert: Previously, collection polling was asking for collections from the open fronts across all priorities. By deriving the current priority from the path, we can just poll for the collections open for the current priority.

client-v2/src/services/faciaApi.ts
```diff
@@ -166,7 +166,9 @@ async function saveClipboard(
   }
 }
 
-async function saveOpenFrontIds(frontIds?: string[]): Promise<void> {
+async function saveOpenFrontIds(frontIds?: {
+  [id: string]: string[];
```
RichieAHB: Does this `id` in `id: string` represent priority? If so can we rename it?

jonathonherbert: Good point, yes -- this could do with a rename.

jonathonherbert:
@RichieAHB has noted the possibility of finding the front rather than requiring priority in the actions, as frontIds are unique.

client-v2/src/bundles/frontsUIBundle.ts
```diff
@@ -346,7 +383,7 @@ const reducer = (state: State = defaultState, action: Action): State => {
     }
     case REMOVE_SUPPORTING_ARTICLE_FRAGMENT:
     case REMOVE_GROUP_ARTICLE_FRAGMENT:
-    case REMOVE_CLIPBOARD_ARTICLE_FRAGMENT: {
+    case 'REMOVE_CLIPBOARD_ARTICLE_FRAGMENT': {
       const articleFragmentId = action.payload.articleFragmentId;
       const selectedFrontId = Object.keys(state.selectedArticleFragments).find(
```
jonathonherbert: This is a string to remove a circular dependency -- handling this and keeping the constant would mean a bigger refactor, but we should definitely deal with this down the line in a more referential way.

app/controllers/UserDataController.scala
```diff
@@ -28,18 +31,48 @@ class UserDataController(dynamo: Dynamo, val deps: BaseFaciaControllerComponents
...
-      case Some(ids) => {
-        val userEmail = request.user.email
-        Scanamo.exec(dynamo.client)(userDataTable.update('email -> request.user.email, set('frontIds -> ids)))
+    // We handle both cases here, because of the migration from a list to a map by priority
+    val maybeFrontIds: Option[List[String]] = request.body.asJson.flatMap(
+      _.asOpt[List[String]])
+    val maybeFrontIdsByPriority: Option[Map[String, List[String]]] = request.body.asJson.flatMap(
+      _.asOpt[Map[String, List[String]]])
+    Logger.info(maybeFrontIds.toString)
+    Logger.info(maybeFrontIdsByPriority.toString)
```
Reettaphant: Why are these log lines here?

app/controllers/UserDataController.scala
```diff
@@ -28,18 +31,48 @@ class UserDataController(dynamo: Dynamo, val deps: BaseFaciaControllerComponents
...
+    (maybeFrontIds, maybeFrontIdsByPriority) match {
+      case (Some(frontIds), _) =>
+        Scanamo.exec(dynamo.client)(userDataTable.update('email -> request.user.email, set('frontIds -> frontIds)))
         Ok
-      }
-      case None => BadRequest
+      case (_, Some(frontIdsByPriority)) =>
+        Scanamo.exec(dynamo.client)(userDataTable.update('email -> request.user.email, set('frontIdsByPriority -> frontIdsByPriority)))
```
Reettaphant: Just to double check: does the data that we get from the client have editorial fronts marked as having editorial priority instead of empty priority?

jonathonherbert: Good question, and yes, it's explicit 👍

app/controllers/UserDataController.scala
jonathonherbert: Ah, an old rebase -- they should be gone.

PR #627, opened 2019-03-15
Rv fronts width

Reettaphant:
## What's changed?
Display narrower fronts if we have more than one front open to make it easier to see many at once on a screen.
Screenshot 2019-03-15 at 11 46 05
Screenshot 2019-03-15 at 11 46 30

## Implementation notes
Needed to move the frontsIds selector from the frontsUIBundle to frontsSelectors because it was causing circular decencies there. Probably worth looking into looking at how we split selectors in these two files in more detail.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/selectors/frontsSelectors.ts
+++ b/client-v2/src/selectors/frontsSelectors.ts
@@ -5,7 +5,6 @@ import { State } from 'types/State';
 import { AlsoOnDetail } from 'types/Collection';
 import { breakingNewsFrontId } from 'constants/fronts';
 import { selectors as frontsConfigSelectors } from 'bundles/frontsConfigBundle';
-import { selectEditorFrontIds } from 'bundles/frontsUIBundle';
 
 import { CollectionItemSets, Stages } from 'shared/types/Collection';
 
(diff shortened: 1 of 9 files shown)


client-v2/src/selectors/frontsSelectors.ts
```diff
@@ -20,6 +20,8 @@ interface FrontsByPriority {
   [id: string]: FrontConfig[];
 }
 
+const selectEditorFrontIds = (state: State) => state.editor.frontIds;
```
jonathonherbert: Would it be better if these selectors hung out with the frontsUIBundle? They relate to the UI state, not the fronts config state.

Reettaphant: Because frontsUIBundle depends on `getFronts` , we can't depend on anything from that file in frontsSelectors, so if `selectEditorFrontIds` is moved, `getFrontsWithPriority` should go as well.

jonathonherbert: Sorry, I've just seen in your implementation notes -- `Probably worth looking into looking at how we split selectors in these two files in more detail.` 👍 let's do that, tools like [madge](https://github.com/pahen/madge) might help!

PR #630, opened 2019-03-18
Presence

RichieAHB:
## What's changed?

This adds a basic implementation of presence.

ezgif com-video-to-gif

## Implementation notes

There are many things to note here
- The presence client is not that well documented so there may be some missing edge cases here!
- I've not added the presence domain client to the config so currently that's hard coded to `local`!
- The position / style of the presence indicators is well up for debate!
- The way of generating a unique key for our article fragments is using the internal code page - this works because of deduping. Additionally, there's quite a lot of boilerplate around this path creating that could do with refactoring and making a bit DRYer potentially.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/actions/__tests__/Collections.spec.ts
+++ b/client-v2/src/actions/__tests__/Collections.spec.ts
@@ -73,10 +73,13 @@ describe('Collection actions', () => {
 
   describe('Get Collections thunk', () => {
     beforeEach(() => fetchMock.reset());
-    const store = configureStore({
-      config,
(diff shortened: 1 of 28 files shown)


akash1810:
Given the DNM label and the rather big conflict, tempted to close this. Any objection @RichieAHB ?

akash1810:
Closing as fairly old and out of date - can reopen if we need to add Presence.

PR #642, opened 2019-03-20
Discard button

Reettaphant:
## What's changed?

We have a discard button which erases all changes made to draft after the article was launched last.

 @harpreetpurewal and @akemitakagi I had to make the collection metadata text smaller to fit the discard button in. 

I think we are going to have to think the way we display collection metadata anyway because currently we display only one metadata tag, we should have space to display multiple. Perhaps just adjust the front according to how many tags we have?

Screenshot 2019-03-20 at 14 29 03

## Implementation notes

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/actions/Collections.ts
+++ b/client-v2/src/actions/Collections.ts
@@ -3,7 +3,11 @@ import {
   getArticlesBatched,
   getCollections as fetchCollections,
   updateCollection as updateCollectionFromApi,
-  fetchVisibleArticles
+  discardDraftChangesToCollection as discardDraftChangesToCollectionApi,
+  fetchVisibleArticles,
+  fetchLastPressed as fetchLastPressedApi,
+  publishCollection as publishCollectionApi,
+  getCollection as getCollectionApi
 } from 'services/faciaApi';
 import { VisibleArticlesResponse } from 'types/FaciaApi';
 import {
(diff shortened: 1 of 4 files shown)


client-v2/src/actions/Collections.ts
```diff
@@ -390,6 +389,18 @@ function publishCollection(
   };
 }
 
+function discardDraftChangesToCollection(
+  collectionId: string
+): ThunkResult<Promise<void | string[]>> {
+  return (dispatch: Dispatch, getState: () => State) => {
+    return discardDraftChangesToCollectionApi(collectionId)
+      .then(() => dispatch(getCollections([collectionId])))
+      .catch(() => {
```
akash1810: Whilst we're waiting for error handling, it might be better to omit this empty `catch` and allow sentry to pick it up.

client-v2/src/services/faciaApi.ts
```diff
@@ -100,6 +100,27 @@ async function fetchVisibleArticles(
   }
 }
 
+async function discardDraftChangesToCollection(
+  collectionId: string
+): Promise<void> {
+  // The server does not respond with JSON
```
akash1810: should we make it?

Reettaphant: This endpoint is also used by v1 so unless we make a whole different api for v2 I'm not too keen to start changing it.

Reettaphant:
I'm merging this now as this feature has been anticipated for so long but I'll address comments in another pr.

PR #643, opened 2019-03-21
Header fixes

Reettaphant:
## What's changed?

Fixes a couple of issues with the header:
- Removes the weird black header on the right when we only have one front open. When we have only one front open we still get a lot extra space appearing on the right. Had a quick chat with @harpreetpurewal and we decided fixing this is not a priority. 
- Adds correct scrollbars to firefox so we scroll overview and collections separately rather than the whole front along with it's header!

Before: 
Screenshot 2019-03-21 at 12 12 07

After: 
Screenshot 2019-03-21 at 12 11 58

## Implementation notes

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/Front.tsx
+++ b/client-v2/src/components/FrontsEdit/Front.tsx
@@ -35,8 +35,11 @@ import CollectionItem from './CollectionComponents/CollectionItem';
 import { ValidationResponse } from 'shared/util/validateImageSrc';
 import { initialiseCollectionsForFront } from 'actions/Collections';
 
+// min-height required here to display scrollbar in Firefox:
+// https://stackoverflow.com/questions/28636832/firefox-overflow-y-not-working-with-nested-flexbox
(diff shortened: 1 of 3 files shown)


client-v2/src/components/FrontsEdit/Front.tsx
```diff
@@ -37,6 +37,7 @@ import { initialiseCollectionsForFront } from 'actions/Collections';
 
 const FrontContainer = styled('div')`
   display: flex;
+  min-height: 0;
```
akash1810: this is odd, my css skills aren't strong enough to fully understand why we'd do this...

Reettaphant: Apparently this is a firefox thing which is needed for the scroll bars to appear in a flex container. I'll add a comment to make this clearer.

PR #616, opened 2019-03-12
Proxy return same status code as CAPI

RichieAHB:
## What's changed?

Ensure the client receives errors from CAPI rather than a 200 when CAPI returns non-200s

## Implementation notes

I _think_ this is how to do it and it all seems to be working ok, I've just not had any errors from CAPI and testing a controller seems _arduous_.

Will also check on CODE first.

### General
- [ ] 🤖 Relevant tests added
- [x] ✅ CI checks / tests run locally
- [ ] 🔍 Checked on CODE

### Client
- [ ] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ ] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #644, opened 2019-03-21
increase debounce time

Reettaphant:
## What's changed?

Just increases the debounce time so that we don't see capi errors when people pause for a bit to add quotes etc. Played with it and it didn't seem to slow to me. I considered adding a new capi endpoint for v2 to respond with the capi status code as well instead of always returning 200 but since I'm not really sure what we would be using these codes for I'm not sure it's worth it.
## Implementation notes

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FeedsContainer.tsx
+++ b/client-v2/src/components/FeedsContainer.tsx
@@ -134,7 +134,7 @@ class FeedsContainer extends React.Component<
     super(props);
     this.debouncedRunSearchAndRestartPolling = debounce(
       () => this.runSearchAndRestartPolling(),
-      250
+      750
     );
   }
 


akash1810:
> I considered adding a new capi endpoint for v2 to respond with the capi status code as well instead of always returning 200

This would address that, right? https://github.com/guardian/facia-tool/pull/616

Reettaphant:
@akash1810 I had to revert that PR because it would've caused a large number of error messages in v1 so that's why it would have to be a separate v2 endpoint.

akash1810:
Ah, hadn't seen the revert. 👍

PR #645, opened 2019-03-21
move fetch-config.sh

akash1810:
## What's changed?
I really liked @sihil 's suggestion of using [this](https://github.com/github/scripts-to-rule-them-all) directory structure. Turns out only `fetch-config` was not in a `scripts` dir. It is now!

## Implementation notes

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


sihil:
Note that the spec uses `script` not `scripts`. Discuss.

sihil:
Also - the main benefit, from what I gather, is that the naming convention is consistent. Each project has a set of scripts with exactly the same name so as a user I know I can clone a repo and then run `script/setup && script/server` to get an instance up and running.

PR #658, opened 2019-03-27
Style search input field with Text Sans

paperboyo:
No idea if that’s a correct way/place to do this. I would rather style all sans-serif to be Text Sans (there shouldn’t be any system fonts as they are unpredictable), but don’t know how/where. Feel free to nuke this momentous contribution of mine!

search input


paperboyo:
I’ve noticed that Seach input box also has unstyled system font… won’t fix, because I feel there is a better way to do this up the cascade (?). Unless, components?

PR #662, opened 2019-03-28
Disable launch/discard when edit form is open

jonathonherbert:
## What's changed?

Disable launch/discard buttons if the edit form is open. This prevents people from mistakenly launching changes that haven't been saved, or discarding changes and re-saving previous edits over the article meta.

discard

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/package.json
+++ b/client-v2/package.json
@@ -121,7 +121,7 @@
     "redux-batched-actions": "^0.4.1",
     "redux-form": "^7.4.2",
     "redux-thunk": "^2.2.0",
-    "reselect": "^3.0.1",
+    "reselect": "^4.0.0",
     "sanitize-html": "^1.20.0",
     "styled-components": "^3.2.5",
     "ts-optchain": "^0.1.2",
--- a/client-v2/src/shared/selectors/__tests__/collection.spec.ts
+++ b/client-v2/src/shared/selectors/__tests__/collection.spec.ts
@@ -1,4 +1,7 @@
-import { selectArticlesInCollections } from '../collection';
+import {
+  selectArticlesInCollections,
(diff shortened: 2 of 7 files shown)


client-v2/package.json
```diff
@@ -121,7 +121,7 @@
     "redux-batched-actions": "^0.4.1",
     "redux-form": "^7.4.2",
     "redux-thunk": "^2.2.0",
-    "reselect": "^3.0.1",
+    "reselect": "^4.0.0",
```
jonathonherbert: The main difference for 4.0 -- the updated typings, which lets selectors share props objects which have nothing in common. For the selector in this PR, we rely on { collectionId: string, collectionItemSet: string } in one selector and { articleFragmentId: string } in the next. Previously, when we used these selectors in createSelector the typings would complain that the signatures had nothing in common (true, but still valid) -- now, all is 👌

client-v2/src/shared/selectors/__tests__/collection.spec.ts
```diff
@@ -32,4 +35,25 @@ describe('Collection selectors', () => {
...
+    it('should return true if the article is within a given collection', () => {
+      expect(
+        selectIsArticleInCollection(stateWithCollection.shared, {
+          collectionId: 'exampleCollection',
+          collectionSet: 'live',
+          articleFragmentId: '95e2bfc0-8999-4e6e-a359-19960967c1e0'
+        })
+      ).toEqual(true);
+    });
+    it("should return false if it's not", () => {
```
Reettaphant: Double quotes!

PR #674, opened 2019-04-04
Image display

Reettaphant:
## What's changed?

Network front editors wanted to see the capi trail image on the edit form when nothing had been overridden. To accommodate the request I replicated the imageReplace behaviour from the old tool where you have to explicitly select imageReplace and add an image from the grid to replace an image. Without this toggle it would have been unclear in the form if the image had been replaced.  

Screenshot 2019-04-04 at 07 25 57

## Implementation notes

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
+++ b/client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
@@ -10,6 +10,7 @@ import {
 } from 'redux-form';
 import { styled } from 'constants/theme';
 import Button from 'shared/components/input/ButtonDefault';
+import Thumbnail from 'shared/components/Thumbnail';
 import ContentContainer from 'shared/components/layout/ContentContainer';
 import ContainerHeadingPinline from 'shared/components/typography/ContainerHeadingPinline';
 import {
@@ -38,9 +39,9 @@ import {
   ArticleFragmentFormData,
   getArticleFragmentMetaFromFormValues,
   getInitialValuesForArticleFragmentForm,
-  getCapiValuesForArticleTextFields
+  getCapiValuesForArticleFields
 } from 'util/form';
-import { CapiTextFields } from 'util/form';
+import { CapiFields } from 'util/form';
 import { Dispatch } from 'types/Store';
 import { articleFragmentImageCriteria as imageCriteria } from 'constants/image';
 import { selectors as collectionSelectors } from 'shared/bundles/collectionsBundle';
@@ -168,7 +169,8 @@ class FormComponent extends React.Component<Props, FormComponentState> {
       showKickerTag,
       showKickerSection,
       frontId,
-      articleExists
+      articleExists,
+      imageReplace
     } = this.props;
 
     return (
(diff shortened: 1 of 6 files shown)


client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
```diff
@@ -357,18 +359,45 @@ class FormComponent extends React.Component<Props, FormComponentState> {
...
+                    permittedFields={editableFields}
+                    name="imageReplace"
+                    component={InputCheckboxToggle}
+                    label="Replace media"
+                    id={getInputId(articleFragmentId, 'image-replace')}
+                    type="checkbox"
+                    default={false}
+                    onChange={e => {
+                      change('imageHide', false);
+                      change('imageReplace', true);
```
Reettaphant: This is to have similar behaviour in the old and new tool and avoid weird behaviours if people switch between the two.

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
```diff
@@ -357,18 +359,45 @@ class FormComponent extends React.Component<Props, FormComponentState> {
...
+                      component={InputImage}
+                      disabled={imageHide}
+                      criteria={imageCriteria}
+                      frontId={frontId}
+                    />
+                  </ImageWrapper>
+                )}
+                {!imageReplace && (
+                  <ThumbnailEditForm
+                    style={{
```
jonathonherbert: It would be nice to have the component be responsible for its own styles here, e.g. image={img} hide={imageHide} -- not a blocker, though!

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
```diff
@@ -357,18 +359,45 @@ class FormComponent extends React.Component<Props, FormComponentState> {
...
+                      name="primaryImage"
+                      component={InputImage}
+                      disabled={imageHide}
+                      criteria={imageCriteria}
+                      frontId={frontId}
+                    />
+                  </ImageWrapper>
+                )}
+                {!imageReplace && (
+                  <ThumbnailEditForm
```
jonathonherbert: Just to check -- 

- what happens here when there's no image for the article? (not common, but I think possible?)
- I presume the difference between this and an image that the user had added in fronts is the presence of a rubbish bin?

Reettaphant: - Nothing particularly interesting happens if the image is missing, we just see a blank where an image should be.
- Yes, if the user has added the image there is a rubbish bin and the imageReplace toggle is turned on.

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
Reettaphant: 👍

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
jonathonherbert: 👍

PR #677, opened 2019-04-04
Add a scaffold for keyboard shortcuts, and a clipboard action to demonstrate how to use it

jonathonherbert:
## What's changed?

Adds a scaffold for keyboard shortcuts, and a clipboard paste action to demonstrate how to use it - the user can now focus the keyboard with a click and hit `ctrl-v` to paste an article or snaplink URL, or an article id, into the clipboard.

You're right --  a gif would be pointless for a keyboard shortcut. And here it is!

clipboard

(note the correct behaviour for snap links, etc.)

## Implementation notes

We represent the mapping between `shortcuts` and `actions` as `shortcut -> thing-that-is-focused -> action`, where `thing-that-is-focused` and any necessary accompanying data is stored in the application state at the point of focus. The boilerplate around this is in keyboard/index.ts, and see keyboard/keyboardActionMaps for the clipboard paste implementation. Me and @marialivia16 reasoned about a few scenarios and this arrangement seemed to cover them, but there may (will) be gaps or exceptions -- please take a look and feed back!

## Implementation notes

We've now added clipboard up and down navigations to this and it is ready for review. Going to chat to central prod before merging to make sure the arrow navigation isn't too confusion but changing the keybindings will always be the straightforward thing to do.

Not keeping the article which we are moving up/down focused is a bit of a glitch but adding it wasn't straightforward so we'd like to move on to collections now and fix the focus state later if we get time. 

I could do a gif but to get a feel for how it works it is better to just run the branch in local machine.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/components/Clipboard.tsx
```diff
@@ -84,89 +98,99 @@ class Clipboard extends React.Component<ClipboardProps> {
...
     return (
-      <StyledDragIntentContainer
-        active={!this.props.isClipboardOpen}
-        onDragIntentStart={() => this.setState({ preActive: true })}
-        onDragIntentEnd={() => this.setState({ preActive: false })}
-        onIntentConfirm={() => this.props.toggleClipboard(true)}
+      <ClipboardWrapper
+        tabIndex={0}
+        onFocus={this.handleFocus}
+        onBlur={this.handleBlur}
```
jonathonherbert: There's a question here as to how we keep the application focus state sync'ed with the DOM. We've opted for handling focus and blur events explicitly on the relevant components.

client-v2/src/components/Clipboard.tsx
```diff
@@ -62,13 +90,43 @@ interface ClipboardProps {
...
+  handleBlur: () => void;
   dispatch: Dispatch;
+  focusedArticle?: string;
+  isClipboardFocused: boolean;
 }
 
+// Styled component typings for ref seem to be broken so any refs
+// passed to styled components has to be any for now.
+type Ref = any;
+type TClipboardWrapper = any;
```
Reettaphant: Sorry, couldn't do better than this

client-v2/src/keyboard/keyboardActionMaps/move.ts
```diff
@@ -0,0 +1,42 @@
+import Raven from 'raven-js';
+import { KeyboardActionMap, ApplicationFocusStates } from 'keyboard';
+import { Dispatch } from 'types/Store';
+import { keyboardArticleFragmentMove } from 'actions/KeyboardNavigation';
+
+const moveUp: KeyboardActionMap = {
```
Reettaphant: There is some potential for cleaning up here and removing code duplication, going to look at refactoring when we start doing the collection actions

client-v2/src/components/Clipboard.tsx
RichieAHB: That should work with a bit of HTMLElement but it's not the end of the world and styled components may well be the issue here

client-v2/src/keyboard/keyboardActionMaps/move.ts
RichieAHB: Make sense 👌

client-v2/src/keyboard/keyboardActionMaps/paste.ts
```diff
@@ -0,0 +1,31 @@
...
+import { insertClipboardArticleFragment } from 'actions/Clipboard';
+import { createArticleFragment } from 'shared/actions/ArticleFragments';
+
+const paste: KeyboardActionMap = {
+  clipboard: (_: ApplicationFocusStates) => async (dispatch: Dispatch) => {
+    try {
+      if (!navigator || !(navigator as any).clipboard) {
+        throw new Error('No navigator available on paste');
+      }
+      // A temporary any here pending proper typings
```
RichieAHB: Link for context: https://github.com/Microsoft/TypeScript/issues/26728

client-v2/src/selectors/__tests__/keyboardNavigationSelectors.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+      nextIndexAndGroupSelector(stateWithGroups, 'group1', 'fragment-3', 'down')
+    ).toEqual({ toIndex: 0, nextGroupId: 'group2' });
+  });
+
+  it('return next group id when moving up between groups', () => {
+    expect(
+      nextIndexAndGroupSelector(stateWithGroups, 'group2', 'fragment-4', 'up')
+    ).toEqual({ toIndex: 3, nextGroupId: 'group1' });
+  });
+});
```
RichieAHB: Top testing 🔥

client-v2/src/components/Clipboard.tsx
Reettaphant: Yes, I couldn't make it work with HTMLElement and seems to be an ongoing issue with styled components.

PR #680, opened 2019-04-11
Move collection articles with keyboard

Reettaphant:
## What's changed?

Move articles inside collections and between collections using up and down arrows. 

Annoyingly we don't have a good solution for keyboard shortcuts for selecting specific fronts/collections yet as mousetrap does not understand regexes in keyboard commands.

## Implementation notes

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/keyboard/index.ts
+++ b/client-v2/src/keyboard/index.ts
@@ -24,6 +24,7 @@ interface BaseFocusState {
   articleFragment?: ArticleFragment;
   groupId?: string;
   collectionId?: string;
+  frontId?: string;
 }
 
 export type ApplicationFocusStates = BaseFocusState;
--- a/client-v2/src/keyboard/index.ts
+++ b/client-v2/src/keyboard/index.ts
@@ -22,6 +22,7 @@ type FocusableTypes =
 interface BaseFocusState {
   type: FocusableTypes;
   articleFragment?: ArticleFragment;
+  groupId?: string;
   collectionId?: string;
 }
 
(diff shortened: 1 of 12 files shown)


client-v2/src/keyboard/keyboardActionMaps/move.ts
```diff
@@ -36,6 +53,23 @@ const moveDown: KeyboardActionMap = {
     } catch (e) {
       Raven.captureMessage(`Moving item up in clipboard failed: ${e.message}`);
```
jonathonherbert: Presumably this is `down` 👍

client-v2/src/keyboard/keyboardActionMaps/move.ts
```diff
@@ -36,6 +53,23 @@ const moveDown: KeyboardActionMap = {
...
+        keyboardArticleFragmentMove(
+          'down',
+          'collection',
+          focusData.articleFragment,
+          focusData.groupId,
+          focusData.frontId
+        )
+      );
+    } catch (e) {
+      Raven.captureMessage(`Moving item up in clipboard failed: ${e.message}`);
```
jonathonherbert: As above

client-v2/src/keyboard/keyboardActionMaps/move.ts
jonathonherbert: Presumably this is `down` 👍

client-v2/src/keyboard/index.ts
```diff
@@ -13,11 +13,18 @@ import paste from './keyboardActionMaps/paste';
 import { moveUp, moveDown } from './keyboardActionMaps/move';
 import { ArticleFragment } from '../shared/types/Collection';
 
-type FocusableTypes = 'clipboard' | 'clipboardArticle';
+type FocusableTypes =
```
jonathonherbert: We could express BaseFocusState as an intersection of different `FocusableTypes` and their unique properties here, a bit like Action types in redux -- this might save us some null checking later on! Not a blocker but might be handy 👍

PR #681, opened 2019-04-15
Fetch missing collections as well

Reettaphant:
## What's changed?

Fixes a bug where new/automated collections were always displaying as loading and we were not able to add content in them. This is because we weren't correctly handling the case of the collection.json file not existing for these collections. This PR fixes this by constructing an empty collection from the collectionConfig if a separate collection json file does not exist.

## Implementation notes

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/actions/Collections.ts
+++ b/client-v2/src/actions/Collections.ts
@@ -45,6 +45,7 @@ import {
 import { actions as collectionActions } from 'shared/bundles/collectionsBundle';
 import { getCollectionConfig, getFront } from 'selectors/frontsSelectors';
 import { Dispatch, ThunkResult } from 'types/Store';
+import { Action } from 'types/Action';
 import {
   collectionItemSets,
   noOfOpenCollectionsOnFirstLoad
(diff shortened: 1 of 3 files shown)


client-v2/src/actions/Collections.ts
```diff
@@ -179,11 +205,27 @@ function getCollections(
           id
         })
       );
+
+      let missingCollectionActions: Action[][];
```
Reettaphant: Decided using let was more readable than a reduce.

client-v2/src/actions/Collections.ts
```diff
@@ -102,6 +103,31 @@ function fetchStaleOpenCollections(
   };
 }
 
+// These are collections which exist in the config but have never had any content
+// added to them so they will not have their own collection file. They may be newly
+// created collections or fully automated collections.
+function getCollectionActionForMissingCollection(
```
Reettaphant: Thought a separate function is tidier than a lot of if/elses in the old getCollectionActions function

client-v2/src/actions/Collections.ts
jonathonherbert: Agreed, it felt like a filter -> map 👍

PR #685, opened 2019-04-16
Optimise dnd

RichieAHB:
## What's changed?

This speeds up the drag and drop. There are still some issues in there that could stop unnecessary re-renders but it doesn't seem necessary to keep optimising. One example here is the new `focusedArticle`, which - when changed (i.e. when dragging / dropping) will trigger a whole re-render of the a front, but only on focus - so not on drop.

## Implementation notes

I've added a `createShallowCompareResultsSelector`, which will shallow compare the results of a selector and, if they're the same, return the old reference to ensure React / redux doesn't force a re-render on the basis of, say a new array who contents are the same.

There weren't many tests to add as the interfaces largely stayed the same here but I've changed them where they were needed.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/util/collectionUtils.ts
```diff
@@ -2,11 +2,26 @@ import React from 'react';
 import { ThunkResult, Dispatch } from 'types/Store';
 import { PosSpec } from 'lib/dnd';
 import { insertArticleFragment } from 'actions/ArticleFragments';
+import { CapiArticle } from 'types/Capi';
 
-const dropToArticle = (e: React.DragEvent): string | null => {
+export interface RefDrop {
+  type: 'REF';
```
jonathonherbert: What does 'REF' mean? (I have it mentally pegged as 'anything that hasn't come from the feed')

client-v2/src/shared/util/selectorUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+      return lastResult;
+    } else {
+      last = [args, result];
+      return result;
+    }
+  };
+}
+
+const createShallowEqualResultSelector = createSelectorCreator(
+  resultCheckMemoize as any,
```
jonathonherbert: 👍  but also 😢

client-v2/src/shared/actions/ArticleFragments.ts
```diff
@@ -145,9 +146,15 @@ const persistAndReturnFragment = (
  *  - an external link.
  */
 function createArticleFragment(
-  resourceId: string
+  drop: MappableDropType
 ): ThunkResult<Promise<ArticleFragment | undefined>> {
   return async (dispatch: Dispatch) => {
+    if (drop.type === 'CAPI') {
```
jonathonherbert: Brilliant, this will be much faster!

client-v2/src/components/FrontsEdit/Front.tsx
```diff
@@ -343,16 +345,13 @@ const mapDispatchToProps = (
     dispatch,
     initialiseFront: () =>
       dispatch(initialiseCollectionsForFront(props.id, props.browsingStage)),
-    selectArticleFragment: (
-      frontId: string,
+    selectArticleFragment: (frontId: string) => (
```
jonathonherbert: Hardly a blocker, but is there a reason this is curried?

client-v2/src/components/FrontsEdit/CollectionComponents/CollectionItem.tsx
```diff
@@ -161,7 +161,9 @@ const mapDispatchToProps = (dispatch: Dispatch) => {
         dispatch(
           insertArticleFragment(
             { id: 'clipboard', type: 'clipboard', index: 0 },
-            id,
+            // @TODO - if this is proving too slow we can pass the whole external article into
```
jonathonherbert: Does this @todo still apply? (looking at transformExternalArticle etc. below)

client-v2/src/components/FrontsCAPIInterface/FeedItem.tsx
```diff
@@ -148,62 +142,57 @@ const getArticleLabel = (
...
           <ScheduledPublication>
             {distanceInWordsStrict(
-              new Date(scheduledPublicationDate),
+              new Date(article.fields.scheduledPublicationDate),
               Date.now()
             )}
           </ScheduledPublication>
         )}
-        {publicationDate && (
+        {article.webPublicationDate && (
```
jonathonherbert: Presumably this is a just a correction you spotted during your travels 👍

client-v2/src/components/FrontsCAPIInterface/Feed.tsx
```diff
@@ -29,25 +25,7 @@ const Feed = ({ articles = [], error }: FeedProps) => (
...
-              article.fields && getId(article.fields.internalPageCode)
-            }
-            firstPublicationDate={article.fields.firstPublicationDate}
-            isLive={!article.fields.isLive || article.fields.isLive === 'true'}
-            scheduledPublicationDate={article.fields.scheduledPublicationDate}
-            thumbnail={getThumbnail(article, article.frontsMeta.defaults)}
-            tone={article.frontsMeta.tone}
-          />
-        ))
+        .map(article => <FeedItem key={article.id} article={article} />)
```
jonathonherbert: Heh. Just for the sake of cleanliness, or were there optimisation gains to be had here? (Possibly getThumbnail?)

client-v2/src/shared/util/selectorUtils.ts
RichieAHB: I should add a comment here ...

client-v2/src/components/FrontsEdit/CollectionComponents/CollectionItem.tsx
RichieAHB: Yeah it does, currently we don't have the external article in the scope of a`CollectionItem`. So either I could add another `MappableDropType` (maybe not the nicest name). So `CAPI`, `REF` and now `CAPI_ARTICLE_ID_THAT_IS_ALREADY_FETCHED_AND_IN_EXTERNAL)ARTICLES_I_PROMISEEEE` that would mean we could just pass a CAPI id and assume it didn't need fetching (bit dangerous?). Or otherwise pass in the whole external article to the `COLLECTION_ITEM` and change `REF` to `CAPI`. Does that make _any_ sense?

client-v2/src/components/FrontsCAPIInterface/FeedItem.tsx
RichieAHB: No actually - it's not that clear I did this - but because I needed the whole article inside a `FeedItem` now I just ditched all of the props and passed the `article` as a prop. Previously we were calling a `FeedItem` like `<FeedItem publicationDate={article.webPublicationDate} />`.

client-v2/src/components/FrontsCAPIInterface/Feed.tsx
RichieAHB: Ah as above really, just needed the whole article in there for serialisation so decided it would be cleaner to remove the props that were just derived from an article.

client-v2/src/util/collectionUtils.ts
jonathonherbert: 👍

client-v2/src/components/FrontsCAPIInterface/FeedItem.tsx
jonathonherbert: Ah, gotcha.

client-v2/src/components/FrontsCAPIInterface/Feed.tsx
jonathonherbert: 👍

PR #686, opened 2019-04-17
Prevent drop zones from blocking cursor events when a drag is not occurring

jonathonherbert:
## What's changed?

Prevent drop zones from blocking cursor events when a drag is not occurring. Drop zones obscure the bottom part of articles, which affects hover and drag operations.

This change gives us a little more freedom to style our drop zones without worrying that they get in the way of other cursor operations, so dropzones for supporting articles have been expanded to be less fiddly, and the space between articles has been reduced by 5px, giving us about an extra article's worth of space on the 900 or so px you get on a MacBook screen 👍 

## Implementation notes

It'd be great to have @RichieAHB's input here to check that my changes are consistent with his lib code!

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/lib/dnd/DropZone.tsx
+++ b/client-v2/src/lib/dnd/DropZone.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { StoreConsumer } from './Root';
 import { Store, Sub } from './store';
 import { NO_STORE_ERROR } from './constants';
+import { styled } from 'constants/theme';
 
 interface OuterProps {
   parentKey: string;
@@ -17,10 +18,15 @@ type Props = OuterProps & ContextProps;
 
 interface State {
   isTarget: boolean;
+  isActive: boolean;
 }
 
+const DropZoneContainer = styled.div<{ isActive: boolean }>`
+  ${({ isActive }) => !isActive && 'pointer-events: none'}
+`;
+
 class DropZone extends React.Component<Props, State> {
-  public state = { isTarget: false };
+  public state = { isTarget: false, isActive: false };
 
   public componentDidMount() {
     if (!this.props.store) {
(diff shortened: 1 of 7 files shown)


client-v2/src/lib/dnd/DropZone.tsx
```diff
@@ -17,10 +18,15 @@ type Props = OuterProps & ContextProps;
 
 interface State {
   isTarget: boolean;
+  isActive: boolean;
 }
 
+const DropZoneContainer = styled.div<{ isActive: boolean }>`
+  ${({ isActive }) => !isActive && 'pointer-events: none'}
```
jonathonherbert: We could also choose not to render these elements, but this approach minimizes the impact on the DOM. Interested to hear any thoughts.

client-v2/src/lib/dnd/Level.tsx
```diff
@@ -146,6 +146,13 @@ class Level<T> extends React.Component<Props<T>> {
     }
 
     e.preventDefault();
+
+    // Reset the dragging state on drop
```
RichieAHB: I think this resetting is handled by the onDragEnd in the Root so needn’t happen here?

client-v2/src/lib/dnd/DropZone.tsx
```diff
@@ -37,10 +43,19 @@ class DropZone extends React.Component<Props, State> {
...
-    return this.props.children(this.state.isTarget);
+    return (
+      <DropZoneContainer isActive={this.state.isActive}>
+        {this.props.children(this.state.isTarget)}
+      </DropZoneContainer>
+    );
   }
 
   private handleStoreUpdate: Sub = (id, hoverIndex) => {
+    const state = this.props.store && this.props.store.getState();
```
RichieAHB: This is definitely the right way to do this now put potentially some form of PureComponent wrapper might be a bit nicer here in future given we’re just checking for changes everywhere.

client-v2/src/lib/dnd/DropZone.tsx
RichieAHB: This sounds reasonable to me!

client-v2/src/lib/dnd/DropZone.tsx
jonathonherbert: 👍

client-v2/src/lib/dnd/Level.tsx
jonathonherbert: You're absolutely right.

PR #687, opened 2019-04-18
Debounce save requests and add tests

RichieAHB:
## What's changed?

This change introduces debouncing to saves. This means that collections will only fire one save request for example: a thunk with multiple actions that might have previously triggered multiple save requests. This was happening in `moveArticleFragment` before this change.

These tests have highlighted another issue that wasn't being spotted before now. Namely that when moving "upwards" between collections. The fetch requests to save the collection that had just had the article removed was not being fired.

This is being caused at the crossover of the persistence collection-finding logic and the way the collection cap logic has been handled. We have inserted the moved article fragment into its new position before firing the remove action. When we fire the remove action, the persistence middleware finds the articleFragment in its new collection first (this collection appears first when searching for it) and misses the actual collection we want to save - the place where it previously was, which is in a collection further along in the collections map.

I've not fixed this as it was existing before this change but there is a commented out test that tests this issue, so will pick this up in another PR. 

## Implementation notes

Relatively simple debouncing logic, although it's worth noting (as mentioned in a comment) that we have to change the debounce time in the tests, despite using Jest's fake timers. See https://github.com/facebook/jest/issues/3465

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/actions/__tests__/Persistence.spec.ts
+++ b/client-v2/src/actions/__tests__/Persistence.spec.ts
@@ -1,0 +1,207 @@
+import { createStore, compose, applyMiddleware } from 'redux';
+import { persistCollectionOnEdit } from 'util/storeMiddleware';
+import thunk from 'redux-thunk';
+import { enableBatching } from 'redux-batched-actions';
+import { moveArticleFragment } from 'actions/ArticleFragments';
+import { Dispatch } from 'types/Store';
+import fetchMock from 'fetch-mock';
+import rootReducer from 'reducers/rootReducer';
+import { NestedArticleFragment } from 'shared/types/Collection';
+import { updateCollection } from 'actions/Collections';
+
+const A1 = {
+  uuid: 'a1',
+  id: 'a',
+  frontPublicationDate: 1000,
+  meta: {}
+};
+
+const A2 = {
+  uuid: 'a2',
+  id: 'b',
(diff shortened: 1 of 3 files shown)


client-v2/src/actions/__tests__/Persistence.spec.ts
```diff
@@ -0,0 +1,208 @@
+import { createStore, compose, applyMiddleware } from 'redux';
```
jonathonherbert: This test covers a good surface area, nice to see 👍

client-v2/src/actions/__tests__/Persistence.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+      });
+      const c2afs = groupedArticleFragmentIds(
+        JSON.parse(calls[1][1].body).collection.live
+      );
+      expect(c2afs).toEqual({
+        g2: ['a', 'c']
+      });
+    });
+
+    // @TODO - fix this issue in another PR
```
jonathonherbert: Excellent spot 👍

client-v2/src/actions/__tests__/Persistence.spec.ts
RichieAHB: I wonder whether I should go through and test more actions that might exhibit this behaviour ...

PR #690, opened 2019-04-18
Dry run the deduping before applying any changes

RichieAHB:
## What's changed?

Our deduplication logic lead to inconsistent states when persisting. It ran inserts on the state, which caused the same article fragment to be in two places at once. Sometimes this inconsistency was very brief - only within the same stack - sometimes it was while a collection config modal was open. Either way it was long enough to break some invariants for our persistence layer, and it could also behave oddly when a modal was opened and collection polling overwrote a collection.

This now allows an external caller to run the reducer behaviour outside the reducer in order to reflect on the changes that will be made before they are committed. This means we don't have to commit any changes to the state until we're ready, ensuring that we're not in an inconsistent state for any period of time.

This is still up for discussion, hence the DO NOT MERGE. Also, there is [a test here](https://github.com/guardian/facia-tool/compare/rahb/jsh/upwards-collection-move?expand=1#diff-bce28573bf43e4679e120998f0484767R299) that was working before this change and now isn't and - on reflection - _looks_ like it shouldn't have been working before. So that needs some more investigation to make sure this is testing the right thing, and why it wasn't working before.

## Implementation notes

There is a minor coupling between how we figure out whether a group belongs to a collection that has exceeded its collection cap here but I think it's fine. Because we don't actually update the redux state object we can only count the article fragments in the group we get back in the `patch` rather than have the `groupSiblingsArticleCountSelector` work this our for us. Under the hood the selector calls the same function (it is exported for this reason) to derive the same number, but perhaps there is a nicer way to do this.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/actions/ArticleFragments.ts
+++ b/client-v2/src/actions/ArticleFragments.ts
@@ -120,9 +120,8 @@ const maybeInsertGroupArticleFragment = (
               applyBeforeReducer: true
             })(id, collectionCap)
           ]),
-          // otherwise just undo the insertion and don't persist as nothing
-          // has actually changed
-          [removeGroupArticleFragment(id, articleFragmentId)]
+          // otherwise do nothing
+          []
         )
       );
     } else {
--- a/client-v2/src/actions/ArticleFragments.ts
+++ b/client-v2/src/actions/ArticleFragments.ts
@@ -86,15 +86,6 @@ const maybeInsertGroupArticleFragment = (
   removeAction: Action | null
 ) => {
   return (dispatch: Dispatch, getState: () => State) => {
-    // run the action and put the article fragment into the group
-    // if this was triggered with a move, this will be the same article fragment
-    // with the same uuid as the moved article fragment and until the modal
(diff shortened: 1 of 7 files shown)


client-v2/src/actions/ArticleFragments.ts
```diff
@@ -118,14 +111,15 @@ const maybeInsertGroupArticleFragment = (
...
+            insertGroupArticleFragment(id, index, articleFragmentId),
+            maybeAddFrontPublicationDate(articleFragmentId),
             addPersistMetaToAction(capGroupSiblings, {
               id: articleFragmentId,
               persistTo,
               applyBeforeReducer: true
             })(id, collectionCap)
-          ],
+          ]),
           // otherwise just undo the insertion and don't persist as nothing
```
jonathonherbert: A quick note to say we shouldn't need this anymore! Yesss 🎉

RichieAHB: Good eyes 👀

PR #695, opened 2019-04-23
Fix article visibility details

RichieAHB:
## What's changed?

This fixes an issue where our `Front` render function was impure and as a result, would not render the visibility icons in all cases. Specifically this was happening when components below a `Front` were re-rendering but the front was not -  due to optimisations we'd made around drag and drop. I've refactored the selector and also moved a bit of logic to a _different_ `Collection` component, to make selecting this easier.

Fixes: https://trello.com/c/1j5jwQsE/543-adding-new-articles-to-collections-should-update-mobile-desktop-indicators

## Implementation notes

I've assumed the visibility numbers are not zero-indexed so I'm subtracting 1 here in order to get the index of the ordered article fragments to pick from, which I think follows what was being done before. I've also added a test for this.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/Collection.tsx
+++ b/client-v2/src/components/FrontsEdit/Collection.tsx
@@ -20,7 +20,7 @@ import {
 import { resetFocusState, setFocusState } from 'bundles/focusBundle';
 import { connect } from 'react-redux';
 import { State } from 'types/State';
-import { createArticleVisiblityDetailsSelector } from 'selectors/frontsSelectors';
+import { createArticleVisibilityDetailsSelector } from 'selectors/frontsSelectors';
 
 const getArticleNotifications = (
   id: string,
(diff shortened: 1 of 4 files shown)


client-v2/src/components/FrontsEdit/Collection.tsx
```diff
@@ -0,0 +1,246 @@
+import React from 'react';
```
RichieAHB: This rather large diff is me moving out the `Collection` to a separate container so that we can use a `mapStateToProp` factory in order to then create the `articleVisibilityDetailSelector` for each collection in a tidier way.

jonathonherbert: Ah, this is nice to see, the Fronts component was very large!

client-v2/src/components/FrontsEdit/Collection.tsx
```diff
@@ -0,0 +1,246 @@
...
+        </Collection>
+      </CollectionWrapper>
+    );
+  }
+}
+
+const createMapStateToProps = () => {
+  const articleVisiblityDetailsSelector = createArticleVisiblityDetailsSelector();
+  return (state: State, props: CollectionContextProps) => {
+    const articleVisiblityDetails = articleVisiblityDetailsSelector(state, {
```
jonathonherbert: Not a blocker, but tiiiny typo 😁

client-v2/src/selectors/__tests__/frontsSelectors.spec.ts
```diff
@@ -163,3 +164,72 @@ describe('Selecting which front collection is also on correctly', () => {
...
+        uuid: 'a5'
+      },
+      a6: {
+        uuid: 'a5'
+      }
+    }
+  }
+};
+
+describe('Article visibility selector', () => {
```
jonathonherbert: 👍

client-v2/src/selectors/frontsSelectors.ts
```diff
@@ -282,17 +287,49 @@ const visibleArticlesSelector = createSelector(
...
-    return collectionVisibilities[collectionSet];
-  }
-);
+  ) =>
+    articlesInCollectionSelector(selectSharedState(state), {
+      ...extra,
+      includeSupportingArticles: false
+    });
+
+  return createShallowEqualResultSelector(
```
jonathonherbert: 👍

RichieAHB:
:scout

> Well spotted, I certainly didn't see it! 👀

Trying to keep it scouty!

PR #698, opened 2019-04-24
Add hr between clipboard sublinks

RichieAHB:
## What's changed?

As title suggests! Also reduced font sizes as per: https://trello.com/c/vJPSVd7W/473-clipboard (and a margin below the picture because why not)

Screenshot 2019-04-24 at 14 25 04

## Implementation notes

N/A

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/Clipboard.tsx
+++ b/client-v2/src/components/Clipboard.tsx
@@ -83,6 +83,13 @@ const StyledDragIntentContainer = styled(DragIntentContainer)`
   height: 100%;
 `;
 
+const SupportingDivider = styled.hr`
+  border: 0;
+  border-top: 1px solid #ccc;
+  margin: 0.5em 0 0.25em;
+  width: 50%;
+`;
+
 interface ClipboardProps {
   selectArticleFragment: (id: string, isSupporting?: boolean) => void;
   clearArticleFragmentSelection: () => void;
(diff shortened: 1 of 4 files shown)


client-v2/src/components/Clipboard.tsx
```diff
@@ -211,24 +218,27 @@ class Clipboard extends React.Component<ClipboardProps> {
...
-                                this.props.selectArticleFragment(id, true)
-                              }
-                              onDelete={() =>
-                                this.props.removeSupportingCollectionItem(
-                                  articleFragment.uuid,
-                                  supporting.uuid
-                                )
-                              }
-                            />
+                          {(supporting, sProps, i, arr) => (
```
jonathonherbert: Could `arr` be named more clearly? I guess it refers to all elements in the list?

jonathonherbert: (obviously not a blocker, it caught my eye just as I'd approved!)

RichieAHB: I missed this, I just treated it like the `arr` arg in a `Array#map` so named it like it's named in MDN. But you're right it probably should be clearer - I should probably do that from now on 😞

PR #702, opened 2019-04-24
Update docs with naming convention descriptions/aspirations

jonathonherbert:
## What's changed?

Documented a few of the naming decisions we've made, and acknowledged some inconsistencies as tech debt. Let me know if I've missed anything, and feedback welcome!

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/types/Collection.ts
+++ b/client-v2/src/shared/types/Collection.ts
@@ -8,8 +8,10 @@ interface Group {
   articleFragments: string[];
 }
 
-type Stages = 'draft' | 'live';
+// CollectionItemSets represent all of the lists of collectionItems available in a collection.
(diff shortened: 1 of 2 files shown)


client-v2/src/shared/types/Collection.ts
```diff
@@ -8,8 +8,10 @@ interface Group {
   articleFragments: string[];
 }
 
-type Stages = 'draft' | 'live';
+// CollectionItemSets represent all of the lists of collectionItems available in a collection.
```
jonathonherbert: I've added this here as I'm not convinced developers will find it in a readme when they most need it, but happy to move it if anyone else has strong opinions.

PR #706, opened 2019-04-24
Remove redundant build step

RichieAHB:
## What's changed?

The integration test runs the build again, why not save time and do our actual build first, then test with that? 🚧 

## Implementation notes
N/A

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/package.json
+++ b/client-v2/package.json
@@ -14,7 +14,8 @@
     "dev-server": "yarn compileWithBabel $(yarn bin)/webpack-dev-server --config config/webpack.config.dev.js",
     "dev": "yarn install && yarn dev-server --open",
     "test": "yarn jest",
-    "test-integration": "BUILD_ENV=integration echo 'RUNNING ON EXISTING PROD BUILD - run yarn build before this to catch changes since last build' && node ./integration/run.js",
+    "test-integration": "BUILD_ENV=integration yarn build && node ./integration/run.js",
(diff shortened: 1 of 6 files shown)


client-v2/package.json
```diff
@@ -14,7 +14,7 @@
     "dev-server": "yarn compileWithBabel $(yarn bin)/webpack-dev-server --config config/webpack.config.dev.js",
     "dev": "yarn install && yarn dev-server --open",
     "test": "yarn jest",
-    "test-integration": "BUILD_ENV=integration yarn build && node ./integration/run.js",
+    "test-integration": "BUILD_ENV=integration echo 'RUNNING ON EXISTING PROD BUILD - run yarn build before this to catch changes since last build' && node ./integration/run.js",
```
jonathonherbert: Might be worth two commands to differentiate between dev (where we probably want the convenience of build + run) and prod (where we just want run)?

jonathonherbert:
Somebody stop this PR machine! 👍

RichieAHB:
Not sure why this isn't triggering a test build - will look at tomorrow ...

client-v2/package.json
RichieAHB: I thought about this but just didn't know what to call it - I will append `ci` to the one that doesn't built.

PR #718, opened 2019-04-30
CLIPBOARD: Stop metadata loss when "add to clipboard" button is used 

ajwl:
## What's changed?

When articles were taken OUT of a collection and added to the clipboard, any meta changes that had been made to the story were lost. (eg. any edits to the headline or supporting articles added).

Because drag and drop between the collection and the clipboard worked fine, I have reused this logic for the button. This required a new thunk to be created. 

So this now uses: `moveArticleFragment`
Instead of:  `insertArticleFragment` 

A test has also been added. 

A styling change to sublinks has also been added to this PR 

https://trello.com/c/tdNV4sx7/550-copying-articles-to-clipboard-loses-the-furniture-of-the-article

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/actions/ArticleFragments.ts
+++ b/client-v2/src/actions/ArticleFragments.ts
@@ -11,7 +11,8 @@ import {
 import { ArticleFragment } from 'shared/types/Collection';
 import {
   selectSharedState,
-  articleFragmentsSelector
+  articleFragmentsSelector,
+  articleFragmentSelector
 } from 'shared/selectors/shared';
 import { ThunkResult, Dispatch } from 'types/Store';
 import { addPersistMetaToAction } from 'util/storeMiddleware';
@@ -289,6 +290,21 @@ const moveArticleFragment = (
   };
 };
 
+const cloneArticleFragmentToTarget = (
+  uuid: string,
+  toType: 'clipboard' | 'collection'
+): ThunkResult<void> => {
+  return (dispatch, getState) => {
+    const to = { id: toType, type: toType, index: 0 };
+    const fragment = articleFragmentSelector(
+      selectSharedState(getState()),
+      uuid
+    );
+    const from = null;
+    dispatch(moveArticleFragment(to, fragment, from, toType));
+  };
+};
+
 const addImageToArticleFragment = (
   uuid: string,
   imageData: ValidationResponse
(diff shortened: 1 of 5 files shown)


client-v2/src/actions/ArticleFragments.ts
```diff
@@ -289,6 +290,21 @@ const moveArticleFragment = (
...
+    const to = { id: toType, type: toType, index: 0 };
+    const fragment = articleFragmentSelector(
+      selectSharedState(getState()),
+      uuid
+    );
+    const from = null;
+    dispatch(moveArticleFragment(to, fragment, from, toType));
+  };
+};
+
```
jonathonherbert: thunk-tastic 👍

client-v2/src/actions/__tests__/utils.ts
```diff
@@ -7,20 +7,23 @@ export interface ArticleFragmentMap {
 }
 
 export type ArticleFragmentSpec = [
-  string,
-  string,
-  Array<[string, string]> | undefined
+  string, // uuid
+  string, // id
+  Array<[string, string]> | undefined, // all of supporting articles,
+  object // metadata changes
```
jonathonherbert: Super minor: we have a type, ArticleFragmentMeta, that could be used to make the comment here redundant 👍

client-v2/src/actions/__tests__/ArticleFragments.spec.ts
```diff
@@ -47,19 +49,19 @@ const root = (state: any = {}, action: any) => ({
 
 const buildStore = (added: ArticleFragmentSpec, collectionCap = Infinity) => {
   const groupA: ArticleFragmentSpec[] = [
-    ['a', '1', [['g', '7']]],
-    ['b', '2', undefined],
-    ['c', '3', undefined]
+    ['a', '1', [['g', '7']], {}],
```
jonathonherbert: Minor: tuples can have optional elements on the right hand side, so we could introduce a question mark in the definition of ArticleFragmentSpec for the meta to make these new object literals redundant 👍

ajwl:
Thanks @jonathonherbert 🙏

PR #719, opened 2019-04-30
First pass at adding edition priority

RichieAHB:
## What's changed?

This gives us an `edition` priority that is always hidden. Not going to merge this yet as I want to do some more research on whether I need to add anything else here.

## Implementation notes

To defend against client errors in dev I've added a santisation on the server to ensure fronts with edition priority are always hidden despite what the server says. I've also applied this logic to training because, why not.

I'm not sure whether `350` fronts will be enough long term but we can change this if needs be and it's the value used by `editorial` / `commercial` priorities.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/util/SanitizeInput.scala
+++ b/app/util/SanitizeInput.scala
@@ -6,23 +6,12 @@ import scala.util.matching.Regex
 
 object SanitizeInput {
   val sanitizeRegex: Regex = "(<.*?>|<[^>]*$)".r
-  val hiddenPriorities = List("edition", "training")
 
   def fromString(s: String) = sanitizeRegex.replaceAllIn(s, "")
 
-  def sanitizeConfig(config: ConfigJson): ConfigJson = {
-    runSanitize(config, sanitizeSeoInputFromFront, ensureHiddenPriorities)
+  def fromConfigSeo(config: ConfigJson): ConfigJson = {
+    config.copy(fronts = config.fronts.mapValues(sanitizeSeoInputFromFront))
   }
-
-  private def runSanitize(config: ConfigJson, fns: FrontJson => FrontJson*) = {
-    config.copy(fronts = config.fronts.mapValues(front => fns.foldLeft(front)((f, fn) => fn(f))))
-  }
-
-  private def ensureHiddenPriorities(front: FrontJson): FrontJson = front.copy(
-    isHidden = front.priority.flatMap(p =>
-      Some(true).filter(_ => hiddenPriorities.contains(p)).orElse(front.isHidden)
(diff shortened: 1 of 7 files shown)


app/util/SanitizeInput.scala
```diff
@@ -6,13 +6,24 @@ import scala.util.matching.Regex
...
 object SanitizeInput {
   val sanitizeRegex: Regex = "(<.*?>|<[^>]*$)".r
+  val hiddenPriorities = List("edition", "training")
 
   def fromString(s: String) = sanitizeRegex.replaceAllIn(s, "")
 
-  def fromConfigSeo(config: ConfigJson): ConfigJson = {
-    config.copy(fronts = config.fronts.mapValues(sanitizeSeoInputFromFront))
+  def sanitizeConfig(config: ConfigJson): ConfigJson = {
+    runSanitize(config, sanitizeSeoInputFromFront, ensureHiddenPriorities)
```
Reettaphant: why have we added `ensureHiddenPriorities` function here?

Reettaphant: I've just read your implementation notes! I'm not sure if we need this extra check as we've made the editions work just like training fronts which have stayed well hidden for at least as long as I've been here. Was there some specific use case you wanted to guard against?

public/src/js/constants/defaults.js
```diff
@@ -139,6 +139,12 @@ export default {
             hasGroups: false,
             isTypeLocked: false,
             isHiddenLocked: false
+        },
+        'edition': {
+            maxFronts: 350,
+            hasGroups: true,
```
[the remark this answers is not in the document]
RichieAHB: Well, I don't know if it's true - good point. But Jon and I briefly discussed it and assumed it had to be but maybe I can ask @blishen ... will we have groups ("huge", "very big") etc?

app/util/SanitizeInput.scala
RichieAHB: The edge case was that while moving quickly we might forget to add this on the client, and as a result I decided to just add in `training` as there seemed no reason not to. It can be removed, it just seemed safer while we were developing.

Reettaphant: If we are not going to be developing on these as our actual editions might as well remove this.

public/src/js/constants/defaults.js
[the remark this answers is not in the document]
Reettaphant: David says probably not, not needed in our mock data at least!

PR #720, opened 2019-05-01
Fix sorting groups

Reettaphant:
## What's changed?

Collections layouts get switched frequently and the number of groups in a collection can change 
### Example: we can go from having 4 groups to having 2 groups to having 4 four groups again. 
- If we switch from 4 groups to 2 groups we should display all the articles in the old top two groups in the new top group. In addition, we need to keep the information about the groups these articles originally belonged to so that when we switch back to the layout containing four groups, all the stories are displayed in their original positions. To achieve this, I have changed the reducer logic so that all old/current groups are stored in state and selector logic so that articles belonging to groups which don't currently exist in config still get displayed in the current top group. 

- In addition, when we move articles around the indices we get for these moves are indices and ids of groups in the selected state and we need to adjust these indices so that they have the correct values with respect to our stored state. We don't allow dragging articles to positions which are occupied by articles belonging to groups which don't currently exist in config. This might be slightly buggy behaviour but the same behaviour exists in the old tool, no one has ever complained and Mariana is happy with this. 

I've tried to keep the code as simple as possible and provide comments were necessary, please let me know if it's not clear. 

## Implementation notes

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/actions/ArticleFragments.ts
+++ b/client-v2/src/actions/ArticleFragments.ts
@@ -18,6 +18,10 @@ import {
 import { ThunkResult, Dispatch } from 'types/Store';
 import { addPersistMetaToAction } from 'util/storeMiddleware';
 import { cloneFragment } from 'shared/util/articleFragment';
+import {
+  getFromGroupIndecesWithRespectToState,
+  getToGroupIndecesWithRespectToState
+} from 'util/moveUtils';
 import { PosSpec } from 'lib/dnd';
 import { Action } from 'types/Action';
 import {
@@ -273,29 +277,43 @@ const moveArticleFragment = (
       return;
     }
 
-    // if from is not null then assume we're copying a moved article fragment
-    // into this new position
-    const { parent, supporting } = !from
-      ? cloneFragment(
-          fragment,
-          articleFragmentsSelector(selectSharedState(getState()))
-        )
-      : { parent: fragment, supporting: [] };
+    const sharedState = selectSharedState(getState());
+    const fromDetails: {
+      fromWithRespectToState: PosSpec | null;
+      fromOrphanedGroup: boolean;
+    } = getFromGroupIndecesWithRespectToState(from, sharedState);
 
-    if (!from) {
-      dispatch(articleFragmentsReceived([parent, ...supporting]));
-    }
-
-    dispatch(
-      insertActionCreator(
-        to.id,
-        to.index,
-        parent.uuid,
-        from && removeActionCreator
-          ? removeActionCreator(from.id, fragment.uuid)
-          : null
-      )
+    const toWithRespectToState: PosSpec | null = getToGroupIndecesWithRespectToState(
+      to,
+      sharedState,
+      fromDetails.fromOrphanedGroup
     );
+    if (toWithRespectToState) {
+      const { fromWithRespectToState } = fromDetails;
+
+      // if from is not null then assume we're copying a moved article fragment
+      // into this new position
+      const { parent, supporting } = !fromWithRespectToState
+        ? cloneFragment(fragment, articleFragmentsSelector(sharedState))
+        : { parent: fragment, supporting: [] };
+
+      if (toWithRespectToState) {
+        if (!fromWithRespectToState) {
+          dispatch(articleFragmentsReceived([parent, ...supporting]));
+        }
+
+        dispatch(
+          insertActionCreator(
+            toWithRespectToState.id,
+            toWithRespectToState.index,
+            parent.uuid,
+            fromWithRespectToState && removeActionCreator
+              ? removeActionCreator(fromWithRespectToState.id, fragment.uuid)
+              : null
+          )
+        );
+      }
+    }
   };
 };
 
--- a/client-v2/src/actions/ArticleFragments.ts
+++ b/client-v2/src/actions/ArticleFragments.ts
@@ -13,7 +13,8 @@ import { ArticleFragment } from 'shared/types/Collection';
 import {
   selectSharedState,
   articleFragmentsSelector,
-  articleFragmentSelector
+  articleFragmentSelector,
+  articleGroupSelector
 } from 'shared/selectors/shared';
 import { ThunkResult, Dispatch } from 'types/Store';
 import { addPersistMetaToAction } from 'util/storeMiddleware';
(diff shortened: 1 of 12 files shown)


client-v2/src/actions/ArticleFragments.ts
```diff
@@ -16,6 +16,10 @@ import {
 import { ThunkResult, Dispatch } from 'types/Store';
 import { addPersistMetaToAction } from 'util/storeMiddleware';
 import { cloneFragment } from 'shared/util/articleFragment';
+import {
+  getFromGroupIndecesWithRespectToState,
```
LATaylor-guardian: `s/indeces/indices/` ?

client-v2/src/actions/ArticleFragments.ts
```diff
@@ -16,6 +16,10 @@ import {
...
+  getToGroupIndecesWithRespectToState
```
LATaylor-guardian: As above?

client-v2/src/shared/util/__tests__/shared.spec.ts
```diff
@@ -147,12 +148,36 @@ describe('Shared utilities', () => {
...
         groups: undefined
       });
-      const groupId = result.normalisedCollection.live![0];
-      expect(result.groups[groupId].articleFragments).toHaveLength(3);
+      const groupId1 = result.normalisedCollection.live![0];
+      expect(result.groups[groupId1].articleFragments).toHaveLength(1);
+      const groupId2 = result.normalisedCollection.live![1];
+      expect(result.groups[groupId2].articleFragments).toHaveLength(2);
+    });
+    it("should create empty groups for groups in the config which don't have collections in them", () => {
```
jonathonherbert: Do you mean `collections` here, or collectionItems/articleFragments?

client-v2/src/shared/util/shared.ts
```diff
@@ -26,38 +26,74 @@ const createGroup = (
...
+    }
+    return { ...group, name };
+  });
+
+  // We may have empty groups in the config which would not show up in the normalised
+  // groups result. We need to add these into the groups array.
+  if (collectionConfig.groups) {
+    collectionConfig.groups.forEach((group, index) => {
+      if (
+        !groupsWithNames.some(addedGroup => {
```
jonathonherbert: Could this live in a little helper that explains its purpose? We're a few indentations deep and it might make things a bit clearer -- but it might be just because it's half five and I should be on my bike already!

Reettaphant: Yes, excellent idea. I think it's much clearer now. Let me know what you think

client-v2/src/shared/util/__tests__/shared.spec.ts
Reettaphant: articleFragments... Good catch

client-v2/src/shared/util/shared.ts
```diff
@@ -26,38 +26,78 @@ const createGroup = (
...
+  }
+
+  // If we have no article fragments and no groups in a collection we still need to create
+  // and empty group for articles.
+  if (groupsWithNames.length === 0) {
+    groupsWithNames = [createGroup(null, null, getAllArticleFragments(groups))];
+  }
+
+  // Finally we need to sort the groups according to their ids.
+  groupsWithNames.sort((g1, g2) => {
```
jonathonherbert: Only a suggestion, but you could remove ten+ lines in a long-ish function by using lodash's sortBy, which accepts a predicate

client-v2/src/shared/util/shared.ts
```diff
@@ -26,38 +26,78 @@ const createGroup = (
...
+  });
 
   return {
-    addedGroups: keyBy(addedGroups, getUUID),
-    groupIds: addedGroups
-      .map(getUUID)
-      .slice()
-      .reverse()
+    addedGroups: keyBy(groupsWithNames, getUUID),
+    groupIds: groupsWithNames.map(getUUID).slice()
```
jonathonherbert: Is the slice to copy the array? map() has already given you a new one 😄

client-v2/src/shared/util/shared.ts
```diff
@@ -26,38 +26,78 @@ const createGroup = (
...
+const addGroupsForStage = (
   groupIds: string[],
   entities: { [id: string]: Group },
   collectionConfig: CollectionConfig
 ) => {
   const groups = groupIds.map(id => entities[id]);
-  const addedGroups = collectionConfig.groups
-    ? collectionConfig.groups.map((name, i) => {
-        const existingGroup = groupByIndex(groups, i);
+  let groupsWithNames = groups.map(group => {
```
jonathonherbert: Super minor: do we need a `let` here? As far as I can see the only reassignment occurs on 79, and this reassignment could be a `push` as it only occurs when `groupsWithNames` is empty.

client-v2/src/shared/selectors/shared.ts
```diff
@@ -172,8 +172,35 @@ const createCollectionStageGroupsSelector = () => {
...
+      // groups and display them in the top group.
+      const orphanedFragments: string[] = grps
+        .filter(grp => !grp.name && grp.id)
+        .reduce((frags: string[], g) => frags.concat(g.articleFragments), []);
+
+      // The final array of groups consist of groups where all groups without names but with ids
+      // are filtered out as these groups no longer exist in the config of the collection.
+      const finalGroups = grps.filter(grp => grp.name || !grp.id);
+      if (finalGroups.length > 0) {
+        const grpF = finalGroups ? finalGroups[0].articleFragments : [];
```
jonathonherbert: We already know that `finalGroups` exists and has a length -- is this ternary necessary?

RichieAHB:
I’ve just looked at the tests for this and I feel like, unless I’m missing something, we could do this in a simpler way ... I may be missing something though and this looks great if not. I’ve not reviewed this properly but would be good to talk about when I’m in.

client-v2/src/shared/selectors/shared.ts
jonathonherbert: (also just to satisfy my own curiosity does grpF mean first group? 🤓 )

client-v2/src/shared/selectors/shared.ts
```diff
@@ -172,8 +172,35 @@ const createCollectionStageGroupsSelector = () => {
...
-      ((collection && collection[stage]) || []).map(id => groups[id])
+    ): Group[] => {
+      const grps = ((collection && collection[stage]) || []).map(
+        id => groups[id]
+      );
+      if (grps.length < 2) {
+        return grps;
+      }
+
+      // Groups without names and ids are groups which no longer exist in the config because
```
jonathonherbert: Praise spam: these comments are 💯

jonathonherbert:
re: @RichieAHB you may well have seen forest for trees -- let's all chat when you're in!

RichieAHB:
I may also have just imagine a forest amid the trees 🙄

client-v2/src/shared/selectors/shared.ts
Reettaphant: Variables really shouldn't be called grpF - I have renamed to `originalFirstGroupArticleFragments` to be more descriptive.

jonathonherbert:
Per another conversation, the extra commits look fine, well spotted 👍

PR #723, opened 2019-05-01
Add a drop level container

jonathonherbert:
## What's changed?

Adds a drop level container, which lets us avoid having to depend on the drop root to listen for dragover events. This means a) we're not listening in two places, and b) we can avoid passing in a blacklist to the root element -- important for #724.

## Implementation notes

@RichieAHB -- I wasn't quite sure how to write a test for this -- be interested in your thoughts!

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/Clipboard.tsx
+++ b/client-v2/src/components/Clipboard.tsx
@@ -16,7 +16,7 @@ import {
   editorOpenClipboard,
   editorCloseClipboard
 } from 'bundles/frontsUIBundle';
-import { clipboardId, collectionDropZoneBlacklist } from 'constants/fronts';
+import { clipboardId } from 'constants/fronts';
 import {
   ArticleFragment as TArticleFragment,
   ArticleFragmentMeta,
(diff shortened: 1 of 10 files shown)


RichieAHB:
So to be clear, this is a "union" of all types that have been added to the `Levels` below? As it is I think this is fine, I wonder if we might want to drop the two sources of truth and have Levels pass this info back up ... not sure that's really an issue for now.

As for tests it should be doable - might be worth pairing on this tomorrow? We can setup a Root with a Level child, simulate bubbling of a fake dragOver event from the Level up through the Root and check the drop zones aren't rendering when that event is blacklisted?

jonathonherbert:
I think if anything Root should pass this state down rather than accrue it from its children -- but we can imagine a situation where different Levels have different drop requirements. So possibly we'd combine the base-level Root blacklist with component-specific blacklists in child Levels? Not sure, but yes, let's chat tomorrow. Re: tests, yes that sounds sensible, thanks!

RichieAHB:
Sounds good - I agree if we can avoid passing up that would be good. I just thought maybe Root needed to definitively ignore anything that was seen in a blacklist below but I think it’s actually more nuanced than that. It may be that this approach is actually fine as it is.

PR #726, opened 2019-05-02
VERTICAL COMPRESSION: remove trail text 

ajwl:
## What's changed?
Design have asked us to remove the trail text in the collection pane. When images are shrunk in size this will reduce the vertical height of stories. 

BEFORE 
Screenshot 2019-05-02 at 12 35 06

AFTER 
Screenshot 2019-05-02 at 12 33 02

## Implementation notes
I could have removed all the code around trail text... but I have opted not to because we have built an on-off switch for it which could be useful in future designs if we want to have a more expanded view. 

Please let me know if you disagree / think we should clear it out. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


Reettaphant:
I'd be comfortable with just removing all of the code around the trail because everything is in git anyway if we need to refer to it later.

ajwl:
@Reettaphant yes sure can do

ajwl:
@Reettaphant hello. I believe this removes all trailText code related to their display... 
Mentions of the trail text still remain in the edit form and types around CAPI which we will want to keep I imagine. 
have I missed anything else that could be deleted?

ajwl:
Great. Will just merge this to keep it separate from the image changes.. thanks

PR #733, opened 2019-05-10
Fix deleting sublinks

Reettaphant:
## What's changed?

I broke deleting sublinks, I fixed deleting sublinks. We can't select a group if we are deleting a sublinks and we need to handle this case correctly.

## Implementation notes

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


jonathonherbert:
Oh no, good to see a fix! Is it possible to test this behaviour via either unit/integration? We've introduced a few regressions recently and it'd be good to catch them next time.

Reettaphant:
I think for finding these kinds fo bugs integration tests testing removing, adding and moving articles from different positions would be very useful for catching these kinds of issues and we should do them before news desk rollout so that we stop breaking move beahviour. I think it's out of scope for this pr though.

PR #734, opened 2019-05-10
Save form values that have been changed to false

jonathonherbert:
## What's changed?

Save form values that have been changed to false. This allows us to revert truthy defaults on articles. (I'm not sure how these defaults are set -- @Reettaphant @RichieAHB any ideas?)

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


jonathonherbert:
@Reettaphant and I spoke and the above shouldn't be an issue; although the behaviour of the current tool differs from the old it should be within the scope of the model. The bug before was caused by saving 'undefined' as a string value against the isBreaking property.

If this does cause a problem, we can always inspect the article meta defaults and conditionally save properties when they differ from those defaults.

PR #739, opened 2019-05-15
First pass at responsive

RichieAHB:
## What's changed?

First pass at making the edit page responsive. I tried to get the percentages to be roughly aligned with @akemitakagi 's percentages. This is a WIP that is up for review! I've also added this to CODE to play around with (at least at the time of writing). When clicking around on CODE, try zooming out to give a rough example of what it might look like on an iMac.

ezgif com-video-to-gif (2)

## Implementation notes

Worth noting that I haven't closed the clipboard and overviews when 3 fronts are added, this feels like it might be a bit of an imposition but it's very doable. For now I've just changed the CSS. Also, 3 fronts don't fit neatly into a 15" Macbook screen and do overflow - it was starting to feel a bit too tight. Have a play around and see.

Additionally, I don't think (or at least I can only see it being quite fragile) we can make the Overview / Clipboard anything other than fixed size. A percentage size using the current method means that collapsing them just using CSS would be hard / impossible. I'm not sure this is a problem but worthy of note and if it's a real issue there are other methods we can consider here.

Some of the considerations that went into this layout:
- Fronts + clipboard always take up at least 100% of the page width.
- Clipboard / overview collapse properly
- Meta editor opening expands a front without squashing the collections too much
- Collections shrink at a slower pace than the feed when the page narrows so as to try and keep vertical height to a minimum in collections

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/bundles/frontsUIBundle.ts
+++ b/client-v2/src/bundles/frontsUIBundle.ts
@@ -305,11 +305,9 @@ const selectEditorFavouriteFrontIdsByPriority = (
 ): string[] =>
   state.editor.favouriteFrontIdsByPriority[priority] || defaultFavouriteFronts;
 
-const selectHasMultipleFrontsOpen = createSelector(
+const selectOpenFrontsCount = createSelector(
   selectEditorFrontIdsByPriority,
-  frontIdsByPriority => {
-    return frontIdsByPriority.length > 1;
-  }
+  frontIdsByPriority => frontIdsByPriority.length
 );
 
 const selectEditorArticleFragment = <T extends { editor: State }>(
@@ -596,7 +594,7 @@ export {
   editorCloseAllOverviews,
   selectIsClipboardOpen,
   selectIsFrontOverviewOpen,
-  selectHasMultipleFrontsOpen
+  selectOpenFrontsCount
 };
 
 export default reducer;
--- a/client-v2/src/components/FeedSection.tsx
+++ b/client-v2/src/components/FeedSection.tsx
@@ -9,15 +9,18 @@ import FeedSectionHeader from './FeedSectionHeader';
 
 const FeedSectionContainer = styled('div')`
   background-color: ${({ theme }) => theme.shared.base.colors.backgroundColor};
+  width: 100%;
 `;
 
 const FeedWrapper = styled('div')`
-  width: 409px;
+  flex: 3 1 20vw;
   padding-right: 10px;
   margin-right: 10px;
   border-right: ${({ theme }) =>
     `solid 1px ${theme.shared.base.colors.borderColor}`};
 `;
+
+const ClipboardWrapper = styled.div``;
 
 export default () => (
   <FeedSectionContainer>
@@ -26,10 +29,10 @@ export default () => (
       <FeedWrapper>
         <FeedContainer />
       </FeedWrapper>
-      <Clipboard />
-      <div>
-        <ClipboardMeta />
-      </div>
+      <ClipboardWrapper>
+        <Clipboard />
+      </ClipboardWrapper>
+      <ClipboardMeta />
     </SectionContent>
   </FeedSectionContainer>
 );
--- a/client-v2/src/components/FeedSection.tsx
+++ b/client-v2/src/components/FeedSection.tsx
@@ -22,12 +22,12 @@ const FeedWrapper = styled('div')`
 
 const ClipboardWrapper = styled.div``;
 
-export default () => (
+export default ({ fontSize = '14px' }: { fontSize?: string }) => (
   <FeedSectionContainer>
     <FeedSectionHeader />
     <SectionContent>
       <FeedWrapper>
-        <FeedContainer />
+        <FeedContainer fontSize={fontSize} />
       </FeedWrapper>
       <ClipboardWrapper>
         <Clipboard />
--- a/client-v2/src/components/FrontsEdit/CollectionOverview.tsx
+++ b/client-v2/src/components/FrontsEdit/CollectionOverview.tsx
@@ -61,6 +61,9 @@ const Container = styled.div`
 const TextContainerLeft = styled.div`
   flex: 1 1;
   font-size: 14px;
+  overflow-wrap: break-word;
+  word-wrap: break-word;
+  hyphens: auto;
 `;
 
 const TextContainerRight = styled.div`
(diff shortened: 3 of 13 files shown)


client-v2/src/components/FrontsEdit/CollectionOverview.tsx
```diff
@@ -61,6 +61,9 @@ const Container = styled.div`
 const TextContainerLeft = styled.div`
   flex: 1 1;
   font-size: 14px;
+  overflow-wrap: break-word;
```
RichieAHB: This fixes the overflowing exclamation marks in the overview for collections with long names

client-v2/src/components/FrontsEdit/Edit.tsx
```diff
@@ -81,11 +97,21 @@ class FrontsEdit extends React.Component<Props> {
...
+            {this.props.frontIds.length ? (
+              this.props.frontIds.map(id => (
+                <SingleFrontContainer key={id} id={createFrontId(id)}>
+                  <FrontContainer frontId={id} />
+                </SingleFrontContainer>
+              ))
+            ) : (
+              <NoFrontContainer>
+                <span>
+                  Select a front with the{' '}
```
RichieAHB: I added some dummy text to the placeholder here @akemitakagi - feel free to remove / change it but given I had to add an element for when we have no fronts I thought I'd add a bit of sign posting!

Screenshot 2019-05-15 at 18 29 48

RichieAHB:
I've also added this to CODE to play around with @akemitakagi @harpreetpurewal @jonathonherbert @Reettaphant @ajwl

client-v2/src/bundles/frontsUIBundle.ts
```diff
@@ -596,7 +594,7 @@ export {
   editorCloseAllOverviews,
   selectIsClipboardOpen,
   selectIsFrontOverviewOpen,
-  selectHasMultipleFrontsOpen
+  selectOpenFrontsCount
```
jonathonherbert: this _was_ oddly specific

RichieAHB: I kinda liked it but either way - IT DID NOT SUIT MY NEEDS

client-v2/src/components/FrontsEdit/Edit.tsx
```diff
@@ -42,16 +45,32 @@ const FrontsEditContainer = styled('div')`
...
 
 const SingleFrontContainer = styled('div')`
+  flex: 1 1 auto;
   height: 100%;
+  min-width: 626px;
+`;
+
+// This is just to stop the feed / clipboard from filling the screen when no fronts
+// are selected
+const NoFrontContainer = styled.div`
```
jonathonherbert: Would limiting the maximum width of the feed/clipboard achieve the same thing without markup? (I've probably missed something!)

client-v2/src/components/FrontsEdit/Edit.tsx
jonathonherbert: mega 👍

client-v2/src/components/FrontsEdit/Edit.tsx
```diff
@@ -117,7 +146,11 @@ const mapStateToProps = (state: State, props: Props) => ({
     state,
     props.match.params.priority || ''
   ),
-  isCurrentFrontsMenuOpen: selectIsCurrentFrontsMenuOpen(state)
+  isCurrentFrontsMenuOpen: selectIsCurrentFrontsMenuOpen(state),
+  feedFontSize:
+    selectOpenFrontsCount(state, props.match.params.priority) > 1
```
jonathonherbert: Is it worth keeping the idea of 'size' abstract and having individual components interpret it? a feedSize of 's', 'm', 'l' could be used by other components if we wanted to make other changes based on front count etc.

jonathonherbert:
Looks great, a few minor comments that aren't blockers, with one suggestion below:

The only thing I'd say is that the maximum width of the feed looks very large -- the intent being that with one front open the whole viewport is filled? I'd argue that it's fine to have a little space present on the RHS and we don't need to fill it; as such we can limit the width of the feed and single fronts so they're not quite so stretched.

Why not stretch them? Good point! I'd argue that there's unnecessary mouse travel for users as a result (is one front what most casual users will see?) -- the modification buttons for feed items and collection items are on the left, and the form operations are on the right. (Arguably dragging could be a part of this, too, but I suppose the gap between the feed and the front is the same size it ever was).

Anyway, consider this a weak opinion weakly held, but I thought I'd mention it!

client-v2/src/components/FrontsEdit/Edit.tsx
RichieAHB: I did try setting max-width here to 50% and I can't remember what broke ... I'm going to sit here and try and remember.

*time elapses*

Figured it ... it's because we cant' guarantee what we actually want the max-width to be. On sufficiently narrow screens it's over 50% of the viewport but actually we rely on the content on the right to squash us down to (hopefully) 50% when a front pops in thanks to the flex layout.

client-v2/src/components/FeedSection.tsx
```diff
@@ -9,27 +9,30 @@ import FeedSectionHeader from './FeedSectionHeader';
 
 const FeedSectionContainer = styled('div')`
   background-color: ${({ theme }) => theme.shared.base.colors.backgroundColor};
+  width: 100%;
 `;
 
 const FeedWrapper = styled('div')`
-  width: 409px;
+  flex: 3 1 20vw;
```
ajwl: ooh `vw` units. fancy!

client-v2/src/components/FrontsEdit/Edit.tsx
RichieAHB: I did wonder that but it's such a small deviation that I thought `s` and `m` didn't cover it but happy to revisit it, if you think that makes sense!

client-v2/src/components/FeedSection.tsx
RichieAHB: I'm not sure what prompted such a fancy unit - I think it's the first time I've ever used it 👌

RichieAHB:
@jonathonherbert to your point I've added a rather large max width on both the feed and front of 1000px. @akemitakagi let me know if you'd suggest any alterations to this!

PR #741, opened 2019-05-16
Rv metadata overrides

Reettaphant:
## What's changed?

Add the missing metadata overrides. Large headline seems to be the same as boosted headline, just goes by different names, so I have just made the image larger for headlines. Added all three image overrides in and they are now written down in text as it is impossible communicate by images 

See below a large headline and all metadata overrides. 

Screenshot 2019-05-16 at 10 35 41

## Implementation notes

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


RichieAHB:
What happens here when there are multiple pieces of meta to display - it looks like they will stack vertically, is that right?

Reettaphant:
We can only have one the imagemetas selected at a time because it doesn't make sense to for example have image replace and slideshow on at the same time. I made a pr some time ago to enforce this.

RichieAHB:
In which case ...

PR #742, opened 2019-05-17
Drag images between CollectionItem overrides and image inputs

jonathonherbert:
## What's changed?

Drag image overrides and input images _literally anywhere you want_.

1502812622-brian-butterfield

This PR lets you drag overrides from collectionItems into image inputs and vice versa, and between image inputs.

dragimg

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


client-v2/src/components/FrontsEdit/Collection.tsx
```diff
@@ -11,12 +11,8 @@ import GroupLevel from 'components/clipboard/GroupLevel';
 import CollectionItem from './CollectionComponents/CollectionItem';
 import ArticleFragmentLevel from 'components/clipboard/ArticleFragmentLevel';
 import { PosSpec, Move } from 'lib/dnd';
-import { ValidationResponse } from 'shared/util/validateImageSrc';
 import { Dispatch } from 'types/Store';
```
jonathonherbert: This is better colocated with the CollectionItem 👍

client-v2/src/shared/components/input/InputImage.tsx
```diff
@@ -130,12 +229,12 @@ class InputImage extends React.Component<ComponentProps, ComponentState> {
       });
   };
 
-  public handlePasteImgSrcChange = (e: React.FormEvent<HTMLInputElement>) => {
+  private handlePasteImgSrcChange = (e: React.FormEvent<HTMLInputElement>) => {
```
jonathonherbert: I humbly submit that methods should be private by default! Happy to discuss.

client-v2/src/shared/components/input/InputImage.tsx
```diff
@@ -110,17 +114,112 @@ interface ComponentState {
...
     isHovering: false,
     modalOpen: false,
     imageSrc: ''
   };
 
-  public handleDrop = (e: React.DragEvent<HTMLDivElement>) => {
+  public render() {
+    const { small, input, gridUrl } = this.props;
+    const gridSearchUrl = `${gridUrl}?cropType=landscape`;
+    const hasImage = !!input.value && !!input.value.thumb;
```
ajwl: Small thing, but why the double negative? `!!`

ajwl:
From a UI perspective, is it the case that I can only drag from the Grid? And if that’s so, could we make that clearer? 

I was trying to drag images between stories (not a likely use case I’m sure) but I see that didn’t work. A question for design probably @akemitakagi - but in a future iteration could there be some indication that a picture is available to be moved.

client-v2/src/shared/components/input/InputImage.tsx
jonathonherbert: Good question! The double negative coerces the expression that follows it into a boolean type. I think it's fair to say that this is idiomatic in Javascript. In this case it ensures that `hasImage` is of a boolean type. Because of short-circuit evaluation, if the double negatives weren't there, 
- if input.value was falsey, hasImage would be the value of input.value, which could be _anything falsey_ (`undefined`, `null`, `false`, `NaN`, maybe I've missed a few!)
- if input.value was truthy but input.value.thumb was not, hasImage would be the value of input.value.
- if input.value and input.value.thumb were both truthy, hasImage would be the value of input.value.thumb.

ajwl: thanks for explanation!

jonathonherbert:
Me and @ajwl had a conversation to clarify that this PR doesn't cover dragging images in collectionItems that _don't_ have image overrides. I've updated the PR description to reflect that.

@ajwl after is there anything else needed to move this PR forward?

PR #747, opened 2019-05-22
TESTS: New integration tests

ajwl:
## What's changed?
This adds 6 new paths to the integration tests:

1) Drag article between two collections 
2) Drag article from clipboard to collection + check that article remains in the clipboard 
3) Check that the delete button on articles works ❌ _(moved to separate PR hover state not working)_
4) Metadata edits (Headline change / Breaking news toggle button)
5) Discard collection (this means copying what is in a collections `live` property to its `draft` property. 
6) Add to clipboard button on individual stories works ❌ _(moved to separate PR hover state not working)_

## Implementation notes

* Adding a 2nd collection has made it necessary to create more complex functions in `selectors/index.js`. Now we need to select  the collection as well as the item within that collection.

* This breaks some of the nicer code we had wrapping this - open to ways of improving, if you have any thoughts

### Problems with hover state

We have had significant problems with the `hover state` (used to trigger the delete buttons on stories for example). In many runs of this code, the "test mouse" hasn't triggered the hover state, where a "real mouse" has. 

This appears to be a problem when triggering hover on styled components - but doesn't have a fix
https://github.com/DevExpress/testcafe/issues/3830

--> have moved [integration tests involving hover states to a separate PR](https://github.com/guardian/facia-tool/pull/760)

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### DIFF
--- a/client-v2/integration/selectors/index.js
+++ b/client-v2/integration/selectors/index.js
@@ -92,6 +92,14 @@ export const collectionItemDeleteButton = (collectionIndex, itemIndex = 0) =>
     `[data-testid="${DELETE_BUTTON}"]`
   );
 
+export const collectionItemAddToClipboardButton = (
+  collectionIndex,
+  itemIndex = 0
+) =>
+  collectionItem(collectionIndex, itemIndex).find(
+    `[data-testid="${ADD_TO_CLIPBOARD_BUTTON}"]`
+  );
+
 // Edit form //
 export const editForm = maybeGetNth(select(EDIT_FORM));
 
--- a/client-v2/integration/selectors/index.js
+++ b/client-v2/integration/selectors/index.js
@@ -1,5 +1,4 @@
 import { select } from '../helpers';
-import { Selector } from 'testcafe';
 
 const FRONT_SELECTOR = 'test/front';
 const FEED_ITEM_SELECTOR = 'feed-item';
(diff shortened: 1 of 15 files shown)


client-v2/integration/server/server.js
```diff
@@ -110,25 +112,45 @@ module.exports = async () =>
...
     };
 
-    // Attempts at a capture group:
-    // /api/(preview|live)/*
-    // /api/(?:preview|live)/
     app.get('/api/live/*', handler);
     app.get('/api/preview/*', handler);
 
     app.get('/config', (_, res) => res.json(config));
-    app.get('/collection/:id', (_, res) => res.json(collection));
```
ajwl: This GET route was not being used - we actually use a 'POST` request to handle requests for collections.

client-v2/src/shared/components/input/InputCheckboxToggle.tsx
```diff
@@ -66,16 +66,18 @@ const Checkbox = styled('input')`
 type Props = {
   label?: string;
   id: string;
+  dataTestId: string;
```
ajwl: In order to push the `dataTestId` down to the right place I needed to put this in the props and send it down to the `InputContainer`.

client-v2/integration/selectors/index.js
```diff
@@ -34,12 +43,67 @@ export const clipboardDropZone = maybeGetNth(
...
+
+export const collectionItem = (collectionIndex, itemIndex = 0) =>
+  allCollectionItems(collectionIndex).nth(itemIndex);
+
+export const collectionDropZone = (collectionIndex, dropZoneIndex = 0) =>
+  allCollectionDropZones(collectionIndex).nth(dropZoneIndex);
+
+export const collectionItemHeadline = (collectionIndex, itemIndex = 0) =>
+  collectionItem(collectionIndex, itemIndex).find(
+    `[data-testid="${HEADLINE_SELECTOR}"]`
```
ajwl: This pattern of using the `[data-testid="${HEADLINE_SELECTOR}"]` doesn't use the clean `select` helper method that is defined here: https://github.com/guardian/facia-tool/blob/master/client-v2/integration/helpers/index.js

ajwl:
NB: Will also add a test for the Add To Clipboard Button

client-v2/integration/server/server.js
jonathonherbert: 👍

PR #749, opened 2019-05-24
Rv collapse completely

Reettaphant:
## What's changed?

Clipboard and front overviews will now collapse completely as per designs. 

## Implementation notes

The clipboard header is now passed to the feed container as a right hand component so that it can be placed to the left of the clipboard. 

The collapse overview and collapse collection controls have now also moved to the fronts component so we can display then nicely on top of the fronts content. 

Made some other changes as well to margins/paddings but I think the layout makes more sense now and the components which should be in in control or margins around them are now doing so. 

Screenshot 2019-05-31 at 08 42 59

Screenshot 2019-05-31 at 08 27 03

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/Clipboard.tsx
+++ b/client-v2/src/components/Clipboard.tsx
@@ -13,8 +13,6 @@ import {
   editorSelectArticleFragment,
   editorClearArticleFragmentSelection,
   selectIsClipboardOpen,
-  editorOpenClipboard,
-  editorCloseClipboard
 } from 'bundles/frontsUIBundle';
 import { clipboardId } from 'constants/fronts';
 import {
@@ -26,7 +24,6 @@ import ClipboardLevel from './clipboard/ClipboardLevel';
 import ArticleFragmentLevel from './clipboard/ArticleFragmentLevel';
 import CollectionItem from './FrontsEdit/CollectionComponents/CollectionItem';
 import { styled } from 'constants/theme';
-import ButtonCircularCaret from 'shared/components/input/ButtonCircularCaret';
 import DragIntentContainer from 'shared/components/DragIntentContainer';
 import {
   setFocusState,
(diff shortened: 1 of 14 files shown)


RichieAHB:
Just from the screenshots, have we lost some padding around between the clipboard and the first front or is it partly scrolled?

Reettaphant:
@RichieAHB Good catch, I can't remember why I removed it in the first place anymore! The collapse overview is still looks like it's not collapsing completely if the front name is too long because there is too much space in the header. I'm going to make a card to implement the designs to replace the header button text with icons to fix this.

client-v2/src/components/Clipboard.tsx
```diff
@@ -153,112 +133,99 @@ class Clipboard extends React.Component<ClipboardProps> {
...
-              <Root
-                id="clipboard"
-                data-testid="clipboard"
-                style={{ display: 'flex', flex: 1 }}
-              >
-                <ClipboardLevel
-                  onMove={this.handleMove}
-                  onDrop={this.handleInsert}
+      <React.Fragment>
+        {this.props.isClipboardOpen && (
```
jonathonherbert: Super minor and abstract point that probably isn't worth changing in this PR -- is this where we'd want to add this condition? It makes our clipboard component unusable elsewhere. As a rule of thumb I've found that if you're writing a conditional at the root of a component's render fn, it probably belongs somewhere further up the tree. I'll stop writing now ahhh

client-v2/src/components/FrontsCAPIInterface/SearchInput.tsx
```diff
@@ -112,21 +118,24 @@ class SearchInput extends React.Component<SearchInputProps, SearchInputState> {
...
+              searchTermsExist={this.searchTermsExist}
+              onDisplaySearchFilters={this.handleDisplaySearchFilters}
+              onKeyUp={e => {
+                if (e.keyCode === 13) {
+                  this.hideSearchFilters();
+                }
+              }}
+            />
+          </TextInputContainer>
+          {rightHandContainer}
```
jonathonherbert: 👍

jonathonherbert:
Merging this in @Reettaphant's absence, as discussed 👍

PR #752, opened 2019-05-24
DAILY EDITION: turn production data into to scala code

ajwl:
## What's changed?

Going from David's spreadsheet with a list of containers and fronts in it, this should model our Monday - Saturday fronts.

Remaining questions: 

-> Ordering of Fronts. Is this important here or will it be done elsewhere? Reordering not hard, but having different orders on different days would be. 

-> Collection Naming. Should we match the spreadsheet more closely with the `/`s etc? Believe these collection names are for internal use so maybe we should get rid of capitals and spaces. 

let me know thoughts

## Implementation notes

Tests - I deleted the old test fixture because we now have a real one. 

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


ajwl:
@Reettaphant gonna merge this now - @blishen will do the reordering in a separate PR

PR #754, opened 2019-05-28
Clipboard paste

Reettaphant:
## What's changed?

- Fixes a bug where clipboard paste was not working because we were passing the wrong argument to it after drag/drop refactor. 

- Changes the clipboard paste behaviour, we now can past onto the clipboard with command +v without focusing on the clipboard first. I've checked and this does affect behaviour to paste text to input fields.

## Implementation notes

@jonathonherbert Don't know if you have an opinion on a better/neater place where the new paste function could go. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


ajwl:
In order to user test this - what are the steps to replicate this journey? 
(I tried Cmd+C on a story and Cmd+V on Clipboard but this appeared not to work for me. Do I need to hit some keyboard setting or something?)

paperboyo:
This may only pertain to d&d copy/paste, but @ajwl brings attention to a lack of (useful) parity of v2: in v1 a simple cmd+v pastes a link to the clipboard regardless of current app focsu. This, sadly, doesn’t work in v2.

jonathonherbert:
@ajwl I think the behaviour relates to copying a link from elsewhere -- e.g. copy article link in browser, ctrl-v whilst focused on fronts tool, article appears in clipboard. Not to say the behaviour you are describing wouldn't be nice -- (although we do have a + arrow to add articles to clipboard as a workaround)

@paperboyo I think the behaviour you're describing is what's implemented in this PR -- `Changes the clipboard paste behaviour, we now can past onto the clipboard with command +v without focusing on the clipboard first`

paperboyo:
Bombastic! Always happy to learn I’m an idiot!

ajwl:
@jonathonherbert ah thanks for explanation. That does work with external links.

jonathonherbert:
@Reettaphant this looks good to me -- one thing to mention, the focus state for the clipboard component itself is now redundant as we don't require it to be focused for a paste operation. Would it be worth removing it so we don't mislead users by suggesting that it does something useful?

ajwl:
👍 works!

client-v2/src/keyboard/index.ts
```diff
@@ -63,7 +66,31 @@ export const createKeyboardActionMap = (store: Store): KeyboardBindingMap => ({
   'command+v': {
     title: 'Paste',
     description: 'Paste an entity',
-    action: bindActionMap(store, paste)
+    action: () =>
+      (store.dispatch as Dispatch)(async (dispatch: Dispatch) => {
```
jonathonherbert: Me and Anna have just gone through this, and the double parenthesis are a little confusing -- we came up with something along the lines of 
```
action: async () => {
      const dispatch: Dispatch = store.dispatch;
      try {
        if (!navigator || !(navigator as any).clipboard) {
          throw new Error('No navigator available on paste');
        }
        // A temporary any here pending proper typings
        const content: string = await (navigator as any).clipboard.readText();
        if (!content) {
          return;
        }
        const contentData: RefDrop = { type: 'REF', data: content };
        const articleFragment = await dispatch(
          createArticleFragment(contentData)
        );
        if (!articleFragment) {
          return;
        }
        dispatch(
          insertClipboardArticleFragment('clipboard', 0, articleFragment.uuid)
        );
      } catch (e) {
        Raven.captureMessage(`Paste to clipboard failed: ${e.message}`);
      }
```
-- what do you think?

Reettaphant: Yes, this is much better! I will change it.

Reettaphant:
@jonathonherbert We still need the clipboard focus state for moving articles up and down with arrow keys so I think it makes sense to keep it there.

PR #764, opened 2019-06-03
ICONS: make front container buttons icons

ajwl:
## What's changed?
Buttons that current have the words "preview" and "close" should be replaced with icons, according to this trello card. 

https://trello.com/c/JMmmhbLx/628-buttons-in-the-header-should-be-icons-as-in-the-designs

OLD
Screenshot 2019-06-03 at 17 05 41

NEW
Screenshot 2019-06-04 at 11 24 04

## Implementation notes

Also I added bold version italics
And replaced the font weight range with a single value - more readable / descriptive of what actually happens. 
-- last font work. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/components/FrontsEdit/FrontContainer.tsx
```diff
@@ -50,6 +51,11 @@ const FrontsContainer = styled('div')`
   transform: translate3d(0, 0, 0);
 `;
 
+const FrontHeaderButton = styled(Button)`
+  color: ${({ theme }) => theme.shared.colors.greyMediumDark};
+  padding: 0 5px;
```
ajwl: Is there a way I could avoid this with the `Button` properties?

jonathonherbert: @ajwl and I had a brief chat offline -- we could absolutely have a prop on the Button component for button content that required minimal padding, possibly in a future PR 👍

PR #769, opened 2019-06-05
Fix launch instructions

mbarton:
Globally installing jspm leads to JSPM v2 polluting the ~/.jspm cache leading to cryptic failures when running locally with SystemJS. This is doubly bad since it asks you to use sudo, which is not right when using nvm.

As such, we now mandate the use of NVM in the set up instructions and run jspm via npm, essentially fixing the version to the right one from package.json.

Grunt is only required locally when running tests so installation has been moved down.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,62 +54,11 @@ If you have more than one JDK version running, ensure you run JDK 8 by adding th
 export JAVA_HOME=`/usr/libexec/java_home -v 1.8
 ```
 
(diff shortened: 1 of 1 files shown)


RichieAHB:
Nice to add this back in to the README, we updated this to lock `jspm` in CI but I'm only sorry we didn't do this locally / in the README alread. Thanks 👍

PR #770, opened 2019-06-05
Keep the overview open when the article form closes

jonathonherbert:
## What's changed?

A side effect of #749 (that I added) changed the behaviour of the form toggles -- opening a form would close the overview for the front. This isn't desirable re: feedback from Mariana, so this PR reverts that behaviour: opening a form doesn't interfere with the overview state.
 
### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
+++ b/client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
@@ -73,7 +73,8 @@ const FormContainer = styled(ContentContainer.withComponent('form'))`
   min-width: ${formMinWidth}px;
   max-width: 380px;
   margin-left: 10px;
-  height: 100%;
+  margin-top: 10px;
+  height: calc(100% - 10px);
 `;
 
 const FormContent = styled('div')`
--- a/client-v2/src/bundles/__tests__/frontsUIBundle.spec.ts
+++ b/client-v2/src/bundles/__tests__/frontsUIBundle.spec.ts
@@ -393,8 +393,8 @@ describe('frontsUIBundle', () => {
       });
     });
     describe('Collection item form display', () => {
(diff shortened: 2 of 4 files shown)


RichieAHB:
So does the overview actually stay visible here too? My assumption here was that we'd want to hide the overview while the panel was open but it reappears when cancel was clicked (or more generally, that the editor is closed). Mariana did mention the cancel button in the email ... you may already have discussed this and now that I think about it, this solution is probably better. I just wondered whether this is definitely the required behaviour?

jonathonherbert:
No, the overview is hidden by the form when the form is present -- it's basically the old behaviour.

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
```diff
@@ -73,7 +73,8 @@ const FormContainer = styled(ContentContainer.withComponent('form'))`
   min-width: ${formMinWidth}px;
   max-width: 380px;
   margin-left: 10px;
-  height: 100%;
+  margin-top: 10px;
```
jonathonherbert: A sneaky spacing edit to stop the form abutting the fronts header.

RichieAHB:
@jonathonherbert - I get it now. I thought the closing of the overview was the only thing stopping the overview rendering but I see [here](https://github.com/guardian/facia-tool/blob/309c6ce75e0928e3a4ac72b3940ef81ee4c91d79/client-v2/src/components/FrontsEdit/FrontDetailView.tsx#L38-L56) that there is already a check to ensure we don't render the overview if we have a form open 👍

jonathonherbert:
@RichieAHB -- ah, yes, I could have made that clearer!

PR #772, opened 2019-06-05
Refactor breaking news client

mbarton:
## What's changed?

https://github.com/guardian/facia-tool/pull/744 unfortunately introduced a bug where Global breaking news notifications would appear to be sent and confirmed in the UI, even though no notifications were actually sent.

We tracked it down to a change in the `mobile-notifications` client from https://github.com/guardian/mobile-n10n/pull/410 which the Fronts tool uses to trigger the notifications. The code that attempted to prioritise the UK breaking news topic first on a Global send ended up filtering out all the topics and sending nothing. Unfortunately, the way the API was defined meant that the Fronts tool itself could not detect this and the failure was entirely swallowed by the library.

In response to this we have decided to continue using the data structures provided by the library but will make the HTTP requests directly from code in the Fronts tool repo. This PR also takes advantage of this to refactor the code that was handling errors, passing up an `Either` type and handling the errors as close as possible to the user. The intention is to clarify and reduce code to make bugs less likely.

## Implementation notes

...

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/updates/BreakingNewsUpdate.scala
+++ b/app/updates/BreakingNewsUpdate.scala
@@ -15,7 +15,7 @@ import play.api.libs.json.Json
 import play.api.libs.ws.{WSClient, WSResponse}
 import play.api.mvc.Result
 import play.api.mvc.Results.{InternalServerError, Ok}
-import util.{BreakingNewsClient, BreakingNewsRequest}
+import util.{BreakingNewsClient, BreakingNewsError, BreakingNewsRequest, BreakingNewsSuppressedError}
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
@@ -39,8 +39,8 @@ class BreakingNewsUpdate(val config: ApplicationConfiguration, val ws: WSClient,
     email: String
   ): Future[Result] = {
 
-    def structuredLog(update: UpdateMessage, level: String = "info", error: Option[String] = None) = {
-      structuredLogger.putLog(LogUpdate(update, email), level, error.map(new Exception(_)))
+    def structuredLog(update: UpdateMessage, level: String = "info", error: Option[BreakingNewsError] = None) = {
+      structuredLogger.putLog(LogUpdate(update, email), level, error.flatMap(BreakingNewsError.getThrowable))
     }
 
     structuredLog(HandlingBreakingNewsCollection(collectionId))
@@ -53,13 +53,12 @@ class BreakingNewsUpdate(val config: ApplicationConfiguration, val ws: WSClient,
       if(request.trail.alert.getOrElse(false)) {
         client.send(request).map(request -> _)
       } else {
-        val errorMessage = "There has been a problem sending the breaking news alert. Please contact central production before sending additional alerts"
-        structuredLog(SuppressedBreakingNewsSend(collectionId, request.trail, request.topic.name), level = "error", error = Some(errorMessage))
-
-        // TODO MRB: use our own custom error enum rather than the MAPI client
-        Future.successful(
-          request -> Left(UnexpectedApiResponseError(errorMessage))
+        val error = BreakingNewsSuppressedError(
+          "There has been a problem sending the breaking news alert. Please contact central production before sending additional alerts"
         )
+
+        structuredLog(HandlingBreakingNewsTrail(collectionId, request.trail, request.topic.name), level = "error", error = Some(error))
+        Future.successful(request -> Left(error))
       }
     }.map { results =>
       val successes = results.collect { case (request, Right(response)) => request -> response }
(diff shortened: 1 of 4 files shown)


app/updates/BreakingNewsUpdate.scala
```diff
@@ -39,55 +38,66 @@ class BreakingNewsUpdate(val config: ApplicationConfiguration, val ws: WSClient,
     collection: ClientHydratedCollection,
     email: String
   ): Future[Result] = {
+    // TODO MRB:
+    // check we still log in the correct places
+    // does putting it in the web container set the "alert" flag?
```
Reettaphant: Had a little dig around to figure out why we need the alert set to true and there is no good reason for it but we need to work around the fronts publish model. Once we've alerted on a story and removed the story we need to click the remove button for the live version of the breaking news collections to be updated, clicking this remove will post to the same publish route as sending a breaking news alert but because the alert flag is set to false we don't send any alerts on draft articles.

app/updates/BreakingNewsUpdate.scala
```diff
@@ -39,55 +38,66 @@ class BreakingNewsUpdate(val config: ApplicationConfiguration, val ws: WSClient,
...
-        case Success(futureMaybeError) => futureMaybeError
-        case Failure(t: Throwable) =>
-          val message = s"Exception in breaking news client send for trail ${trail.headline} because ${t.getMessage}"
-          logger.error(message, t)
-          Future.successful(Some(message))}
+      if(request.trail.alert.getOrElse(false)) {
+        client.send(request).map(request -> _)
+      } else {
+        // TODO MRB: use custom BreakingNewsError enum
+        logger.error(s"Failed to send a breaking news alert for trail ${request.trail} because alert was missing")
```
Reettaphant: We could use the structured logger as on line 50 to make the logs more discoverable

app/updates/BreakingNewsUpdate.scala
```diff
@@ -39,55 +38,66 @@ class BreakingNewsUpdate(val config: ApplicationConfiguration, val ws: WSClient,
...
+          errors.map { case(request, error) =>
+            // TODO MRB: note this is a change in logging class (used to be HandlingBreakingNewsCollection)
+            structuredLogger.putLog(LogUpdate(HandlingBreakingNewsTrail(collectionId, request.trail), email), "error", Some(new Exception(error.description)))
+            error.description
+          }
+        ))
+      } else {
+        Ok(Json.toJson(
+          successes.map { case(request, response) =>
+            // TODO MRB: structured logging here?
```
Reettaphant: 👍

app/updates/StructuredLogger.scala
```diff
@@ -9,7 +9,7 @@ import services.ConfigAgent
 import scala.collection.JavaConverters._
 
 class StructuredLogger(val config: ApplicationConfiguration, val configAgent: ConfigAgent) extends Logging {
-  def putLog(log: LogUpdate, level: String = "info", error: Option[Exception] = None): Unit = {
+  def putLog(log: LogUpdate, level: String = "info", error: Option[Throwable] = None): Unit = {
```
alexduf: Are you sure you want `Throwable` ?

It's usually reserved for when you want to catch `Exception`s and `Error`s such as`OutOfMemoryError`

app/util/BreakingNewsClient.scala
```diff
@@ -0,0 +1,91 @@
...
+    val requestBody = json.getBytes("UTF-8")
+
+    post(url, apiKey, requestBody) map { response =>
+      if(response.status == 201) {
+        validateFormat[BreakingNewsResponse](response.body)
+      } else {
+        Left(BreakingNewsResponseStatusError(response.status, response.body))
+      }
+    } recover {
+      case NonFatal(e) => Left(BreakingNewsResponseException(e))
```
alexduf: I think here if you use
`case NonFatal(e: Exception) => Left(BreakingNewsResponseException(e))`

you can then use `Exception` everywhere else

app/updates/StructuredLogger.scala
mbarton: Agreed. I've dropped it back to `Exception`

app/updates/BreakingNewsUpdate.scala
mbarton: We agreed IRL to respect the flag for safety, whilst we're still unsure of what user actions would cause it to be set to false

sihil:
Is this still needed @alexduf @mbarton or has the other work superceded this?

alexduf:
I think this is still very relevant, I missed this hadn't been merged before raising https://github.com/guardian/facia-tool/pull/841 though!

@mbarton do you think it's worth rebasing this, even though that means terrible conflicts?

jonathonherbert:
@alexduf @mbarton @Reettaphant happy to pair on the rebase to a) get it out the door and b) hand off knowledge to current team members so you're less ... haunted by the ghost of Fronts tool past 👍

alexduf:
Happy to help, should this be a ticket in the tools team backlog to make sure it gets picked?

mbarton:
I'm happy to pair on getting this rebased. That said the only bits of this PR that are still relevant are propagating the errors all the way up via an `Either` and that was always a speculative improvement so I don't think it's essential work given everything else we are doing at the moment.

sihil:
Given @mbarton 's comment I propose that this is closed and error propagation is added to a backlog somewhere?

PR #773, opened 2019-06-06
Add a test to make sure we can alert on the same story twice

Reettaphant:
## What's changed?

I added a second breaking news test to test that we can alert on a story, remove it and alert on it again as this workflow is a bit complicated and easy to break in any future refactors of the tool.

## Implementation notes

I decided to make a separate test for this rather than continue with the very long test to send breaking news that exists in this file. There is some code repetition with this test and the test that already exists but since the test above does a lot of assertions in the middle of the test which I don't want to do sharing the code for the test didn't feel like worth the trouble for this case.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/public/test/spec/breaking-news.spec.js
+++ b/public/test/spec/breaking-news.spec.js
@@ -241,4 +241,94 @@ describe('Breaking News', function () {
         .then(done)
         .catch(done.fail);
     });
+
+    it('Allows the same alert to be sent multiple times', function(done) {
+        const regions = this.testPage.regions;
+
+        regions.latest().trail(1).copy()
+        .then(() => {
+            return this.testPage.actions.edit(() => {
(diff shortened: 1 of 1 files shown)


mbarton:
Silly question, where do we send the alert a second time in the test?

```scala
return this.testPage.actions.publish(() => {
    return regions.breakingNewsModal().confirm();
})
```

^^^ I can only see this once

Reettaphant:
@mbarton It's not a silly question, I didn't actually test that, I tested that the breaking news button reappears so we would be in the position to send another alert. I could also test that we get the desired behaviour when clicking it.

mbarton:
Agreed IRL that since the server is not involved in these tests it's fine to just check that the button is re-enabled

PR #779, opened 2019-06-13
Editions frontend

RichieAHB:
## What's changed?

Editions can theoretically be rendered in the Fronts tool. This is currently being tested using the integration setup as the API endpoints are not live / haven't settled yet. However, assuming the models stay the same, hooking up to the actual `GET` endpoint should be trivial when it arrives.

## Implementation notes

This is the first pass for allowing edition `Issues` to be rendered (and at some point edited) in the current Fronts tool. The tack I've chose here is to define different strategies based on the routing. If we hit `/edition/edition-id` then we use a different strategy for loading `FrontsConfig`.

It's worth noting (as previously discussed IRL), I'm munging an `EditionIssue` into the shape of `FrontsConfig` here to make everything work out of the gate. This assumes that this is possible going forward (which I think it will be) but it's a key part of the implementation.

The next phase here will be injecting strategies for our middleware to save a collection if we're saving our collections to a different endpoint than at present (which I'm almost 100% sure we are).

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/components/FrontsEdit/FrontsEditStrategyPicker.tsx
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const FrontsEditStrategyPicker = ({
+  match: {
+    params: { priority, editionId }
+  },
+  dispatch,
+  ...props
+}: FrontsEditStrategyPicker) => {
+  const thunk = getConfigThunk(dispatch, priority, editionId);
```
jonathonherbert: Just a thought -- we add our path to the application state, and we can create matchers from our route params; could this logic just exist in our thunk, avoiding the need for an extra component that doesn't have any presentational work to do?

RichieAHB: Oh yeah, I forgot we have that there ... yeah I think that could probably live outside there in the tree in that case 🆗

RichieAHB:
@jonathonherbert - thanks for the heads up on that, here's a little demo pass injecting the stuff we need into our `thunks` through middleware here now that I'm using the `path` from the state.

client-v2/src/strategies.ts
```diff
@@ -0,0 +1,38 @@
...
+  fetchFrontsConfig
+} from 'services/faciaApi';
+
+const isValidPathForEdition = (priority: string, id?: string): id is string =>
+  !!id && priority === 'edition';
+
+const isValidPathForOtherPriority = (priority: string, id?: string) =>
+  !id && priority !== 'edition';
+
+const fetchFrontsConfigStrategy = (v2Path: string) => {
```
jonathonherbert: Super minor: I wonder whether it'd be nicer to have the application state as a dependency here, rather than v2Path? We can then select out v2Path ourselves. Then, as a developer, all you'd have to understand is the `state` type, which the IDE could give you advice on, rather than have to reason about the path type, which is stringly typed and possible to get wrong.

Reettaphant:
Are we still able to edit the current mock editions in the /edition endpoint after merging this pr? We should keep that endpoint working before our new editions stuff is functional.

client-v2/src/services/faciaApi.ts
```diff
@@ -19,6 +19,39 @@ import { CapiArticle } from 'types/Capi';
 import chunk from 'lodash/chunk';
 import { CAPISearchQueryResponse, checkIsResults } from './capiQuery';
 import flatMap from 'lodash/flatMap';
+import { EditionsIssue } from 'types/Edition';
+
+function fetchEditionsIssueAsConfig(editionId: string): Promise<FrontsConfig> {
+  return pandaFetch(`/editions-api/${editionId}`, {
```
Reettaphant: The endpoint for fetching editions is going to look like this: ` /editions-api/editions/:name/issues/:id` so can we pass the edition name for this method as well? We could also do it in a separate pr.

client-v2/src/strategies.ts
RichieAHB: This is the right thing to do, it was a _slightly_ easier refactor as this was the dependency of the function for the first implementation of this - but I did wonder if I should do it as I was pushing this PR.

RichieAHB:
> Are we still able to edit the current mock editions in the /edition endpoint after merging this pr? We should keep that endpoint working before our new editions stuff is functional.

@Reettaphant good point, should be a simple fix

client-v2/src/services/faciaApi.ts
RichieAHB: I think I'll leave this for a second PR as I think we can probably tidy up our routes and have some builders and parsers all in one place 👍

PR #781, opened 2019-06-17
Add strategy for fetching collections

RichieAHB:
## What's changed?

Because there will be different ways of fetching collections between editions and fronts tool (and the incoming models will be different). This is a WIP but it'd be good to get the eyes on this from everyone!

## Implementation notes

Start pulling out a strategy for fetching collections.  I've added a new strategy for this and additionally add a helper for picking strategies.

The aim here is to normalise the collection coming in to have a `draft` key and an empty `live` array. We can either assume and empty `live` key precludes the ability to look at the live - or we can define some `Context` that will control whether we can see a `live` toggle button.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/integration/fixtures/edition.js
+++ b/client-v2/integration/fixtures/edition.js
@@ -23,7 +23,7 @@ module.exports = {
           // lastUpdated?: number,
           // updatedBy?: string,
           // updatedEmail?: string,
-          articles: []
+          items: []
         },
         {
           id: 'collection-2',
@@ -33,7 +33,7 @@ module.exports = {
           // lastUpdated?: number,
           // updatedBy?: string,
           // updatedEmail?: string,
-          articles: []
+          items: []
         }
       ]
     }
--- a/client-v2/integration/fixtures/edition.js
+++ b/client-v2/integration/fixtures/edition.js
@@ -23,8 +23,7 @@ module.exports = {
           // lastUpdated?: number,
           // updatedBy?: string,
           // updatedEmail?: string,
-          live: [],
-          draft: []
+          articles: []
         },
         {
           id: 'collection-2',
@@ -34,8 +33,7 @@ module.exports = {
           // lastUpdated?: number,
           // updatedBy?: string,
           // updatedEmail?: string,
-          live: [],
-          draft: []
+          articles: []
         }
       ]
     }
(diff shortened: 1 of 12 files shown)


client-v2/integration/fixtures/edition.js
```diff
@@ -23,8 +23,7 @@ module.exports = {
           // lastUpdated?: number,
           // updatedBy?: string,
           // updatedEmail?: string,
-          live: [],
-          draft: []
+          articles: []
```
Reettaphant: Rich, I'm sorry, we just had a chat like 30 minutes ago and decided this should be called items instead of articles. Too fast moving!

RichieAHB: Haha - no worries, I just guessed this one anyway so it's no big deal 👍

RichieAHB:
FYI, this is failing the tests our actions now have some extra deps from the thunk args, I'll look into this and work out a nice way for this to be a pain to test with ... it may be fine as is.

RichieAHB:
With the strategy pattern it might be nicer to gather them into an `EditionsStrategy` and `FrontsStrategy`. For the time being they can stay like this until the interfaces become clear, but it's on my mental TODO to help readability of this area of the codebase.

client-v2/src/types/FaciaApi.ts
```diff
@@ -94,6 +97,23 @@ interface CollectionResponse {
   };
 }
 
+interface EditionCollectionFromResponse {
```
RichieAHB: I've put this here as I'm not sure it should be in `shared` but happy to reconsider.

Reettaphant: 👍

client-v2/src/util/configureStore.ts
```diff
@@ -14,17 +14,23 @@ import {
 } from './storeMiddleware';
 import { State } from 'types/State';
 import { ExtraThunkArgs } from 'types/Store';
-import { fetchFrontsConfigStrategy } from 'strategies';
+import { fetchFrontsConfigStrategy } from 'strategies/fetch-fronts-config';
+import { fetchCollectionsStrategy } from 'strategies/fetch-collection';
 
-export default function configureStore(initialState?: State) {
+export default function configureStore(
```
jonathonherbert: Re: conversation, if there's not much in the way of partial application etc. at this level then it might not be necessary to inject with the thunk params 👍

RichieAHB: 💣 they gone

PR #785, opened 2019-06-19
Refactor routing

RichieAHB:
## What's changed?

Changes the way some editions routes are handled such that if they change they all change together!

## Implementation notes

N/A

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/routes/routes.tsx
+++ b/client-v2/src/routes/routes.tsx
@@ -1,0 +1,56 @@
+import { priorities } from '../constants/priorities';
+import { matchPath } from 'react-router';
+
+export const base = '/v2';
+
+const editionsApi = (path: string) => `/editions-api${path}`;
+
+/**
+ * Fronts edit
+ */
+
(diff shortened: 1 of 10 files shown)


client-v2/src/routes/routes.tsx
```diff
@@ -0,0 +1,56 @@
...
+  matchPath<FrontsEditParams>(path, frontsEditPathProps);
+
+/**
+ * Issue
+ */
+
+interface IssueParams {
+  // The `priority` in this path is actually the edition issueId, but is mapped to priority
+  // so as to perserve things like favourite fronts and open fronts
+  priority: string;
```
jonathonherbert: We might want to ship this for demo, so 👍, but I'd love to sit down and solve this problem in such a way that this is unnecessary!

RichieAHB: I wholly agree

PR #788, opened 2019-06-19
Highlight current issues in datepicker

jonathonherbert:
~Downstream from #787 -- review this first, gentle reader!~

## What's changed?

Highlight the current issues in the datepicker, so the user can see what's already there. Yay!

Screenshot 2019-06-19 at 16 30 12

### Implementation details

This adds a date range to the listIssues query method; the search that looks for an issue on a particular day supplies a range corresponding to its day accordingly, rather than a single day.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/Editions/ManageEdition.tsx
+++ b/client-v2/src/components/Editions/ManageEdition.tsx
@@ -140,7 +140,8 @@ class ManageEdition extends React.Component<
           this.setState({
             infoMessage: 'New issue created!',
             isError: false,
-            currentIssue: issue
+            currentIssue: issue,
+            isCreatingIssue: false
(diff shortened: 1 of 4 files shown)


client-v2/src/components/Editions/ManageEdition.tsx
```diff
@@ -201,13 +185,14 @@ class ManageEdition extends React.Component<
   };
 
   private isIssuePresentForDate = (date: Moment) =>
```
RichieAHB: So this _would_ have worked before but we didn't have enough data in the state? Is that right?

PR #790, opened 2019-06-19
PRIVATE LINK: replace capi endpoint with private link endpoint

ajwl:
## What's changed?

In combination with other changes adding a private link endpoint and amending the configuration, this changes the CAPI requests:

1) `FaciaContentApiProxy.scala` - a key is no longer required for the PROD and CODE environments. However, relying on the `getOrElse` check on the request for this key, we will simply not provide a content api key for CODE or PROD

2) `capiQuery.ts` - the base url was used as a default parameter value. Now, I have removed this default, meaning that the function requires a string value to be passed in. The S3 config should pass in the correct string value

## Implementation notes

--> NB: Facia Tool dev environment will still use `contentapis.guardian.com` and a key. 

--> S3 config should be changed to remove the now unneeded keys. 

--> have updated tests so that a value is always passed in and it doesn't rely on the default 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/services/__tests__/capiQuery.spec.ts
+++ b/client-v2/src/services/__tests__/capiQuery.spec.ts
@@ -18,7 +18,8 @@ describe('CAPI', () => {
   describe('search', () => {
     it('makes a network request on a query', () => {
       const apiKey = 'my-api-key';
-      const capi = capiQuery();
+      const capiUrl = 'https://capiurl.guardian.com';
+      const capi = capiQuery(capiUrl);
       capi.search({
         'api-key': apiKey
       });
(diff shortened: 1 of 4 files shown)


ajwl:
@jonathonherbert - let me know if you're happy with this.

client-v2/src/services/__tests__/capiQuery.spec.ts
```diff
@@ -18,7 +18,8 @@ describe('CAPI', () => {
   describe('search', () => {
     it('makes a network request on a query', () => {
       const apiKey = 'my-api-key';
-      const capi = capiQuery();
+      const capiUrl = 'https://capiurl.guardian.com';
+      const capi = capiQuery(capiUrl);
```
ajwl: I did try to extract these into a global variable, using Jest's [Globals Object ](https://jestjs.io/docs/en/configuration#globals-object). But the code came out more confusing and not much shorter. Ie: 
```
    const apiKey = (global as any).apiKey;
    const capi = capiQuery((global as any).capiUrl);
```
So opted for this longer but more readable option

ajwl:
Thanks @jonathonherbert will deploy on CODE and if it's okay will merge.

PR #791, opened 2019-06-19
Fix id titles

RichieAHB:
## What's changed?

This ensures that we don't use the `EditionCollection` `ids` to display titles, but try the `displayName` and otherwise fallback to `id`. 

## Implementation notes

I could have broken this out into a `getFrontTitle` function but thought that might be overkill ... happy to add it if people disagree.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


RichieAHB:
Screenshot 2019-06-20 at 10 05 09

As can be seen here

PR #792, opened 2019-06-19
Editions Prefill

itsibitzi:
Use the CAPI client to get ourselves some editions prefill!

Pretty straight forward! As far as I know the only field we use in the prefill is the tag. I've hardcoded the search to be the last 3 days of content for now - can change that very easily though!

image

### DIFF
--- a/app/conf/Configuration.scala
+++ b/app/conf/Configuration.scala
@@ -192,7 +192,6 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
     }
   }
 
-
   object cdn {
     lazy val basePath = getString("assets.basePath").getOrElse("/")
   }
@@ -202,7 +201,9 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
 
     val contentApiLiveHost: String = getMandatoryString("content.api.host")
     def contentApiDraftHost: String = getMandatoryString("content.api.draft.iam-host")
+
     val editionsPrefillHost: String = "https://preview.content.guardianapis.com"
+    lazy val editionsKey: String = "test"
 
     lazy val key: Option[String] = getString("content.api.key")
     lazy val timeout: Int = 2000
--- a/app/conf/Configuration.scala
+++ b/app/conf/Configuration.scala
@@ -6,14 +6,13 @@ import java.net.URL
 import com.amazonaws.AmazonClientException
 import com.amazonaws.auth._
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
-import com.amazonaws.services.rds.auth.{GetIamAuthTokenRequest, RdsIamAuthTokenGenerator}
 import org.apache.commons.io.IOUtils
 import play.api.{Configuration => PlayConfiguration}
 import logging.Logging
 
 import scala.collection.JavaConverters._
 import scala.language.reflectiveCalls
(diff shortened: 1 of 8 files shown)


app/services/editions/IssueQueries.scala
```diff
@@ -193,13 +194,15 @@ trait IssueQueries {
           val collectionsForFront = rows
             .flatMap(row => row.collection.filter(_.frontId == front.id))
             .sortBy(_.index)
+            .map(_.collection)
+            .foldLeft(List.empty[EditionsCollection]) {(acc, cur) => if(acc.exists(f => f.id == cur.id)) acc else acc :+ cur}
```
RichieAHB: Is this a case for the `distinctBy` that you used before or am I missing something?

itsibitzi: `distinctBy` was added in scala 2.13 :(

itsibitzi: In another PR i've written an implicit class that enriches `distinctBy` onto `Collection` — I'll see if our libraries are compiles against 2.13 but I'm not going to hold my breath lmao

RichieAHB: No problems, was just double checking as I was just trying to work out what it did. No bigggggy.

app/conf/Configuration.scala
```diff
@@ -202,7 +201,9 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
 
     val contentApiLiveHost: String = getMandatoryString("content.api.host")
     def contentApiDraftHost: String = getMandatoryString("content.api.draft.iam-host")
+
     val editionsPrefillHost: String = "https://preview.content.guardianapis.com"
+    lazy val editionsKey: String = "test"
```
Reettaphant: Is this the key we are using for prod as well?

itsibitzi:
I had a chat with Reetta who agreed with you that we should probably just return an empty list if CAPI is down. In the future when we have our "fetch prefills" button the user will be able to wait for CAPI to be back and click the button so its allllll fiiiinnnee.

RichieAHB:
> In the future when we have our "fetch prefills" button the user will be able to wait for CAPI to be back and click the button so its allllll fiiiinnnee.

@itsibitzi this makes a lot of sense.

app/services/editions/IssueQueries.scala
```diff
@@ -170,13 +171,15 @@ trait IssueQueries {
...
```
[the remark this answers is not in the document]
itsibitzi: There's a collections row per article since we're joining across them.

app/services/Capi.scala
```diff
@@ -47,25 +54,54 @@ class GuardianCapi(config: ApplicationConfiguration)(implicit ex: ExecutionConte
...
+
+    var query = PrintSentQuery()
+      .page(1)
+      .pageSize(20)
+      .showFields("newspaper-edition-date")
+      .showFields("newspaper-page-number")
+      .showFields("internal-page-code")
+      .useDate("newspaper-edition")
+      .orderBy("newest")
+      .fromDate(issueDate.minus(Period.ofDays(3)).toInstant)
```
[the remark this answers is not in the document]
itsibitzi: This was for testing purposes, I've had a chat with David and it should only be that day for `print-sent` - will change

PR #793, opened 2019-06-19
Widen priority so we can use issueId

RichieAHB:
## What's changed?

This widens the `priority` type to `string` so that `issueId` can be used as a `priority`.

## Implementation notes

This sort of sucks because we probably won't be using `issueId` as a `priority` in future. But given how we've changed priority once already _and_ we don't actually use stringified versions of the priorities anywhere, it doesn't seem too calamitous. Again, happy to have other opinions!

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/services/faciaApi.ts
+++ b/client-v2/src/services/faciaApi.ts
@@ -37,7 +37,7 @@ function fetchEditionsIssueAsConfig(issueId: string): Promise<FrontsConfig> {
         fronts[front.id] = {
           ...front,
           collections: front.collections.map(collection => collection.id),
-          priority: 'edition'
+          priority: issueId
         };
         front.collections.forEach(collection => {
           collections[collection.id] = {
--- a/client-v2/src/types/FaciaApi.ts
+++ b/client-v2/src/types/FaciaApi.ts
@@ -1,4 +1,3 @@
-import { PriorityName } from './Priority';
 import { $Diff } from 'utility-types';
 import {
   CollectionFromResponse,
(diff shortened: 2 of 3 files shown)


client-v2/src/services/faciaApi.ts
```diff
@@ -68,6 +68,7 @@ function fetchFrontsConfig(): Promise<FrontsConfig> {
         const front: FrontConfig = {
           ...json.fronts[id],
           id,
+          displayName: id,
```
Reettaphant: Why do we need a display name here?

RichieAHB: Ah good point that was from a PR where I considered adding it as a normalisation step but then thought that might be a bit confusing. Will remove.

RichieAHB:
@Reettaphant there are quite a few things that rely on priority, we have selectors like `selectOpenFrontsByPriority`, favourite fronts, saving open fronts etc.

@jonathonherbert and I had a chat on another PR about it but essentially, I think it should be changed but I think we could probably get away with changing the word `priority` to `frontType` in certain parts of the application to make this clearer.

client-v2/src/services/faciaApi.ts
RichieAHB: Oops, that wasn't even in this PR 😨

Reettaphant:
Yes, I think `frontType` might be a good idea, it could help us to easily differentiate between different kinds of fronts and deal with presentational differences between them.

PR #796, opened 2019-06-20
fix fetch collections query and make collection updated fields NOT NULL

itsibitzi:
Pretty much Ronseal.

I arsed up the previous query by removing the prefixes and forgetting to add the commas back. Nice one m8.

I've also made the `updated_{on,by,email}` fields `NOT NULL` which simplifies the fetch query a little.


itsibitzi:
Attention! You'll need to wipe all collections from your database which have a `NULL` updated field.

The easiest way to do this locally is to drop the container volumes with `docker-compose down -v`. I'll sort out the code and prod DBs.

PR #799, opened 2019-06-24
New Daily Edition Template

itsibitzi:
## What's changed?

David updated the template, but sadly doesn't know how to use git. You know what they say, can't teach an old dog new tricks.

Also had to add a work around for the `newspaper-edition`  which is stored as a `date`[1] in CAPI and is relative to the publishing locale (aka 00:00, in whatever timezone it happens to be published in), so we've got to add a funny conversion. This is slightly problematic because the prefills are ordering by this edition date which means if you want to use the newspaper endpoint in US/AU you could end up with really old articles being prefilled in. As it stands I think the US/AU editions will be using the website's CAPI so hopefully this won't be a problem.

We _could_ make the tools ignore the timezone information as well but I'm not sure that's reasonable.

[1]: https://www.elastic.co/guide/en/elasticsearch/reference/current/date.html

### DIFF
--- a/app/services/Capi.scala
+++ b/app/services/Capi.scala
@@ -3,7 +3,7 @@ package services
 import java.io.IOException
 import java.net.URI
 import java.nio.charset.Charset
-import java.time.{Period, ZonedDateTime}
+import java.time.{Period, ZoneOffset, ZonedDateTime}
 
 import org.apache.http.client.utils.URLEncodedUtils
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
@@ -59,6 +59,9 @@ class GuardianCapi(config: ApplicationConfiguration)(implicit ex: ExecutionConte
       .parse(new URI(capiPrefillQuery.queryString), Charset.forName("UTF-8"))
       .asScala
 
+    // Horrible hack because composer/capi/whoever doesn't worry about timezones in the newspaper-edition date
+    val localDate = issueDate.toLocalDate.atStartOfDay().toInstant(ZoneOffset.UTC)
+
     var query = PrintSentQuery()
       .page(1)
       .pageSize(20)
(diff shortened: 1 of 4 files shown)


app/services/Capi.scala
```diff
@@ -86,7 +89,10 @@ class GuardianCapi(config: ApplicationConfiguration)(implicit ex: ExecutionConte
   }
 
   def getPrefillArticlePageCodes(issueDate: ZonedDateTime, capiPrefillQuery: CapiPrefillQuery): Future[List[String]] = {
-    this.getResponse(geneneratePrefillQuery(issueDate, capiPrefillQuery)) map { response =>
+    val query = geneneratePrefillQuery(issueDate, capiPrefillQuery)
+
+    println(query.toString())
```
jonathonherbert: Do we want to log this explicitly?

test/editions/EditionTemplates.scala
```diff
@@ -18,53 +18,53 @@ class editionTemplateTest extends FreeSpec with Matchers {
...
-      editionTemplateFronts(6) should matchPattern { case EditionsFrontSkeleton("news/international", _, _, _) => }
-      editionTemplateFronts(7) should matchPattern { case EditionsFrontSkeleton("frontpage/frontpage", _, _, _) => }
-      editionTemplateFronts(8) should matchPattern { case EditionsFrontSkeleton("news/national", _, _, _) => }
-      editionTemplateFronts(9) should matchPattern { case EditionsFrontSkeleton("money/money", _, _, _) => }
-      editionTemplateFronts(10) should matchPattern { case EditionsFrontSkeleton("review/review", _, _, _) => }
-      editionTemplateFronts(11) should matchPattern { case EditionsFrontSkeleton("feast/feast", _, _, _) => }
-      editionTemplateFronts(12) should matchPattern { case EditionsFrontSkeleton("special/special", _, _, _) => }
-    }
-  }
+//  "createEdition" - {
```
jonathonherbert: D'aww! But, yes.

app/services/Capi.scala
itsibitzi: Woops that was left over!

PR #801, opened 2019-06-26
Fix the URL escaping in prefills

itsibitzi:
This will be a lot neater when we move to a structured prefill rather than a query string.

We can't simply use a standard URL escaping function because that escapes syntactically relevant parts of the query string (e.g. `?` at the start and the various `&`s)

### DIFF
--- a/app/services/Capi.scala
+++ b/app/services/Capi.scala
@@ -1,7 +1,7 @@
 package services
 
 import java.io.IOException
-import java.net.URI
+import java.net.{URI, URLEncoder}
 import java.nio.charset.Charset
 import java.time.{Period, ZoneOffset, ZonedDateTime}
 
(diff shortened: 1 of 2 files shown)


app/services/Capi.scala
```diff
@@ -56,15 +56,15 @@ class GuardianCapi(config: ApplicationConfiguration)(implicit ex: ExecutionConte
...
+      .parse(new URI(capiPrefillQuery.escapedQueryString()), Charset.forName("UTF-8"))
       .asScala
 
     // Horrible hack because composer/capi/whoever doesn't worry about timezones in the newspaper-edition date
     val localDate = issueDate.toLocalDate.atStartOfDay().toInstant(ZoneOffset.UTC)
 
     var query = PrintSentQuery()
       .page(1)
-      .pageSize(20)
+      .pageSize(200)
```
jonathonherbert: 📖 👍

PR #804, opened 2019-06-27
DRAG & DROP: New styling for dropzones

ajwl:
## What's changed?

This improves drag and drop on the fronts tool to make it clearer where stories can be placed.

It changes:

-> Styling of the dropzones 
-> Styling of the drag object 
-> Behaviour: if you hover over a story, all possible dropzones show, not just the closest ones to your mouse. 

[Trello card](https://trello.com/c/rk1QmLeL/551-dragging-and-dropping-to-move-cards-and-sublinks)

### Dropzones 
Previously dropzones were grey boxes now they look like... 
NB: this is forcing all of them to show. In reality you only see the active zone when you are hovering

Screenshot 2019-06-27 at 17 45 46

### Drag Icon 
This now looks like a yellow box with a document icon. 

## Implementation notes

I have also deleted a test. See the notes below

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


ajwl:
Thank you - will look at the merge conflicts and also tests...

client-v2/src/lib/dnd/__tests__/dnd.spec.tsx
```diff
@@ -111,7 +111,7 @@ describe('Guration', () => {
     expect(edit).toEqual({
       data: { id: '1' },
       from: { type: 'b', id: '0', index: 0 },
-      to: { id: '2', type: 'a', index: 1 }
```
ajwl: Reordering to make it easier to to read

client-v2/src/lib/dnd/__tests__/dnd.spec.tsx
```diff
@@ -165,16 +165,11 @@ describe('Guration', () => {
...
+      to: { type: 'a', id: '2',  index: 0 }
     });
 
     runDrag(nodeProps)(dropProps, false);
 
-    expect(edit).toEqual({
-      data: { id: '1' },
-      from: { type: 'b', id: '0', index: 0 },
-      to: { id: '2', type: 'a', index: 1 }
-    });
```
ajwl: Okay. So I'm deleting a test here `allows dropping on to nodes when renderDrop is defined` (2nd assertion) because it's failing. This is why I think it's the right thing to do. 
Please check my reasoning here because I do not fully understand how this test file works. 
The test expected the article to go *below* the story hovered on (index 1) , with our current code, it's going above (index 0).

```
    Expected value to equal:
      {"data": {"id": "1"}, "from": {"id": "0", "index": 0, "type": "b"}, "to": {"id": "2", "index": 1, "type": "a"}}
    Received:
      {"data": {"id": "1"}, "from": {"id": "0", "index": 0, "type": "b"}, "to": {"id": "2", "index": 0, "type": "a"}}
```

Here's why I'm deleting it:
1) the drag and drop behaviour on the UI is working as expected. Whatever this test is failing on, it's not a behaviour that affects the user

2) As I understand @jonathonherbert 's changes to `Level`, [removing the DropIndexOffset method](https://github.com/guardian/facia-tool/pull/804/files#diff-7abf21e9815f013cb90e805bc9fba4e6L172) means that when you hover over a story the dragged article can only ever go on top. Which simplifies both the code and the user interaction. Therefore this test, checking things can go below, is redundant. 

keen to hear any thoughts 
@jonathonherbert @jennygrahamjones

jonathonherbert: @ajwl your explanation is thorough and makes perfect sense. This failing test is a result of a change I made, so I'm sorry you were the person that had to debug it! Removing this test is the correct thing to do. 👍

client-v2/src/lib/dnd/__tests__/dnd.spec.tsx
```diff
@@ -57,7 +57,7 @@ const runDrag = (type: any, data?: any, json: boolean = true) => (
 const setup = (jsx: React.ReactElement<any>) =>
   TestRenderer.create(jsx).getInstance();
 
-describe('Guration', () => {
+describe('Curation', () => {
```
jonathonherbert: 🤣 this isn't a typo, although @RichieAHB might be happy to see this change -- Guration was the WIP name of the drag and drop library! I don't think there are any other references to it in the codebase now (other than a dependency in package.json we can probably remore) so this is fine 😄

client-v2/src/lib/dnd/__tests__/dnd.spec.tsx
jennygrahamjones: ☝️ what @jonathonherbert said!

client-v2/src/lib/dnd/__tests__/dnd.spec.tsx
ajwl: 🙈 wow. Oh okay!

client-v2/src/lib/dnd/__tests__/dnd.spec.tsx
ajwl: thanks both!

PR #807, opened 2019-07-01
Enable overriding headline

itsibitzi:
You can now override headlines!

The frontend still displays the other fields but they're ignored pending a list of what we should remove from David.


itsibitzi:
To do once this is merged: hide the various unused filed (e.g. kicker)

PR #812, opened 2019-07-03
Expand all

jonathonherbert:
## What's changed?

Add an 'expand all' button to the fronts header. It opens all of the collections for that front.

Screenshot 2019-07-03 at 15 22 32

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/Front.tsx
+++ b/client-v2/src/components/FrontsEdit/Front.tsx
@@ -13,7 +13,8 @@ import {
   editorOpenOverview,
   editorCloseOverview,
   selectIsFrontOverviewOpen,
-  selectEditorArticleFragment
+  selectEditorArticleFragment,
+  editorCloseCollections
 } from 'bundles/frontsUIBundle';
 import {
   CollectionItemSets,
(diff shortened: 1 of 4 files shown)


client-v2/src/shared/components/input/ButtonRoundedWithLabel.tsx
```diff
@@ -1,14 +1,6 @@
...
 
-interface ContainerButtonWithLabelProps {
-  onClick: (e: React.MouseEvent<HTMLButtonElement>) => void;
-  label: string;
-  icon?: ReactNode | null;
-  className?: string;
-}
-
-const ContainerButton = styled('button')`
+export default styled.button`
```
jonathonherbert: A brief tidy up here to make this component more composable and remove an unnecessary SFC.

client-v2/src/components/FrontsEdit/Front.tsx
```diff
@@ -193,14 +196,27 @@ class FrontComponent extends React.Component<FrontProps, FrontState> {
         <FrontContainer>
           <FrontContentContainer>
             <SectionContentMetaContainer>
+              <CollapseAllButton
```
ajwl: Could we rename this component to `CollapseExpandButton`? It's just we have an "expand all" button made out of a component called `CollapseAllButton`

jonathonherbert: Absolutely, and thanks for spotting this -- as you say, this makes very little sense as it stands.

paperboyo: I, personally, would go further with Anna’s suggestion and rename it (dynamically) and deduplicate it also in the UI. Less clutter. One button for both actions, just like all the rest of expand/collapse buttons. Also – less mouse movements. Less movements, that is, if not for a jump:

jump

jonathonherbert: @paperboyo this would be ideal, but the difficult is that 'expanded' and 'collapsed' are not discrete states.

For example -- we can be in a situation where some articles in some collections are open, and some are not. In this situation (which is I think the case the majority of the time for most users) both 'collapse all' and 'expand all' are valid actions.

It's only when everything is open, or everything is closed, that only one action applies -- and this situation is I think quite rare! But it might be that I've missed something here, let me know 👍

jonathonherbert: @paperboyo -- also, re: collapsing, this is an artifact of the responsiveness of individual fronts, definitely not desirable as you point out, we should address it!

PR #813, opened 2019-07-03
Initial publishing scaffolding

itsibitzi:
This PR adds capabilities to convert a working edition into the publlished edition format and provide endpoints to display the published format as well as to publish it into an S3 bucket.

It adds a new mandatory config for the output bucket.

TODO
------
 - [x] create buckets in the CMS fronts account prior to deploying

### DIFF
--- a/app/model/editions/EditionsArticle.scala
+++ b/app/model/editions/EditionsArticle.scala
@@ -1,13 +1,13 @@
 package model.editions
 
-import com.gu.editions.PublishedArticle
+import com.gu.editions.{PublishedArticle, PublishedArticleMetadata}
 import play.api.libs.json.Json
 import scalikejdbc.WrappedResultSet
 
 case class EditionsArticle(pageCode: String, addedOn: Long) {
-  def toPublishedArticle(): PublishedArticle = PublishedArticle(
+  def toPublishedArticle: PublishedArticle = PublishedArticle(
     pageCode.toLong,
-    None
+    PublishedArticleMetadata(None, None, None) // TODO (sihil): Store in DB and populate here
   )
 }
 
--- a/app/model/editions/EditionsArticle.scala
+++ b/app/model/editions/EditionsArticle.scala
@@ -1,9 +1,15 @@
 package model.editions
 
+import com.gu.editions.PublishedArticle
 import play.api.libs.json.Json
 import scalikejdbc.WrappedResultSet
 
-case class EditionsArticle(pageCode: String, addedOn: Long)
+case class EditionsArticle(pageCode: String, addedOn: Long) {
+  def toPublishedArticle(): PublishedArticle = PublishedArticle(
+    pageCode.toLong,
+    None
+  )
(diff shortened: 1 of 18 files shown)


app/services/editions/publishing/PublishedIssuesBucket.scala
```diff
@@ -0,0 +1,32 @@
...
+    .build()
+
+  private def createIssuePath(issue: PublishedIssue, user: User) = {
+    val firstName = user.firstName.replace(" ", "_")
+    val lastName = user.lastName.replace(" ", "_")
+    s"${issue.name}/${issue.issueDate.toLocalDate.toString}/${OffsetDateTime.now().toString}_${firstName}_$lastName.json"
+  }
+
+  def putIssue(issue: PublishedIssue, user: User) = {
+    client.putObject(bucket, createIssuePath(issue, user), Json.stringify(Json.toJson(issue)))
```
sihil: this will wind up with something that is `text/plain` which might not be ideal. probably fine for now.

app/model/editions/EditionsIssue.scala
```diff
@@ -17,7 +18,14 @@ case class EditionsIssue(
     launchedBy: Option[String],
     launchedEmail: Option[String],
     fronts: List[EditionsFront]
-)
+) {
+  def toPublishedIssue(): PublishedIssue = PublishedIssue(
```
sihil: Not a big fan of this pattern of enshrining conversion logic into the model. Would prefer this sat in a conversion companion object somewhere.

sihil:
Planning to merge this in and deploy tomorrow morning.

app/model/editions/EditionsArticle.scala
```diff
@@ -1,9 +1,15 @@
 package model.editions
 
+import com.gu.editions.PublishedArticle
 import play.api.libs.json.Json
 import scalikejdbc.WrappedResultSet
 
-case class EditionsArticle(pageCode: String, addedOn: Long)
+case class EditionsArticle(pageCode: String, addedOn: Long) {
```
akash1810: Why a `Long` and not a `DateTime` here?

akash1810:
> convert a working edition into the publlished edition

Is a "working edition" a draft edition?

app/model/editions/EditionsArticle.scala
itsibitzi: It’s what the client expects

itsibitzi:
A working edition would be the tools version of the model and the published edition what the editions API wants.

app/model/editions/EditionsArticle.scala
sihil: Can we block @itsibitzi until he's back from holiday? The new model uses a `ZonedDateTime` IIRC.

sihil:
I've made some CFN changes to create and permission the bucket: https://github.com/guardian/editorial-tools-platform/pull/302

PR #820, opened 2019-07-10
Condense the two image inputs into a single, beautiful synthesis of both

jonathonherbert:
## What's changed?

There are currently two image inputs that control a single output -- the current image override. After much discussion with @akemitakagi and @paperboyo we think we've got a solution that uses just one:

single-image-form

A few notes --

- You can drag any image source into this input, and you can always drag the current image out -- even if it's a default.
- Adding an image now always automatically enables the 'Replace Image' toggle, which saves a click.
- The messaging for 'add image' changes to 'add cutout' in cutout mode -- hopefully that's a bit clearer.

This component has lots of different interactions -- please feed back! I'll also be running this by CP before merging.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


paperboyo:
💎

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
```diff
@@ -495,6 +458,35 @@ class FormComponent extends React.Component<Props, FormComponentState> {
...
+    }
+    return 'primaryImage';
+  };
+
+  private handleImageChange: EventWithDataHandler<React.ChangeEvent<any>> = (
+    e,
+    ...args: [any?, any?, string?]
+  ) => {
+    // If no image overrides are enabled, enable the default imageReplace property.
+    // This saves the user a click; adding an image without enabling would be very unusual.
```
ajwl: Small thing - I don't understand this comment. Is it that: "If no images are overriden, enable the default imageReplace property. This saves the user a click, because they no longer need to enable this to start changing the image?

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
```diff
@@ -495,6 +458,35 @@ class FormComponent extends React.Component<Props, FormComponentState> {
...
+  private getImageFieldName = () => {
+    if (this.props.imageCutoutReplace) {
+      return 'cutoutImage';
+    }
+    return 'primaryImage';
+  };
+
+  private handleImageChange: EventWithDataHandler<React.ChangeEvent<any>> = (
+    e,
+    ...args: [any?, any?, string?]
```
ajwl: do we need this 2nd param? doesn't look like it gets used?

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
ajwl: Jenny's suggestion: `Given a state where nothing is toggled, and an image is added - we would assume that the imageReplace toggle should be automatically activated, as this is the most likely reason for the user to add an image.`

ajwl:
BTW - I get two TSLint errors on these lines with a mismatching type problem: 
- `/ArticleFragmentForm.tsx(130,11)`
- `ArticleFragmentForm.tsx(451,19)`

Otherwise very small changes. And happy to approve

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
jonathonherbert: Thanks for spotting, the `no` in this comment means the reverse of what was intended! I'll rephrase. 👍

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
jonathonherbert: strangely, we do (to conform to the `EventWithDataHandler` type), but it's not clear from the code why it's present. I'll see if I can refactor to remove, and if not, I'll add a comment to explain.

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
jonathonherbert: Hopefully that's clearer now!

PR #821, opened 2019-07-10
INPUTS: Automatically resize textareas to fit their content

ajwl:
with @jennygrahamjones 

## What's changed?

Users wanted the kicker and headline input fields to show all of the text when the form is opened. Previously it was at a set height which hid some of the words 

BEFORE 
-text is cut off
Screenshot 2019-07-10 at 17 29 21

AFTER 
-textarea automatically expands to text height
Screenshot 2019-07-10 at 17 30 21

[Trello ticket](https://trello.com/c/m0ZJM8Ce/590-make-trail-length-box-bigger-so-editors-can-see-full-text-in-one-glance)

## Implementation notes

In order to use lifecycle methods - we had to convert this functional component into a class component. 

In order to find the height of the content, we need to access the underlying DOM element behind the React component. Using `ref` allows us to do this. 

However, because we are using a Higher Order Component taking a styled component as an input, we need to use `innerRef` a property already set on a styled component, allowing access to the `ref` on itself. 

This allows the actual node to be passed up to the Higher Order Component.  

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
+++ b/client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
@@ -298,7 +298,6 @@ class FormComponent extends React.Component<Props, FormComponentState> {
               placeholder={articleCapiFieldValues.headline}
               component={InputTextArea}
               useHeadlineFont
-              rows="2"
               originalValue={articleCapiFieldValues.headline}
               data-testid="edit-form-headline-field"
             />
--- a/client-v2/src/shared/components/input/CreateInput.tsx
+++ b/client-v2/src/shared/components/input/CreateInput.tsx
@@ -11,7 +11,15 @@ type Props = {
   label?: string;
   // If provided, the user can revert to this value by clicking the 'rewind' button.
   originalValue?: string;
+  component: React.Component<
+    React.HTMLAttributes<HTMLInputElement> & StyledProps<any>
+  >;
+  type?: string;
 } & WrappedFieldProps;
+
+interface State {
+  inputHeight: number;
+}
 
 const RewindButton = styled.button.attrs({
   type: 'button'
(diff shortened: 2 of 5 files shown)


client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
```diff
@@ -298,7 +298,6 @@ class FormComponent extends React.Component<Props, FormComponentState> {
               placeholder={articleCapiFieldValues.headline}
               component={InputTextArea}
               useHeadlineFont
-              rows="2"
```
ajwl: Was overriding our height changes

client-v2/src/shared/components/input/InputTextArea.ts
```diff
@@ -5,7 +5,6 @@ const InputTextAreaBase = InputBase.withComponent('textarea').extend<{
   minHeight?: number;
   maxHeight?: number;
 }>`
-  height: auto;
```
ajwl: Not necessary now height is set by component

client-v2/src/components/FrontsEdit/ArticleFragmentForm.tsx
jonathonherbert: 👍

client-v2/src/shared/components/input/CreateInput.tsx
```diff
@@ -44,20 +52,59 @@ export default (
...
-          </RewindButton>
-        )}
-      </InputLabel>
-    )}
-    <InputComponentContainer>
-      <Component id={label} {...input} {...rest} type={type} />
-    </InputComponentContainer>
-  </InputContainer>
-);
+) => {
```
jonathonherbert: This is really neat! Now that our interface depends upon the incoming component being a text field or textarea, it might be better to make this function name more specific -- it wouldn't make sense to supply it with a checkbox input, for example, but there's nothing to suggest that wouldn't be a good idea.

client-v2/src/shared/components/input/CreateInput.tsx
```diff
@@ -44,20 +52,59 @@ export default (
...
+  return class extends React.Component<Props, State> {
+    private inputElement: React.RefObject<HTMLInputElement>;
+    constructor(props: Props) {
+      super(props);
+      this.inputElement = React.createRef();
+      this.state = { inputHeight: 36 };
```
jonathonherbert: Can we always rely on this value staying in sync with our styles? I wonder whether we can derive it from constants/theme (if there isn't a value for inputs, we could create one), but I'm also wondering whether it's necessary at all --- is there some strange pop-in behaviour when the component first renders, or similar, if it isn't there? We could always avoid adding the height in the `style` prop below if we don't have a value in the state yet.

client-v2/src/shared/components/input/CreateInput.tsx
```diff
@@ -44,20 +52,59 @@ export default (
...
+    }
+
+    public componentDidMount() {
+      this.updateHeight();
+    }
+
+    public componentDidUpdate(prevProps: Props) {
+      if (prevProps !== this.props) {
```
jonathonherbert: As an optimisation, could we go one better here, and only update the height if the prop that is likely to change it (the input value) has changed?

client-v2/src/shared/components/input/CreateInput.tsx
ajwl: Ah yup - will look up the themes to check if there is a pre-defined value.
I think it is beneficial to keep a default there, but will investigate.

client-v2/src/shared/components/input/CreateInput.tsx
jennygrahamjones: It looks like `createInput` is exclusively used by InputText and InputTextArea. As a simple solution, we can rename the file and the function to be more clear about this fact.

client-v2/src/shared/components/input/CreateInput.tsx
jennygrahamjones: The textarea element provides its own default height of 2 rows. We can use this by allowing the`inputHeight` to default to the 'auto' value, rather than specifying a number ourselves.

client-v2/src/shared/components/input/InputTextArea.ts
```diff
@@ -1,14 +1,13 @@
...
   resize: vertical;
   min-height: ${({ minHeight = 40 }) => minHeight}px;
   max-height: ${({ maxHeight = 120 }) => maxHeight}px;
 `;
 
-export default createInput(InputTextAreaBase);
+export default createResizeableTextInput(InputTextAreaBase);
```
jonathonherbert: Super clear 👍

client-v2/src/shared/components/input/CreateResizeableTextInput.tsx
```diff
@@ -0,0 +1,110 @@
...
+const InputComponentContainer = styled.div`
+  position: relative;
+  &:hover {
+    ${RewindButton} {
+      display: block;
+    }
+  }
+`;
+
+const createResizeableTextInput = (
```
jonathonherbert: Super clear 👍

PR #811, opened 2019-07-03
Add inline form

jonathonherbert:
NB: This is a WIP, and is likely to change significantly before it's ready for review.

## What's changed?

Adds an option to open forms inline with the collection. At the moment, this replaces the form on the right hand side, but this is likely to move to a settings-based option where the user can choose one or the other.

inline

## Implementation notes

TBC

### General
- [ ] 🤖 Relevant tests added
- [ ] ✅ CI checks / tests run locally
- [ ] 🔍 Checked on CODE

### Client
- [ ] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ ] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #822, opened 2019-07-10
Highlight overview collection on scroll

jonathonherbert:
## What's changed?

Highlight the 'current' collection in the overview -- by this interpretation, the one that's closest to the top of the container.

scroll-behaviour

## Implementation notes

At the moment, this is less efficient than it should be, because of the way the components are arranged. This is likely to change as a result of #811, so I didn't make that optimisation here, as performance seems adequate.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/Front.tsx
+++ b/client-v2/src/components/FrontsEdit/Front.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { styled } from 'constants/theme';
+import { styled, theme as globalTheme } from 'constants/theme';
 import { connect } from 'react-redux';
 import { Root, Move, PosSpec } from 'lib/dnd';
 import { State } from 'types/State';
@@ -36,6 +36,8 @@ import ButtonCircularCaret from 'shared/components/input/ButtonCircularCaret';
 import ButtonRoundedWithLabel, {
   ButtonLabel
 } from 'shared/components/input/ButtonRoundedWithLabel';
+import sortBy from 'lodash/sortBy';
+import debounce from 'lodash/debounce';
 
 const FrontContainer = styled('div')`
   height: 100%;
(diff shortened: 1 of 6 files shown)


client-v2/src/components/FrontsEdit/Front.tsx
```diff
@@ -236,23 +288,30 @@ class FrontComponent extends React.Component<FrontProps, FrontState> {
                 />
               </OverviewToggleContainer>
             </SectionContentMetaContainer>
-            <FrontCollectionsContainer>
+            <FrontCollectionsContainer
+              onScroll={this.handleScroll}
+              innerRef={ref => (this.collectionContainerRef = ref)}
```
ajwl: Quick thought - according to [React docs](https://reactjs.org/docs/refs-and-the-dom.html#dont-overuse-refs) there is a more modern way to do this... 
We're already doing it both ways in the code. Just preference but I'd argue the modern way with `React.createRef();` is more readable. 

Also! can we name the `ref` param as `element` so it doesn't get mixed up with `ref` the attribute?

client-v2/src/components/FrontsEdit/Front.tsx
```diff
@@ -236,23 +288,30 @@ class FrontComponent extends React.Component<FrontProps, FrontState> {
...
-                    alsoOn={this.props.alsoOn}
-                    handleInsert={this.handleInsert}
-                    handleMove={this.handleMove}
-                    selectArticleFragment={this.props.selectArticleFragment(
-                      this.props.id
-                    )}
-                    handleArticleFocus={this.handleArticleFocus}
-                  />
+                  <CollectionContainer
+                    innerRef={ref => (this.collectionRefs[collectionId] = ref)}
```
ajwl: Quick thing - can we rename `ref` here to something like `element`? just get a bit confused between the param and the attribute `ref`.

jonathonherbert: Great point 👍

PR #828, opened 2019-07-11
Adds irregular magazine content to the edition

blishen:
Adds some irregular and regular but not weekly content.


blishen:
Adds additional prefills and one front to the edition to cater for irregular magazine content

PR #830, opened 2019-07-11
Remove width limit for fronts

jonathonherbert:
## What's changed?

This PR allows fronts to expand to fill the available space.

responsive-full

There's no limit on how wide fronts can now get -- after chats with Robert and @akemitakagi we're confident that users will size the tool for their own needs, but we can revisit this assumption if necessary.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


jonathonherbert:
The new inline form work is incoming and it's unlikely to play nicely with this design -- @akemitakagi what do you think?

PR #829, opened 2019-07-11
Lint :)

jonathonherbert:
## What's changed?

A previous merge to master broke our build, because ... linting :) This quick PR should restore order. It also gets rid of some red ink in the console that should really have been caught by our e2e tests -- worth investigating when we get the time.

### General
- [ ] 🤖 Relevant tests added
- [ ] ✅ CI checks / tests run locally
- [ ] 🔍 Checked on CODE

### Client
- [x] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [x] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #831, opened 2019-07-11
New collection and clipboard styling

jonathonherbert:
~Dear reader -- this is in draft, as it depends on #804 and #829! The diff will be bloated as a result -- probably best not to review just yet.~ This is now ready for review.

## What's changed?

The clipboard and collection now look ... fancy!

Before:

Screenshot 2019-07-11 at 18 37 09

After:

Screenshot 2019-07-11 at 18 49 24

## Implementation notes

This simplifies our implementations a little, as we can remove the Polaroid components 🎉 (@ajwl!)

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/components/FrontsEdit/CollectionComponents/Sublinks.tsx
```diff
@@ -12,15 +12,13 @@ import { collectionDropTypeBlacklist } from 'constants/fronts';
...
   display: flex;
   min-height: 30px;
-  border-top: ${({ isClipboard }) =>
-    isClipboard ? 'none' : '1px solid #c9c9c9'};
-  background-color: ${({ isClipboard, dragHoverActive }) =>
-    dragHoverActive ? `#ededed` : isClipboard ? '#f6f6f6' : '#fff'};
-  flex-direction: ${({ isClipboard }) => (isClipboard ? 'column' : 'row')};
+  border-top: 1px solid #c9c9c9;
+  background-color: ${({ dragHoverActive }) =>
+    dragHoverActive ? `#ededed` : '#fff'};
```
jennygrahamjones: In other places, we reference the theme.ts constants declared for these colours. Is there a reason we don't do it here?

jonathonherbert: No, we should use the theme 👍 I'll make the change.

PR #832, opened 2019-07-12
Separate clipboards

Reettaphant:
## What's changed?

Implements two different clipboards, editions clipboard and fronts clipboard because editing these 
## Implementation notes

Both clipboards are saved in state in case a user switches between the two editing modes without refreshing.

I have made some compromises to get this feature rolled out in the timely fashion. 
- The two clipboard are stored as separate arrays in dynamo instead of putting them in a neat clipboard object to avoid the need to migrate existing user data
- I have added the clipboard type as an optional field in the shared `InsertArticleFragmentPayload` type - I've added a comment explaining that this is not ideal but because of the code shared between invoking different insert actions passing this in as an optional parameter seemed easier than getting typescript to agree about the types when doing type refinement.  

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/UserDataController.scala
+++ b/app/controllers/UserDataController.scala
@@ -16,18 +16,11 @@ class UserDataController(frontsApi: FrontsApi, dynamo: Dynamo, val deps: BaseFac
 
 
   def putClipboardContent() = APIAuthAction { request =>
-    println("-----")
-    println(request.body)
-    println(request.body.asJson)
 
     val clipboardArticleRequest: Option[Map[String, List[Trail]]] = request.body.asJson.flatMap(jsValue => {
-        //TODO we should not fail if the list is missing
-      println("js value is ", jsValue)
       val js = jsValue.as[Map[String, List[Trail]]]
-      println("js is ", js)
       jsValue.asOpt[Map[String, List[Trail]]]
     })
-    println(clipboardArticleRequest)
 
     clipboardArticleRequest match {
       case Some(cliboardArticleLists) => {
--- a/app/controllers/UserDataController.scala
+++ b/app/controllers/UserDataController.scala
@@ -16,14 +16,24 @@ class UserDataController(frontsApi: FrontsApi, dynamo: Dynamo, val deps: BaseFac
 
 
   def putClipboardContent() = APIAuthAction { request =>
+    println("-----")
+    println(request.body)
+    println(request.body.asJson)
 
-    val clipboardArticles: Option[List[Trail]] = request.body.asJson.flatMap(jsValue =>
-      jsValue.asOpt[List[Trail]])
+    val clipboardArticleRequest: Option[Map[String, List[Trail]]] = request.body.asJson.flatMap(jsValue => {
+        //TODO we should not fail if the list is missing
+      println("js value is ", jsValue)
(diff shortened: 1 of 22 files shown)


client-v2/src/shared/types/Action.ts
```diff
@@ -29,10 +30,16 @@ interface UpdateArticleFragmentMeta {
   };
 }
 
+// The clipboard type here is only needed for when we are inserting
+//  into a clipboard. Ideally clipboard type would only belong to the
+//  clipboard insert action because a lot of the code to handle insertions
+//  in different places in shared is cleaner to have clipboard type here
+//  as optional
```
Reettaphant: If there is a simpler way of doing please let me know. Not keen to start making the code to create insert actions even more complicated, I like the fact that a lot of inserting code is shared.

client-v2/src/reducers/clipboardReducer.ts
```diff
@@ -21,15 +24,51 @@ const clipboard = (
...
     }
+
     case INSERT_CLIPBOARD_ARTICLE_FRAGMENT: {
-      return insertAndDedupeSiblings(
-        state,
-        [action.payload.articleFragmentId],
-        action.payload.index,
-        articleFragmentsSelector(prevSharedState)
-      );
+      if (action.payload.clipboardType === 'fronts') {
```
Reettaphant: I know there is a lot of code duplication here but I sort of like it because it makes the intention and what is being updated very clear.

app/controllers/UserDataController.scala
```diff
@@ -17,13 +17,16 @@ class UserDataController(frontsApi: FrontsApi, dynamo: Dynamo, val deps: BaseFac
 
   def putClipboardContent() = APIAuthAction { request =>
 
-    val clipboardArticles: Option[List[Trail]] = request.body.asJson.flatMap(jsValue =>
-      jsValue.asOpt[List[Trail]])
+    val clipboardArticleRequest: Option[Map[String, List[Trail]]] = request.body.asJson.flatMap(jsValue => {
+      val js = jsValue.as[Map[String, List[Trail]]]
+      jsValue.asOpt[Map[String, List[Trail]]]
+    })
```
jonathonherbert: Just a small typo :)

app/controllers/UserDataController.scala
```diff
@@ -17,13 +17,16 @@ class UserDataController(frontsApi: FrontsApi, dynamo: Dynamo, val deps: BaseFac
...
 
-    clipboardArticles match {
-      case Some(articles) => {
+    clipboardArticleRequest match {
+      case Some(cliboardArticleLists) => {
```
jonathonherbert: Just a small typo :)

client-v2/src/actions/ArticleFragments.ts
```diff
@@ -54,6 +57,11 @@ type InsertThunkActionCreator = (
   removeAction: Action | null
 ) => ThunkResult<void>;
 
+const getEditingMode = (state: State): ClipboardType => {
```
jonathonherbert: Can we use the `select` prefix here to make it clear this is operating on the app state?

jennygrahamjones:
@Reettaphant are we okay to delete this branch?

PR #836, opened 2019-07-15
Handle empty storiesVisibleByStage object in collection API responses

jonathonherbert:
## What's changed?

We now handle empty storiesVisibleByStage objects in our collection responses. This was very unlikely to be empty in website- or app-facing collections, but is very common in e-mail collections, which were failing to parse as a result. As a result, e-mail fronts should now appear as expected.
 
## Implementation notes

This was exposed by correcting the typings and following the code, and makes a strong case for us deriving our API types from the Scala if at all possible!

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/fixtures/collectionsEndpointResponse.ts
+++ b/client-v2/src/fixtures/collectionsEndpointResponse.ts
@@ -48,7 +48,7 @@ export const scJohnsonPartnerZoneCollection = [
   null
 ];
 
-export const getCollectionsThunkFaciaApiResponse = [
+export const getCollectionsApiResponse = [
   {
     id: 'testCollection1',
     collection: {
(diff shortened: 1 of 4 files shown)


client-v2/src/fixtures/collectionsEndpointResponse.ts
```diff
@@ -48,7 +48,7 @@ export const scJohnsonPartnerZoneCollection = [
   null
 ];
 
-export const getCollectionsThunkFaciaApiResponse = [
+export const getCollectionsApiResponse = [
```
jonathonherbert: A quick rename here, as otherwise things got very unwieldy later on!

client-v2/src/types/FaciaApi.ts
```diff
@@ -90,8 +90,8 @@ interface CollectionResponse {
   id: string;
   collection: CollectionFromResponse;
   storiesVisibleByStage: {
-    live: VisibleArticlesResponse;
-    draft: VisibleArticlesResponse;
+    live?: VisibleArticlesResponse;
```
jonathonherbert: The offending types.

PR #838, opened 2019-07-15
persist image overrides for Editions

akash1810:
## What's changed?
[Card](https://trello.com/c/a2P58r1x/36-edit-the-relevant-metadata-fields-for-editions-exact-fields-to-store-tbc)

Stores image overrides.

Also splits out the database model from the client model with the intention of the database model being a simpler representation that gets serialised into the client model.

## Implementation notes
n/a

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


app/model/editions/EditionsClientCollection.scala
```diff
@@ -1,57 +1,58 @@
 package model.editions
 
+import model.editions.client.ClientArticleMetadata
 import play.api.libs.json.Json
 
 // Ideally the frontend can be changed so we don't have this weird modelling!
 
-case class EditionsFrontendArticle(id: String, frontPublicationDate: Long, meta: Option[ArticleMetadata])
+case class EditionsClientArticle(id: String, frontPublicationDate: Long, meta: Option[ClientArticleMetadata])
```
jonathonherbert: I really like the naming here, 👍

app/model/editions/client/ClientArticleMetadata.scala
```diff
@@ -0,0 +1,99 @@
+package model.editions.client
+
+import model.editions.{ArticleMetadata, Image, ImageOption}
+import play.api.libs.json.Json
+
+case class ClientArticleMetadata (
```
jonathonherbert: Huh -- how did we get away with not modelling this before? Did we just transparently pass the overrides through as JSON?

akash1810: Looks like we have some models [here](https://github.com/guardian/facia-scala-client/blob/master/facia-json/src/main/scala/com/gu/facia/client/models/Collection.scala#L126) which we could reuse.

akash1810:
For some reason, `imageCutoutReplace` is always `true` which isn't correct. DNM for now.

app/model/editions/ImageOption.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import enumeratum.{EnumEntry, PlayEnum}
+import enumeratum.EnumEntry.Camelcase
+
+sealed abstract class ImageOption extends EnumEntry with Camelcase
+object ImageOption extends PlayEnum[ImageOption] {
+  case object UseArticleTrail extends ImageOption
+  case object Hide extends ImageOption
+  case object ReplaceCutout extends ImageOption
+  case object ReplaceSlideshow extends ImageOption
```
sihil: I'm 99% sure that we're not supporting slideshows but @blishen should be able to clarify.

sihil: I understand this came from @blishen so all good - just sounds a little like scope creep 😿

akash1810:
I'm going to split this out into multiple PRs tomorrow. One for image stuff (which currently has a client side bug as far as I can tell) and one for the rest. This'll mean we deliver value faster. I'll also write tests against the image stuff for my sanity.

sihil:
I've merged the first couple of commits in #844 and also rebased this on top of master. You might want to hard reset against origin.

app/model/editions/EditionsArticle.scala
```diff
@@ -1,35 +1,72 @@
 package model.editions
 
-import com.gu.editions.{PublishedArticle, PublishedFurniture}
+import com.gu.editions.{PublishedArticle, PublishedFurniture, MediaUrl}
 import play.api.libs.json.Json
 import scalikejdbc.WrappedResultSet
 
-case class ArticleMetadata(
+case class Image (
```
akash1810: This and `MediaUrl` would be the same model - need to confirm with @AWare about the structure needed.

app/model/editions/client/ClientArticleMetadata.scala
```diff
@@ -0,0 +1,106 @@
...
+      articleMetadata.headline,
+      articleMetadata.customKicker,
+      articleMetadata.customKicker.map(_ => true),
+      articleMetadata.trailText,
+      articleMetadata.showQuotedHeadline,
+      articleMetadata.showByline,
+      articleMetadata.byline,
+
+      articleMetadata.imageOption match {
+        case Some(io) if io == ImageOption.Hide => Some(true)
```
akash1810: Must be a better way to do this! `if articleMetadata.imageOption is defined and == ImageOption.Hide then Some(true) else None`

sihil: I think you want `articleMetadata.imageOption.collect { case ImageOption.Hide => true }`

sihil: You can use this in other places where your default pattern maps to `None`

PR #840, opened 2019-07-16
Editions clipboard

Reettaphant:
## What's changed?

Creates a separate clipboard for editions and for fronts.

## Implementation notes

Only keeps one of the clipboards in the redux state at a time. This is going to cause problems if a user switches between fronts editing and editions editing without refreshing. We are not going to have the same user editing fronts and editions in the first instance so can look into fixing this later.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/V2App.scala
+++ b/app/controllers/V2App.scala
@@ -18,6 +18,7 @@ class V2App(isDev: Boolean, val acl: Acl, dynamo: Dynamo, val deps: BaseFaciaCon
 
   def index(priority: String = "", frontId: String = "") = getCollectionPermissionFilterByPriority(priority, acl)(ec) { implicit req =>
 
+    val editingEdition = priority.startsWith("issues")
     val userDataTable = Table[UserData](config.faciatool.userDataTable)
 
     val jsFileName = "dist/app.bundle.js"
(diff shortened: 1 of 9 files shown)


app/controllers/V2App.scala
```diff
@@ -39,6 +40,12 @@ class V2App(isDev: Boolean, val acl: Acl, dynamo: Dynamo, val deps: BaseFaciaCon
     val maybeUserData: Option[UserData] = Scanamo.exec(dynamo.client)(
       userDataTable.get('email -> userEmail)).flatMap(_.right.toOption)
 
+    val clipboardArticles = if (editingEdition)
```
jonathonherbert: As discussed, this is fine for now, but it'd be brilliant to have our data model take care of this logic by itself with a Map[EditingMode, ClipboardArticles] or similar.

client-v2/src/actions/Clipboard.ts
```diff
@@ -56,11 +57,23 @@ function storeClipboardContent(clipboardContent: NestedArticleFragment[]) {
...
-  return () =>
-    saveClipboard(clipboardContent.articles).catch(() => {
+}): ThunkResult<Promise<NestedArticleFragment[]>> {
+  return async (_, getState: () => State) => {
+    const saveClipboardResponse = await saveClipboardStrategy(
+      getState(),
+      clipboardContent.articles
+    );
+    try {
+      if (!saveClipboardResponse) {
```
jonathonherbert: Why the try/catch here -- it doesn't look like it's possible for this code to throw?

client-v2/src/strategies/save-clipboard.ts
```diff
@@ -0,0 +1,16 @@
+import { State } from 'types/State';
+import { saveClipboard, saveEditionsClipboard } from 'services/faciaApi';
+import { runStrategy } from './run-strategy';
+import { NestedArticleFragment } from 'shared/types/Collection';
+
+const saveClipboardStrategy = (
```
jonathonherbert: 👍

app/controllers/V2App.scala
itsibitzi: I agree but I think that could be part of a larger change to move away from an editing editions boolean to a proper editing mode flag. We're currently littering the code base with `isEditingEditions` and I'd prefer this was a enum such as `EditingMode`

jonathonherbert: absolutely 👍

PR #841, opened 2019-07-16
Remove the notification client and directly use the models.

alexduf:
## What's changed?

This has the side effect of sending only one push notification for a global breaking news, which in turn might make the notification slightly faster. It also means all the editions will be processed at the same time, and should give editorial satisfaction as the UK won't be processed last.

## Implementation notes

It also introduces the parsing of the notification ID, in order to potentially poll the report API and check on the status, although nothing's done with it for now in order to limit my changes to small area.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/updates/BreakingNewsUpdate.scala
+++ b/app/updates/BreakingNewsUpdate.scala
@@ -2,8 +2,8 @@ package updates
 
 import java.net.URI
 import java.security.InvalidParameterException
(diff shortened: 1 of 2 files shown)


alexduf:
Tested on CODE copiously, I think it can go live

PR #842, opened 2019-07-16
Add an extremely temporary v2-prototype path for testing the inline form

jonathonherbert:
## What's changed?

Add an extremely temporary v2-prototype path for testing the inline form.

Ideally, we'd use a feature switch to change the UI, however it's been estimated to be about a weeks worth of effort to implement. We want to get feedback from a few Editors in their daily routine before shipping to 100% of users; CODE isn't a viable option as that would return limited feedback. In the absence of feature switches, this provides a short lived solution.

## Implementation notes

The approach we've taken is to provide a prototype client endpoint, and a prototype client build with its own configuration. We extract the prototype branch manually during CI/CD (there's no way to run git commands with the current teamcity setup), build it, and move it to the appropriate place in the root project. The v2-prototype endpoint then serves it as usual.

This approach has the advantage of being quick, and the disadvantage of being moderately cursed -- we're basically arbitrarily dumping a branch into master. To mitigate concerns here, we should protect the v2-prototype branch in the same way we do master -- e.g. only allow pull requests to be merged.

Finally, this adds around 2 minutes to the build time.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/config/webpack.config.prod-prototype.js
+++ b/client-v2/config/webpack.config.prod-prototype.js
@@ -1,10 +1,16 @@
 import prod from './webpack.config.prod';
 const path = require('path');
+const webpack = require('webpack');
+
+const definePlugin = new webpack.DefinePlugin({
+  'process.env.APP_URL_PREFIX': 'v2-prototype'
+});
 
 export default {
   ...prod,
   output: {
     path: path.resolve(__dirname, '../../public/client-v2/dist-prototype'),
     filename: 'app.bundle.js'
-  }
+  },
+  plugins: [definePlugin]
 };
--- a/client-v2/config/webpack.config.prod-prototype.js
+++ b/client-v2/config/webpack.config.prod-prototype.js
@@ -3,7 +3,7 @@ const path = require('path');
 const webpack = require('webpack');
 
 const definePlugin = new webpack.DefinePlugin({
-  'process.env.APP_URL_PREFIX': 'v2-prototype'
+  'process.env.APP_URL_PREFIX': JSON.stringify('v2-prototype')
 });
 
 export default {
--- a/client-v2/config/webpack.config.prod-prototype.js
+++ b/client-v2/config/webpack.config.prod-prototype.js
@@ -1,0 +1,10 @@
+import prod from './webpack.config.prod';
+const path = require('path');
+
+export default {
+  ...prod,
+  output: {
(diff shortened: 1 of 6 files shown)


client-v2/config/webpack.config.prod-prototype.js
```diff
@@ -0,0 +1,16 @@
+import prod from './webpack.config.prod';
+const path = require('path');
+const webpack = require('webpack');
+
+const definePlugin = new webpack.DefinePlugin({
+  'process.env.APP_URL_PREFIX': JSON.stringify('v2-prototype')
```
ajwl: So I guess we're also adding in a `.env` file to get this working here? Would it be more in keeping to put a new config variable in the s3 bucket? Was this approach quicker?

ajwl: Ah I guess this is what you're doing here? https://webpack.js.org/plugins/define-plugin/

jonathonherbert: @ajwl that's correct 👍

jonathonherbert:
@akash1810 thanks for the detail -- I've stolen it for the main PR text, I hope you don't mind!

scripts/teamcity-ci.sh
```diff
@@ -31,8 +31,20 @@ javascriptV2() {
     yarn lint
     yarn test
     yarn run build
-    yarn test-integration-ci
 
+    # Temporary steps to inject the client of the v2-prototype branch into the project
+    wget https://github.com/guardian/facia-tool/archive/v2-prototype.zip
+    unzip v2-prototype.zip
+    pushd facia-tool-2-prototype/client-v2
+    yarn
```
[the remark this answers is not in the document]
jonathonherbert: ooh, I thought the new branch depended on a different version of `redux-form`, but I've just checked and this isn't the case, so, yes! Brilliant idea.

sihil:
This crosses some lines in the sand for me, principally the fact that this changes the behaviour of the build to not be reproducable. I'm concerned that using this an approach full of hacks and kludges on a production system sets an extraordinarily bad precedent and abandons all pretext of engineering discipline. This is totally acceptable for a hack day

Would like to talk through alternative options prior to this being merged.

jonathonherbert:
We've chatted to @sihil in person, and these concerns are substantial enough that we're going to find another way to do this.

PR #843, opened 2019-07-16
DB testing framework

sihil:
## What's changed?
There is currently no testing around the serialisation and deserialisation of data into and out of the postgres DB. This PR adds a framework around the editions code and writes a couple of sample tests. Medium term we should aim to have 100% test coverage which will provide a high degree of confidence that we're not going to delete or corrupt our users' data.

## Implementation notes
This uses docker to spin up a postgres container that is torn down at the end of every test run. During each test class the same instance of postgres is used. 

I've introduced a UsesDatabase tag which controls when the tests are run. During a normal `test` run, tests with these tags are ignored. When run with `database-int:test` only these tests will be run.

## Checklist
 - [x] Check that the version of postgres being used is appropriate
 - [x] Configure TeamCity to run these tests sensibly

### General
- [X] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE


app/services/editions/db/IssueQueries.scala
```diff
@@ -13,7 +13,8 @@ trait IssueQueries {
   def insertIssue(
                    name: String,
                    skeleton: EditionsIssueSkeleton,
-                   user: User
+                   user: User,
+                   now: OffsetDateTime
```
akash1810: It feels a little odd to pass `now` as a param. Would renaming it to `date` or defaulting it to `DateTime.now` be better?

build.sbt
```diff
@@ -103,13 +103,16 @@ libraryDependencies ++= Seq(
...
     "com.typesafe.play" %% "play-json-joda" % "2.6.9",
 
     "org.postgresql"           %  "postgresql"                   % "42.2.5",
     "org.scalikejdbc"          %% "scalikejdbc"                  % "3.1.0",
     "org.scalikejdbc"          %% "scalikejdbc-config"           % "3.1.0",
-    "org.scalikejdbc"          %% "scalikejdbc-play-initializer" % "2.6.0-scalikejdbc-3.1"
+    "org.scalikejdbc"          %% "scalikejdbc-play-initializer" % "2.6.0-scalikejdbc-3.1",
+
+    "org.scalatest" %% "scalatest" % "3.0.5" % "test",
+    "com.whisk" %% "docker-testkit-scalatest" % "0.9.9" % "test",
```
akash1810: 😐 not a fan of launching containers this way.

test/services/editions/db/EditionsDBTest.scala
```diff
@@ -0,0 +1,143 @@
...
+      val issueDate = OffsetDateTime.ofInstant(Instant.ofEpochMilli(retrievedIssue.issueDate), ZoneId.of(retrievedIssue.timezoneId))
+      issueDate.getDayOfWeek shouldBe DayOfWeek.MONDAY
+      retrievedIssue.launchedOn.isDefined shouldBe false
+      retrievedIssue.launchedBy.isDefined shouldBe false
+      retrievedIssue.launchedEmail.isDefined shouldBe false
+      retrievedIssue.fronts shouldBe Nil
+    }
+
+    "should list the empty issue" taggedAs UsesDatabase in {
+      val issues = editionsDB.listIssues("daily-edition", LocalDate.of(2019, 9, 29), LocalDate.of(2019, 10, 3))
```
akash1810: Does this test depend on the previous one? Could they be independent?

test/services/editions/db/EditionsDBTest.scala
```diff
@@ -0,0 +1,143 @@
...
+                name = "opinion",
+                prefill = Some(CapiPrefillQuery("magic-opinion-query")),
+                presentation = CollectionPresentation(),
+                items = List(
+                  "54321",
+                  "65432"
+                )
+              ),
+              EditionsCollectionSkeleton(
+                name = "brexshit",
```
akash1810: 😂

app/services/editions/db/IssueQueries.scala
akash1810: Update: From reading the tests, this makes a lot of sense.

build.sbt
sihil: You said. Largely because it makes it hard to collaborate with other consumers using different CI or other reasons? This guy has an alternative approach: https://medium.com/@afsal.taj06/docker-compose-management-using-sbt-and-property-based-tests-bb8ae7df2759 happy to switch if it alleviates your concerns. I just don't want to think about spinning up, waiting for them to be ready and teaaring down containers.

test/services/editions/db/EditionsDBTest.scala
sihil: It does. It could. If we want them to be independent then we should redo evolutions each time to enforce this, which is very possible.

sihil: I've isolated the tests by running the evolutions each time (and rolling them back).

sihil:
This is ready for a further review now @akash1810

PR #846, opened 2019-07-17
Hide unused elements & show edition name/date

itsibitzi:
## What's changed?
You can now see what edition you're editing!

image

Also hides:
* Launch button on collections
* Warning in collection overview

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/EditionsController.scala
+++ b/app/controllers/EditionsController.scala
@@ -2,7 +2,7 @@ package controllers
 
 import logging.Logging
 import model.forms._
-import play.api.libs.json.Json
+import play.api.libs.json.{JsObject, Json}
 import services.editions.EditionsTemplating
 import java.time.LocalDate
 
@@ -36,6 +36,12 @@ class EditionsController(db: EditionsDB,
   def getIssue(id: String)= AccessAPIAuthAction { _ =>
     db.getIssue(id).map { issue =>
       Ok(Json.toJson(issue))
+    }.getOrElse(NotFound(s"Issue $id not found"))
+  }
+
+  def getIssueSummary(id: String)= AccessAPIAuthAction { _ =>
+    db.getIssueSummary(id).map { issue =>
+      Ok(Json.toJson(issue).as[JsObject] - "fronts")
     }.getOrElse(NotFound(s"Issue $id not found"))
   }
 
--- a/client-v2/src/components/FrontsEdit/FrontContainer.tsx
+++ b/client-v2/src/components/FrontsEdit/FrontContainer.tsx
@@ -19,7 +19,6 @@ import {
   selectFront,
   createSelectAlsoOnFronts
 } from 'selectors/frontsSelectors';
-import { selectIsEditingEditions } from 'selectors/pathSelectors';
 import Front from './Front';
 import SectionHeader from '../layout/SectionHeader';
 import SectionContent from '../layout/SectionContent';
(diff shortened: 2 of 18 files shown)


client-v2/src/components/FrontsEdit/FrontContainer.tsx
```diff
@@ -152,31 +159,33 @@ class Fronts extends React.Component<FrontsComponentProps, ComponentState> {
...
+                        key={key}
+                        name={`${this.props.frontId} - frontStages`}
+                        checked={frontStages[key] === this.state.collectionSet}
+                        onChange={() => this.handleCollectionSetSelect(key)}
+                        label={toTitleCase(frontStages[key])}
+                      />
+                    ))}
+                  </RadioGroup>
+                </StageSelectButtons>
+              )}
```
sihil: Can we have a `<EditionsOnly></EditionsOnly>` component (and inverse) that only renders when in the right state? That would reduce the boiler plate in most places I believe by moving the binding to redux inside the component?

itsibitzi: In #840 there's a comment that we should probably have a typed enum for the various modes the tool can be in, `fronts`, `editions`, etc.

Could possibly have a component like `<Visibility mode="fronts"> ... </Visibility>`? 

I think this should be done as a single, separate, PR, sooner rather than later though!

itsibitzi: Though this only solves our issues for rendering - if there's logic based on the mode, such as fetching the issue information on mount, then that'll have to be handled separately.

itsibitzi:
Made a new endpoint which only returns a summary of a given issue. Keeping the old endpoint because it's handy for debugging, but we can happily remove it later. Chose not to create another case class for the summary version, instead opting to use the `EditionsIssue` remove the fronts list from the JSON in the controller - perhaps sketch?

Added a new component called `EditModeVisibility` which can hide based on the `editMode` state.

Could possibly turn this into `EditionsOnly` and `FrontsOnly` but maybe having a single one would make refactoring in the future easier?

akash1810:
This is quite a big diff, could we break it down? I think we'd at least want to ship the new endpoint separately from the client-side changes to guarantee the endpoint exists for the new client.

client-v2/src/components/util/EditModeVisibility.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { State } from 'types/State';
+import { connect } from 'react-redux';
+
+interface EditModeVisibilityProps {
+  currentMode: EditMode;
+  visibleMode: EditMode;
+  children: ReactNode;
+}
+
+const EditModeVisibility = ({
```
akash1810: This component feels useful for the feature switching work @jonathonherbert and @jennygrahamjones are doing for inline editing.

client-v2/src/types/EditMode.ts
```diff
@@ -0,0 +1 @@
+export type EditMode = 'fronts' | 'editions';
```
akash1810: we might want to export `fronts` and `editions` as strings to remove the magic strings above?

app/controllers/EditionsController.scala
```diff
@@ -39,6 +39,12 @@ class EditionsController(db: EditionsDB,
     }.getOrElse(NotFound(s"Issue $id not found"))
   }
 
+  def getIssueSummary(id: String)= AccessAPIAuthAction { _ =>
+    db.getIssueSummary(id).map { issue =>
+      Ok(Json.toJson(issue).as[JsObject] - "fronts")
```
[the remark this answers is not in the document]
itsibitzi: So we can remove the empty list of fronts! I initially had a `EditionsIssueSummary` case class but I think the overhead isn't worth while when really a summary is just a view over the top of the regular `EditionsIssue`

PR #849, opened 2019-07-18
Add preview bucket and update contents when collections change

sihil:
## What's changed?

This provides an initial cut of preview capabilities for editions. Everytime any change is made we write a full copy of the current issue into a new `preview-editions` bucket under `<edition-name>/<issue-date>/preview.json`.

The idea then is that the editions app can use the keys in that bucket as a source of data for populating the preview mode of the app.

## Implementation notes

At present this synchronously queries the DB and writes to the bucket on every single collection update. It would be better to do this asynchronously and debounce it so it is re-published no more frequently than once every couple of seconds.

## Checklist
- [ ] Ensure https://github.com/guardian/editorial-tools-platform/pull/304 is applied and merged

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/services/editions/publishing/PublishedIssuesBucket.scala
+++ b/app/services/editions/publishing/PublishedIssuesBucket.scala
@@ -4,12 +4,9 @@ import java.time.OffsetDateTime
 
 import com.amazonaws.services.s3.AmazonS3
 import com.gu.editions.PublishedIssue
-import com.gu.pandomainauth.model.User
 
-class PublishedIssuesBucket(s3Client: AmazonS3, bucketName: String) extends EditionsBucket[User](s3Client, bucketName) {
-  override def createIssueFilename(issue: PublishedIssue, publicationMetadata: User): String = {
-    val firstName = publicationMetadata.firstName.replace(" ", "_")
-    val lastName = publicationMetadata.lastName.replace(" ", "_")
-    s"${OffsetDateTime.now().toString}_${firstName}_$lastName.json"
+class PublishedIssuesBucket(s3Client: AmazonS3, bucketName: String) extends EditionsBucket(s3Client, bucketName) {
+  override def createIssueFilename(issue: PublishedIssue): String = {
+    s"${OffsetDateTime.now()}.json"
   }
 }
--- a/app/services/editions/publishing/PublishedIssuesBucket.scala
+++ b/app/services/editions/publishing/PublishedIssuesBucket.scala
@@ -2,31 +2,14 @@ package services.editions.publishing
 
 import java.time.OffsetDateTime
 
-import com.amazonaws.services.s3.AmazonS3ClientBuilder
+import com.amazonaws.services.s3.AmazonS3
 import com.gu.editions.PublishedIssue
-import conf.ApplicationConfiguration
-import play.api.libs.json.Json
-import services._
-import PublishedIssueFormatters._
 import com.gu.pandomainauth.model.User
 
-class PublishedIssuesBucket(config: ApplicationConfiguration, awsEndpoints: AwsEndpoints) {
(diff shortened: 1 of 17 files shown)


app/services/editions/publishing/PublishedIssuesBucket.scala
```diff
@@ -2,31 +2,14 @@ package services.editions.publishing
...
-    s"${issue.name}/${issue.issueDate.toLocalDate.toString}/${OffsetDateTime.now().toString}_${firstName}_$lastName.json"
-  }
-
-  def putIssue(issue: PublishedIssue, user: User) = {
-    client.putObject(bucket, createIssuePath(issue, user), Json.stringify(Json.toJson(issue)))
+class PublishedIssuesBucket(s3Client: AmazonS3, bucketName: String) extends EditionsBucket[User](s3Client, bucketName) {
+  override def createIssueFilename(issue: PublishedIssue, publicationMetadata: User): String = {
+    val firstName = publicationMetadata.firstName.replace(" ", "_")
+    val lastName = publicationMetadata.lastName.replace(" ", "_")
+    s"${OffsetDateTime.now().toString}_${firstName}_$lastName.json"
```
akash1810: do we need the `firstName` and `lastName` on the filename?

test/services/editions/db/EditionsDBTest.scala
```diff
@@ -160,6 +162,46 @@ class EditionsDBTest extends FreeSpec with Matchers with EditionsDBService with
...
+          val now = OffsetDateTime.of(2019, 7, 16, 17, 51, 44, 0, ZoneOffset.ofHours(1))
+
+          val id = editionsDB.insertIssue("daily-edition",
+            skeleton,
+            User("Billy", "Bragg", "billy.bragg@justice.example.com", None),
+            now
+          )
+
+          val retrievedIssue = editionsDB.getIssue(id).value
+          val collectionsId = retrievedIssue.fronts.head.collections.head.id
```
akash1810: do we want to test `retrievedIssue.fronts.length == 1`?

sihil: That case is covered by other tests so I don't think it is necessary. What I have done though is inserted a couple of extra issues that shouldn't be selected so we know that this isn't working by virtue of returning the single issue that exists in the DB.

app/services/editions/publishing/PublishedIssuesBucket.scala
sihil: Killed - need to check with @AWare prior to merging but fairly sure he doesn't rely on the user in the file name.

app/services/editions/publishing/PublishedIssuesBucket.scala
```diff
@@ -2,31 +2,11 @@ package services.editions.publishing
...
-  val client = AmazonS3ClientBuilder
-    .standard()
-    .withCredentials(config.aws.credentials.get)
-    .withRegion(config.aws.region)
-    .build()
-
-  private def createIssuePath(issue: PublishedIssue, user: User) = {
-    val firstName = user.firstName.replace(" ", "_")
-    val lastName = user.lastName.replace(" ", "_")
-    s"${issue.name}/${issue.issueDate.toLocalDate.toString}/${OffsetDateTime.now().toString}_${firstName}_$lastName.json"
```
akash1810: 🎉

app/services/editions/publishing/PublishedIssuesBucket.scala
sihil: @AWare is on holiday till Thurs so I'm going to merge and seek forgiveness later...

PR #856, opened 2019-07-23
Update edition front names and insert 'the fashion' in the running order

blishen:
## What's changed?
The names of fronts has been made human readable, and are what actually gets rendered on the app. Note that while some fronts share names (and did previously too), we never show two fronts with the same name on the same day.

In addition, I noticed we defined, but never used the fashion magazine front, so I've added to Saturday, where it will belong when we actually produce it.

## Implementation notes
The fronts names are already used to pick the appropriate swatch to apply by the Editions back end. I've given Rich a snippet that supports both old and new names. A better solution would be for the fronts to pass in presentation properties (that exist in the model) that define which swatch of colours to use when rendering. This would simplify the backend and make everything more easily configurable.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/editions/templates/DailyEdition.scala
+++ b/app/model/editions/templates/DailyEdition.scala
@@ -432,7 +432,7 @@ object FrontLifeFashion {
     hidden = true
   )
   val front = FrontTemplate(
-    name = "Lifestyle",
+    name = "The Fashion",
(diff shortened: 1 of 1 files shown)


sihil:
Whoop! Thanks for setting that up @akash1810

PR #859, opened 2019-07-24
Add Publish Issue button

akash1810:
## What's changed?
Adds a Publish button to an Edition Issue to...publish.

## Implementation notes
There is no feedback to the user atm.

### Clipboard opened
image

### Clipboard closed
image

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FeedSectionHeader.tsx
+++ b/client-v2/src/components/FeedSectionHeader.tsx
@@ -23,6 +23,7 @@ import { getEditionIssue } from 'actions/Editions';
 import { EditionsIssue } from 'types/Edition';
 import { selectors as editionsIssueSelectors } from 'bundles/editionsIssueBundle';
 import { EditMode } from 'types/EditMode';
+import EditModeVisibility from './util/EditModeVisibility';
 
 const FeedbackButton = Button.extend<{
   href: string;
@@ -99,11 +100,11 @@ const CloseButtonInner = styled.div`
 
 const EditionIssueInfo = styled.div`
   height: 100%;
-  display: flex;
+  display: inline-block;
   flex-direction: column;
   justify-content: center;
   margin-left: 12px;
-  line-height: initial;
+  line-height: 1em;
 `;
 
 const EditionTitle = styled.div`
@@ -112,6 +113,11 @@ const EditionTitle = styled.div`
 
 const EditionDate = styled.div`
   font-size: 16px;
+`;
+
+const EditionPublish = styled.div`
+  float: right;
+  margin: 5px 10px 0 0;
 `;
 
 type ComponentProps = {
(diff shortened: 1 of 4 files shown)


client-v2/src/components/FeedSectionHeader.tsx
```diff
@@ -114,6 +115,11 @@ const EditionDate = styled.div`
   font-size: 16px;
 `;
 
+const EditionPublish = styled.div`
+  float: right;
```
akash1810: I'm being lazy here...

client-v2/src/services/editionsApi.ts
```diff
@@ -71,3 +71,13 @@ export const getIssueSummary = async (id: string): Promise<EditionsIssue> => {
...
   });
 };
+
+export const publishIssue = async (id: string): Promise<EditionsIssue> => {
+  return pandaFetch(`/editions-api/issues/${id}/publish`, {
+    method: 'POST',
+    credentials: 'same-origin'
+  }).then(response => {
+    // TODO publish endpoint returns a 204 No Content, we probably want something...
+    return response.json();
```
sihil: perhaps the newly published hash (based on our discussions about tracking changes)? regardless - this certainly doesn't return an `EditionsIssue`.

client-v2/src/components/FeedSectionHeader.tsx
```diff
@@ -169,16 +175,7 @@ class FeedSectionHeader extends Component<ComponentProps> {
...
-                <EditionIssueInfo>
-                  <EditionTitle>
-                    {startCase(editionsIssue.displayName)}
-                  </EditionTitle>
-                  <EditionDate>
-                    {new Date(editionsIssue.issueDate).toDateString()}
-                  </EditionDate>
-                </EditionIssueInfo>
-              ) : null
+              this.renderEditionsActions()
```
sihil: Is this is a pattern that we're using elsewhere in facia?

client-v2/src/components/FeedSectionHeader.tsx
```diff
@@ -192,6 +189,38 @@ class FeedSectionHeader extends Component<ComponentProps> {
...
+          <EditionDate>
+            {new Date(editionsIssue.issueDate).toDateString()}
+          </EditionDate>
+        </EditionIssueInfo>
+        <EditionPublish>
+          <EditModeVisibility visibleMode="editions">
+            <Button
+              size="l"
+              priority="primary"
+              onClick={() => publishEdition(editionsIssue.id)}
```
sihil: nitpick: name better as `publishEditionIssue`?

client-v2/src/components/FeedSectionHeader.tsx
akash1810: I don't think so, however pushing Edition specific rendering into its own function makes things easier to read.

sihil: I agree it makes it easier to read locally but it's mixing styles. Looks like this is done elsewhere with nested components. What's your opinion of this style @jonathonherbert ?

akash1810:
As discussed IRL we probably want a to use `startConfirmModal` here, however that requires a bit of a refactor - are we happy to ship this as is (no user feedback) and iterate?

sihil:
:shipit: but please add a new card to the should column.

PR #860, opened 2019-07-24
Add inline editing behind a feature switch

jennygrahamjones:
## What's changed?

This PR introduces the ability to use inline form editing. Users can now toggle (see #848) between side editing (which allows a single form to be edited) and inline editing (which allows multiple forms to be edited). This is a global change, across all fronts and the clipboard. 

The PR makes no changes to the existing side form editing functionality, ensuring no impact to users who are not involved in testing inline editing.

## Implementation notes

To allow both forms to co-exist, the `ArticleFragmentForm` has been copied completely into the `ArticleFragmentFormInline`. Likewise, `InputCheckboxToggleInline` largely duplicates the existing `InputCheckboxToggle`. 

There are some user experience / usability elements of the inline form that may require some tweaking, such as the layout of the inline form as width increases and decreases. 
 
Additionally, cutout and slideshow toggles for images are not available on inline forms. These features will be enabled in another PR to follow.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/components/Clipboard.tsx
```diff
@@ -273,7 +269,12 @@ const mergeProps = (
   ...stateProps,
   ...dispatchProps,
   selectArticleFragment: (articleId: string, isSupporting = false) =>
-    dispatchProps.selectArticleFragment(clipboardId, articleId, isSupporting)
+    dispatchProps.selectArticleFragment(
+      articleId,
+      clipboardId,
+      clipboardId,
```
ajwl: Just add a comment to explain passing `clipboardId` twice

client-v2/src/components/Row.tsx
```diff
@@ -1,8 +1,10 @@
 import { styled } from 'constants/theme';
 
-const Row = styled('div')`
+const Row = styled('div')<{ flexDirection?: string; gutter?: number }>`
   display: flex;
   margin: 0 ${({ gutter = 10 }: { gutter?: number }) => `${-(gutter / 2)}px`};
+  flex-direction: ${({ flexDirection }) =>
+    flexDirection ? flexDirection : ''};
```
ajwl: 👍

client-v2/src/components/Clipboard.tsx
jennygrahamjones: Done 👍

client-v2/src/components/Row.tsx
jennygrahamjones: 💪

PR #861, opened 2019-07-24
"Suggested Articles" prefill button

itsibitzi:
## What's changed?

Add the ability to fetch suggested articles from CAPI. This required quite a bit of reworking the feed component in the frontend so is a little large.

## How does this work?

When you're viewing a collection which has a prefill available you'll see a `Suggested Articles` button. Click it and the feed will be replaced with a list of articles which match the prefill query and aren't currently in the collection.

From here you can click/drag them in as before!

image

## Implementation notes

Some things I'm not mega happy with which would be good to discuss:
* Use of `newspaperEditionDate` as an alternative to `isLive` isn't very semantically clear.
* Rather clowny adding of `"{\"response\":" + json + "}" to work around the fact the CAPI client is slightly different to the actual HTTP response from CAPI.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/EditionsController.scala
+++ b/app/controllers/EditionsController.scala
@@ -107,6 +107,7 @@ class EditionsController(db: EditionsDB,
   def getPrefillForCollection(id: String) = AccessAPIAuthAction.async { req =>
     db.getCollectionPrefillQueryString(id).map { prefillUpdate =>
       capi.getPrefillArticles(prefillUpdate.issueDate, prefillUpdate.prefill, prefillUpdate.currentPageCodes).map { body =>
+        // Need to wrap this in a "response" object because the CAPI client and CAPI API return different JSON
         val json = "{\"response\":" + body.asJson.noSpaces + "}"
         val decorated = rewriteBody(json)
         Ok(decorated).as("application/json")
--- a/app/controllers/EditionsController.scala
+++ b/app/controllers/EditionsController.scala
@@ -7,16 +7,21 @@ import services.editions.EditionsTemplating
 import java.time.{LocalDate, OffsetDateTime}
 
 import model.editions.{EditionsFrontendCollectionWrapper, EditionsTemplates}
+import services.Capi
 import services.editions.db.EditionsDB
-import services.editions.publishing.{EditionsPublishing, PublishedIssuesBucket}
+import services.editions.publishing.EditionsPublishing
 import services.editions.publishing.PublishedIssueFormatters._
+import util.ContentUpgrade.rewriteBody
+import com.gu.contentapi.json.CirceEncoders._
+import io.circe.syntax._
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 
 class EditionsController(db: EditionsDB,
                          templating: EditionsTemplating,
                          publishing: EditionsPublishing,
+                         capi: Capi,
                          val deps: BaseFaciaControllerComponents)(implicit ec: ExecutionContext) extends BaseFaciaController(deps)  with Logging {
 
   def postIssue(name: String) = AccessAPIAuthAction(parse.json[CreateIssue]) { req =>
@@ -97,5 +102,15 @@ class EditionsController(db: EditionsDB,
       publishing.publish(issue, req.user, OffsetDateTime.now())
       NoContent
     }.getOrElse(NotFound(s"Issue $id not found"))
+  }
+
+  def getPrefillForCollection(id: String) = AccessAPIAuthAction.async { req =>
+    db.getCollectionPrefillQueryString(id).map { prefillUpdate =>
+      capi.getPrefillArticles(prefillUpdate.issueDate, prefillUpdate.prefill, prefillUpdate.currentPageCodes).map { body =>
+        val json = "{\"response\":" + body.asJson.noSpaces + "}"
+        val decorated = rewriteBody(json)
+        Ok(decorated).as("application/json")
+      }
+    }.getOrElse(Future.successful(NotFound("Collection not found")))
   }
 }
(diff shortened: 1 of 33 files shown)


client-v2/src/components/FeedsContainer.tsx
```diff
@@ -26,18 +28,28 @@ import { DEFAULT_PARAMS } from 'services/faciaApi';
...
   liveLoading: boolean;
   previewLoading: boolean;
+  prefillLoading: boolean;
   liveError: string | null;
   previewError: string | null;
+  prefillError: string | null;
   livePagination: IPagination | null;
   previewPagination: IPagination | null;
+  prefillPagination: IPagination | null;
 }
```
jonathonherbert: Can we implement this in such a way that our component doesn't have to reason about the source of its data?

(I know that you're simply extending a pattern that's already being used -- this was not ideal before! _But_ it seems like a good opportunity to clean this up 😁)

client-v2/src/components/FeedsContainer.tsx
```diff
@@ -239,93 +272,124 @@ class FeedsContainer extends React.Component<
...
   };
 
   public render() {
     const {
+      isPrefillMode,
       liveArticles,
       previewArticles,
       liveError,
-      previewError
+      previewError,
```
jonathonherbert: As above re: reasoning about the source of this stuff. It'd be cleaner to select this out if we could.

client-v2/src/bundles/capiFeedBundle.ts
```diff
@@ -113,11 +115,58 @@ export const fetchPreview = (
...
+} = createAsyncResourceBundle<FeedState>('prefillFeed', {
+  indexById: false,
+  selectLocalState: state => state.feed.prefillFeed,
+  initialData: []
+});
+
+export const fetchPrefill = (
+  id: string
+): ThunkResult<void> => async dispatch => {
+  dispatch({
```
jonathonherbert: We should make this an action creator! Then we can derive the type using ReturnType<typeof actionCreator> and reduce the number of duplicated strings (see e.g. types/Action)

client-v2/src/bundles/capiFeedBundle.ts
```diff
@@ -113,11 +115,58 @@ export const fetchPreview = (
...
+        )
+      );
+    }
+  } catch (e) {
+    dispatch(prefillActions.fetchError(e.message));
+  }
+};
+
+export const hidePrefills = () => (dispatch: Dispatch) => {
+  dispatch({
```
jonathonherbert: As above re: action creator

jonathonherbert:
Really happy we've gone down this route, I feel like it'll be easy for editors to use. A few comments, mostly to see if we can correct patterns that historically haven't been great and are now being extended -- super happy to discuss/pair on any of the above to get things out the door.

akash1810:
This is failing CI and needs rebasing.

app/controllers/EditionsController.scala
```diff
@@ -98,4 +103,14 @@ class EditionsController(db: EditionsDB,
       NoContent
     }.getOrElse(NotFound(s"Issue $id not found"))
   }
+
+  def getPrefillForCollection(id: String) = AccessAPIAuthAction.async { req =>
+    db.getCollectionPrefillQueryString(id).map { prefillUpdate =>
+      capi.getPrefillArticles(prefillUpdate.issueDate, prefillUpdate.prefill, prefillUpdate.currentPageCodes).map { body =>
+        val json = "{\"response\":" + body.asJson.noSpaces + "}"
```
akash1810: why do we need to do this?

app/services/Capi.scala
```diff
@@ -88,13 +88,65 @@ class GuardianCapi(config: ApplicationConfiguration)(implicit ex: ExecutionConte
     query
   }
 
+  // Sadly there's no easy way of converting a CAPI client response into JSON so we'll just proxy - similar to controllers.FaciaContentApiProxy
+  def getPrefillArticles(issueDate: ZonedDateTime, capiPrefillQuery: CapiPrefillQuery, currentPageCodes: List[String]): Future[SearchResponse] = {
+    val fields = List(
+      "newspaperEditionDate",
```
akash1810: should these be kebab-cased?

client-v2/src/selectors/frontsSelectors.ts
```diff
@@ -122,7 +122,10 @@ const getCollections = (state: State): CollectionConfigMap =>
   frontsConfigSelectors.selectAll(state).collections || {};
 
 const selectCollectionConfig = (state: State, id: string): CollectionConfig =>
-  getCollections(state)[id] || null;
+  getCollections(state)[id];
+
+const selectCollectionHasPrefill = (state: State, id: string): boolean =>
+  !!(selectCollectionConfig(state, id) || { prefill: undefined }).prefill;
```
akash1810: can't tell if I like this pattern or not...

client-v2/src/types/Capi.ts
```diff
@@ -98,6 +98,7 @@ interface CapiArticleFields {
   liveBloggingNow?: CapiBool;
   shortUrl?: string;
   membershipUrl?: string;
+  newspaperEditionDate?: string;
```
akash1810: could this be a `Date` type?

app/controllers/EditionsController.scala
itsibitzi: The CAPI client and the CAPI API return different objects :(

app/services/Capi.scala
itsibitzi: They were previously kebab-case and *seemed* to work, but the docs use camelCase and it also *seems* to work

client-v2/src/types/Capi.ts
itsibitzi: After a chat - I'll add a new field to `frontsMeta` which flags `isPrintSent` to allow us to display draft content as if it was live in the editions version of the tool.

itsibitzi:
Seems happy on CODE!

Added some integration tests as well :)

app/services/Capi.scala
itsibitzi: Except in the `useDate` case, which MUST be kebab case lols.

PR #870, opened 2019-07-25
Update FeedSectionHeader.tsx

blishen:
## What's changed?

Link back to the edition picker from the date
 
## Implementation notes

Errrr, I kind of guessed the code.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FeedSectionHeader.tsx
+++ b/client-v2/src/components/FeedSectionHeader.tsx
@@ -170,7 +170,7 @@ class FeedSectionHeader extends Component<ComponentProps> {
           <FadeTransition active={!isCurrentFrontsMenuOpen} direction="right">
             {editMode === 'editions' ? (
               editionsIssue ? (
-                <Link to='/v2/manage-editions/daily-edition'>
+                <Link to="/v2/manage-editions/daily-edition">
                   <EditionIssueInfo>
                     <EditionTitle>
                       {startCase(editionsIssue.displayName)}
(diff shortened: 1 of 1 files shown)


blishen:
I suspect linting is breaking this is teamcity, but don't suppose someone could fix and push if this doesn't look too mental. Just a way to get back to the edition management from within an edition.

client-v2/src/components/FeedSectionHeader.tsx
```diff
@@ -170,14 +170,16 @@ class FeedSectionHeader extends Component<ComponentProps> {
...
               editionsIssue ? (
-                <EditionIssueInfo>
-                  <EditionTitle>
-                    {startCase(editionsIssue.displayName)}
-                  </EditionTitle>
-                  <EditionDate>
-                    {new Date(editionsIssue.issueDate).toDateString()}
-                  </EditionDate>
-                </EditionIssueInfo>
+                <Link to="/v2/manage-editions/daily-edition">
```
akash1810: I suspect we'd want this to link to the specific edition rather than the Daily? Doesn't look like this is on the EditionsIssue type though - https://github.com/guardian/facia-tool/blob/master/client-v2/src/types/Edition.ts#L24-L35 so we'd need to add it.

blishen: The idea was just to be able to navigate back to the date picker from within the tool. It's a bit unnatural to have to go to the bookmark.

akash1810: 👍 I presume we don't always want to go back to daily-edition though, as we'll have multiple additions?

Let's ship this though and update later if needed.

blishen: Ah - I understand your point now. But agree.

PR #872, opened 2019-07-25
DRAG AND DROP: Make image dragging more obvious 

ajwl:
## What's changed?

When dragging and dropping images - it was not always obvious on the UI where the new image was being dropped. 

Occasionally this led to pictures getting put in weird places. 

This PR does not change the functionality of image drag and drop but just makes the visual feedback clearer. 

BEFORE 
Screenshot 2019-07-26 at 11 21 04

AFTER 
Screenshot 2019-07-26 at 11 25 47

## Implementation notes

Because there were 2 different ways of getting thumbnail image from state, I have deleted one. One using the general article and a specific selector for activeImageUrl. 

We have opted to always go for the general article one. And have deleted the other selector

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/components/Thumbnail.tsx
+++ b/client-v2/src/shared/components/Thumbnail.tsx
@@ -1,38 +1,33 @@
-import { styled } from 'shared/constants/theme';
-import { theme as theme2 } from 'constants/theme';
+import { theme, styled } from 'constants/theme';
 
 const ThumbnailBase = styled('div')`
   background-size: cover;
-  background-color: ${({ theme }) =>
-    theme.shared.base.colors.backgroundColorFocused};
+  background-color: ${theme.shared.base.colors.backgroundColorFocused};
 `;
 
 const ThumbnailSmall = styled(ThumbnailBase)<{
-  url?: string;
+  url?: string | void;
   isDraggingImageOver?: boolean;
   imageHide?: boolean;
 }>`
   position: relative;
-  width: 83px;
-  min-width: ${theme2.shared.thumbnailImage.width};
-  height: ${theme2.shared.thumbnailImage.height};
+  width: ${theme.shared.thumbnailImage.width};
+  min-width: ${theme.shared.thumbnailImage.width};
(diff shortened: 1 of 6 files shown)


client-v2/src/shared/components/Thumbnail.tsx
```diff
@@ -1,16 +1,38 @@
 import { styled } from 'shared/constants/theme';
+import { theme as theme2 } from 'constants/theme';
```
ajwl: @jonathonherbert what is the right way to do this? I constantly get `'constants/theme'` and `'shared/constants/theme'` mixed up. Can we consolidate them for simplicity?

client-v2/src/shared/components/article/ArticleBody.tsx
```diff
@@ -241,10 +244,9 @@ const articleBodyDefault = React.memo(
           ) : (
             <DraggableArticleImageContainer id={uuid} canDrag={canDragImage}>
               <ThumbnailSmall
-                style={{
-                  backgroundImage: `url('${thumbnail}')`,
-                  opacity: imageHide ? 0.5 : 1
-                }}
+                imageHide={imageHide}
+                url={thumbnail}
+                isDraggingImageOver={isDraggingImageOver}
```
ajwl: Extracting the other two variables here (thumbnail and imageHide) so we can get rid of this inline style

jonathonherbert: 👍

client-v2/src/shared/selectors/collectionItem.ts
```diff
@@ -17,33 +17,6 @@ const createSelectCollectionItemType = () =>
     }
   );
 
-const createSelectActiveImageUrl = () =>
```
ajwl: This was only being used in one place.... and it was creating problems because it occasionally produced different results than the simple `externalArticle.fields.thumbnail`. Articles with overridden images behaved in unpredictable ways.

PR #873, opened 2019-07-25
Trust client timestamp and retrofit collection DB tests

sihil:
## What's changed?

This fixes the issue that we were seeing where the following message was being seen when editing content. It typically occurred when inside an edit form when the next poll of colleciton updates were done.

unnamed

The fix has been to trust the clients timestamp and use that in the database rather than the server timestamp. This is a bit of a kludge but any other fix requires a significant overhaul of various bits of the frontend and the introduction of article level ids on the backend.

I highly recommend reviewing this using the individual commits - the actual fix is in the last commit.

## Implementation notes

The original fix was to ensure that the endpoint returned the updated collection read back from the DB (rather than parrot back the one sent by the client). This has been left in as the 'right' behaviour even though we don't rely on it.

Whilst implementing this I have simplified the way with deal with timestamps. We now use millisecond resolution timestamps when saving to the database. Timestamps coming from the client use this resolution anyway but when the server is creating a timestamp this might be an nanosecond resolution. At all points where we having incoming timestamps into a DB endpoint they must be truncated to millis so that queries make sense. This has allowed the rounding logic to be safely removed.

Additionally this adds tests for more of the database interface to increase coverage to the area I've touched in this PR.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/services/editions/db/CollectionsQueries.scala
+++ b/app/services/editions/db/CollectionsQueries.scala
@@ -7,12 +7,11 @@ import model.forms.GetCollectionsFilter
 import play.api.libs.json.Json
 import scalikejdbc._
 import services.editions.DbEditionsArticle
-import services.editions.CollectionsHelpers._
 
 trait CollectionsQueries {
+
   def getCollections(filters : List[GetCollectionsFilter]) = DB readOnly { implicit session =>
     case class TypedFilters(id: String, updatedAt: Option[OffsetDateTime])
-    case class GetCollectionsRow(collection: EditionsCollection, article: Option[DbEditionsArticle])
 
     val sqlFilters = filters.map { f =>
       TypedFilters(
@@ -23,51 +22,15 @@ trait CollectionsQueries {
       )
     }
 
-    val rows = sql"""
-      SELECT
-        collections.id,
-        collections.front_id,
-        collections.index,
-        collections.name,
-        collections.is_hidden,
-        collections.metadata,
-        collections.updated_on,
-        collections.updated_by,
-        collections.updated_email,
-        collections.prefill,
-
-        articles.collection_id AS articles_collection_id,
-        articles.page_code     AS articles_page_code,
-        articles.index         AS articles_index,
-        articles.added_on      AS articles_added_on,
-        articles.metadata      AS articles_metadata
-
-      FROM collections
-      LEFT JOIN articles ON (articles.collection_id = collections.id)
-      WHERE EXISTS (
+    val rows = fetchCollectionsSql(where = sqls"""
+      EXISTS (
         SELECT *
         FROM (VALUES ${sqlFilters.map(f => sqls"(${f.id}, ${f.updatedAt})")}) AS F(id, updated_on)
         WHERE collections.id = F.id AND (F.updated_on IS NULL OR collections.updated_on > F.updated_on::TIMESTAMPTZ)
       )
-      """.map { rs =>
-      GetCollectionsRow(
-        EditionsCollection.fromRow(rs),
-        DbEditionsArticle.fromRowOpt(rs, "articles_"))
-    }
-      .list()
-      .apply()
+    """).apply()
 
-    rows
-      .map(_.collection)
-      .distinctBy(_.id)
-      .map { collection =>
-          val articles = rows
-            .flatMap(_.article)
-            .filter(_.collectionId == collection.id)
-            .sortBy(_.index)
-            .map(_.article)
-        collection.copy(items = articles)
-      }
+    convertRowsToCollections(rows)
   }
 
   def updateCollection(collection: EditionsCollection, now: OffsetDateTime): EditionsCollection  = DB localTx { implicit session =>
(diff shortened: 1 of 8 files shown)


app/services/editions/db/CollectionsQueries.scala
```diff
@@ -93,17 +50,67 @@ trait CollectionsQueries {
...
        """.execute().apply()
     }
 
-    collection
+    val rows = fetchCollectionsSql(where = sqls"collections.id = ${collection.id}").apply()
+
+    val updatedCollections = convertRowsToCollections(rows)
+
+    // we have filtered on a single id so this list should only contain one collection
+    assert(updatedCollections.size == 1)
```
jennygrahamjones: If/when this throws an exception, is it handled anywhere? What should happen if this isn't true??

itsibitzi: Worst case scenario is the play framework will turn it into a 500 - which is probably appropriate here.

sihil: Using exceptions isn't ideal here but the `facia-tool` backend isn't setup to have proper typed error handling done at the edge and it's work that we are unable to prioritise. As @itsibitzi says though, play will turn this into a 500. I will however add an actual string message to this statement to make it easier to track down.

test/fixtures/EditionsDBService.scala
```diff
@@ -24,5 +24,11 @@ trait EditionsDBService extends DockerTestKit with DockerKitDockerJava with Dock
...
   }
 
-  def withEvolutions[T]: (=> T) => T = Evolutions.withEvolutions[T](database)
+  before {
+    Evolutions.applyEvolutions(database)
+  }
+
+  after {
+    Evolutions.cleanupEvolutions(database)
+  }
```
jonathonherbert: `withEvolutions` seems to take care of cleanup -- is this just to remove the need to nest test code in specs?

test/services/editions/db/EditionsDBTest.scala
```diff
@@ -165,6 +165,37 @@ class EditionsDBTest extends FreeSpec with Matchers with EditionsDBService with
...
+        collectionIds.map(GetCollectionsFilter(_, Some(now.toInstant.toEpochMilli-1)))
+      )
+      newerCollections.size shouldBe 5
+
+      // pretend we are a client asking for updates since a time that is more recent than the creation time
+      val olderCollections = editionsDB.getCollections(
+        collectionIds.map(GetCollectionsFilter(_, Some(now.toInstant.toEpochMilli+1)))
+      )
+      olderCollections.size shouldBe 0
+    }
```
jonathonherbert: 👍

jonathonherbert:
This looks good to me. I'll add that the judicious commit history made reviewing this PR a complete pleasure 👌

test/fixtures/EditionsDBService.scala
sihil: This was a suggestion from @akash1810 and I also spent ages trying to figure out why my test failed (I'd forgotten to put it in `withEvolutions`) so I moved from `withEvolutions` to this. Pro: can't forget evolutions; Con: can't not do evolutions if you want to mix DB and non-DB tests.

app/services/editions/db/CollectionsQueries.scala
```diff
@@ -93,17 +50,67 @@ trait CollectionsQueries {
...
        """.execute().apply()
     }
 
-    collection
+    val rows = fetchCollectionsSql(where = sqls"collections.id = ${collection.id}").apply()
+
+    val updatedCollections = convertRowsToCollections(rows)
+
+    // we have filtered on a single id so this list should only contain one collection
+    assert(updatedCollections.size == 1, s"Retrieved ${updatedCollections.size} collections from DB but there should be exactly one. Failing fast.")
```
akash1810: would pushing this into `convertRowsToCollections` be a cleaner pattern?

test/fixtures/EditionsDBService.scala
akash1810: should these be `beforeEach` `afterEach`?

test/services/editions/db/EditionsDBTest.scala
```diff
@@ -34,101 +48,200 @@ class EditionsDBTest extends FreeSpec with Matchers with EditionsDBService with
...
+
+      val future = now.plusMinutes(20)
+      val futureMillis = future.toInstant.toEpochMilli
+
+      val items = EditionsArticle("654789", futureMillis, Some(simpleMetadata)) :: brexshit.items
+
+      val evenMoreBrexshit = brexshit.copy(
+        lastUpdated = Some(futureMillis),
+        updatedBy = Some("BoJo"),
+        updatedEmail = Some("bojo@piffle.paffle"),
```
akash1810: 😂

app/services/editions/db/CollectionsQueries.scala
sihil: `convertRowsToCollections` can legitimately return multiple collections (see other call site of the function).

test/fixtures/EditionsDBService.scala
sihil: `before` and `after` do happen on eash test. As opposed to `beforeAll` and `afterAll` that happen per suite. See http://doc.scalatest.org/1.8/org/scalatest/BeforeAndAfter.html

sihil:
TeamCity is currently broken so this hasn't built or shipped.

PR #875, opened 2019-07-26
Expand clipboard when it contains open forms

jonathonherbert:
## What's changed?

The clipboard now expands when any of its cards are in form mode. Editing card meta in the clipboard is now easier as a result, and we can get rid of clipboard-specific styling.

clipboard-form

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/Clipboard.tsx
+++ b/client-v2/src/components/Clipboard.tsx
@@ -40,7 +40,7 @@ const ClipboardWrapper = styled<
   'data-testid': 'clipboard-wrapper'
 })`
   width: ${({ theme, clipboardHasOpenForms }) =>
-    clipboardHasOpenForms ? theme.front.minWidth : 200}px;
+    clipboardHasOpenForms ? theme.front.minWidth : 220}px;
   background: ${({ theme }) => theme.shared.collection.background};
   border-top: 1px solid ${({ theme }) => theme.shared.colors.greyLightPinkish};
   overflow-y: scroll;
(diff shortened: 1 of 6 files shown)


client-v2/src/constants/theme.ts
```diff
@@ -42,7 +42,9 @@ const capiInterface = {
 const front = {
   frontListBorder: '#5E5E5E',
   frontListLabel: shared.colors.greyMediumLight,
-  frontListButton: shared.colors.greyDark
+  frontListButton: shared.colors.greyDark,
+  minWidth: 380,
```
jonathonherbert: These values were attached to individual components. I've moved them into our theme.

jennygrahamjones:
A couple of the integration tests are failing as a result of this change. Is this because they're testing no longer valid behaviour, or just need updating?

client-v2/src/components/Clipboard.tsx
```diff
@@ -40,7 +40,7 @@ const ClipboardWrapper = styled<
   'data-testid': 'clipboard-wrapper'
 })`
   width: ${({ theme, clipboardHasOpenForms }) =>
-    clipboardHasOpenForms ? theme.front.minWidth : 200}px;
+    clipboardHasOpenForms ? theme.front.minWidth : 220}px;
```
jennygrahamjones: When the integration tests run and attempt to drag an item from the clipboard to a collection, the area of the item that they select is the item _image_, rather than the item itself. By slightly increasing the clipboard width, the correct area of the item is selected.

jonathonherbert: thanks @jennygrahamjones! This should no longer be a problem in subsequent PRs due to #871.

PR #877, opened 2019-07-26
tweak editions link style

akash1810:
## What's changed?
Style the Editions link from #870. Also corrects path.

### Before
image

### After
image

## Implementation notes
img

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/EditionFeedSectionHeader.tsx
+++ b/client-v2/src/components/EditionFeedSectionHeader.tsx
@@ -16,6 +16,11 @@ interface ComponentProps {
   publishEditionsIssue: (id: string) => Promise<void>;
 }
 
+const ManageLink = styled(Link)`
+  color: white;
+  text-decoration: none;
+`;
+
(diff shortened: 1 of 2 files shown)


client-v2/src/components/EditionFeedSectionHeader.tsx
```diff
@@ -44,14 +49,14 @@ class EditionFeedSectionHeader extends React.Component<ComponentProps> {
 
     return (
       <>
-        <Link to="/v2/manage-editions/daily-edition">
```
akash1810: This is because the react app's route is `/v2`, thus all paths are relative to it.

jonathonherbert: it'd be safer to parameterise this -- it's available in constants/urls.ts as appRoot 👍

akash1810: 👍 done.

PR #879, opened 2019-07-26
don't restrict Grid cropping ratio for Editions

akash1810:
## What's changed?
[Card](https://trello.com/c/4zQC27P5/63-picture-overrides-should-allow-any-aspect-ratio-crop)

At the moment the aspect ratio is locked to 5:3 for picture overrides. As the overridden image can be a portrait crop, we should allow other crop sizes when overriding images for Editions fronts cards.

## Implementation notes
Another possible route to achieve this is to have a `restrictToLandscape: boolean` prop on `InputImage`, which is arguable cleaner, however it's a bigger diff... move fast...

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/components/input/InputImage.tsx
+++ b/client-v2/src/shared/components/input/InputImage.tsx
@@ -22,6 +22,8 @@ import { RubbishBinIcon, AddImageIcon } from '../icons/Icons';
 import imageDragIcon from 'images/icons/image-drag-icon.svg';
 import { DRAG_DATA_GRID_IMAGE_URL } from 'constants/image';
 import ImageDragIntentIndicator from '../ImageDragIntentIndicator';
+import { EditMode } from 'types/EditMode';
+import { selectEditMode } from '../../../selectors/pathSelectors';
 
 const ImageContainer = styled('div')<{
   small?: boolean;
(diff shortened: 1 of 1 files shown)


client-v2/src/shared/components/input/InputImage.tsx
```diff
@@ -168,9 +170,11 @@ class InputImage extends React.Component<ComponentProps, ComponentState> {
       gridUrl,
       useDefault,
       defaultImageUrl,
-      message = 'Add image'
+      message = 'Add image',
+      editMode
     } = this.props;
-    const gridSearchUrl = `${gridUrl}?cropType=landscape`;
+    const gridSearchUrl =
+      editMode === 'editions' ? `${gridUrl}` : `${gridUrl}?cropType=landscape`;
```
akash1810: `gridUrl` is of type `string | null` and `gridSearchUrl` needs to be a string... so ``${gridUrl}` rather than just `gridUrl`.

akash1810:
> Won't ${null} just propagate the issue to somewhere else? I'd even prefer an alert if the URL is null to just failing to open the iframe

👍

tbh, I don't really know why it's optional, maybe we can be stricter

akash1810:
> Won't `${null}` just propagate the issue to somewhere else? I'd even prefer an alert if the URL is `null` to just failing to open the iframe

`${null}?cropType=landscape` can happen currently on master, so not sure this is a regression. It's definitely not ideal.

PR #883, opened 2019-07-30
RECENTLY REMOVED: display 5 most recently removed articles on each collection

ajwl:
## What's changed?

This PR limits the number of displayed articles in each collection's "recently removed" section to 5. Previously the number was unlimited.

It also changes the text and title of the box. 

Additionally there was a bug with the old implementation, which meant that articles in recently removed didn't turn up in the order in which they had been removed. They were bunched by group instead of a strict time order. 

OLD
Screenshot 2019-07-30 at 16 27 16

NEW 
Screenshot 2019-07-30 at 16 32 25

## Implementation notes

Why is this PR so complicated?

 Because we need to remake some of the data structures to allow for articleFragments to be managed at a collection level, rather than at a Group level. 

Previously the denormalisation process going from the collections json removed ordering information from the collection articleFragments when putting them into Groups. 

So articleFragments were only accessible as part of Groups. We have to break them out to be saved at the Collection level. 

*How to view the recently removed behaviour*
If you launch a collection, remove an article, then launch the collection again - the most recently removed article should turn up at the top of the removed list. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
+++ b/client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
@@ -18,7 +18,7 @@ import {
   createSelectCollectionStageGroups,
   createSelectCollectionEditWarning,
   selectSharedState,
-  createSelectFivePreviouslyLiveArticlesInCollection
+  createSelectPreviouslyLiveArticlesInCollection
 } from 'shared/selectors/shared';
 import {
   selectIsCollectionOpen,
@@ -211,7 +211,7 @@ class Collection extends React.Component<CollectionProps> {
 const createMapStateToProps = () => {
   const selectCollectionStageGroups = createSelectCollectionStageGroups();
   const selectEditWarning = createSelectCollectionEditWarning();
-  const selectPreviously = createSelectFivePreviouslyLiveArticlesInCollection();
+  const selectPreviously = createSelectPreviouslyLiveArticlesInCollection();
   return (
     state: State,
     {
--- a/client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
+++ b/client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
@@ -54,7 +54,7 @@ type CollectionProps = CollectionPropsBeforeState & {
   hasUnpublishedChanges: boolean;
   canPublish: boolean;
   groups: Group[];
-  previousGroups: Group[];
+  previousGroup: Group;
   displayEditWarning: boolean;
   isCollectionLocked: boolean;
   isEditFormOpen: boolean;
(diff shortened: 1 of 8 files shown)


ajwl:
@jennygrahamjones i fixed the old tests. But I wanted to add a new one.... got halfway there in the last commit. Can you try to make it work? Will pair when I get in

ajwl:
NB: tiny style change as per https://app.zeplin.io/project/5c499d59950ef83785bcc498/screen/5d308915622ec75194474971

client-v2/src/shared/util/shared.ts
```diff
@@ -123,6 +124,18 @@ const addGroups = (
     { live: [], draft: [], previously: [], addedGroups: {} } as ReduceResult
   );
 
+const createPreviouslyArticleFragmentIds = (
```
jennygrahamjones: Is it worth adding some explanation of why we're doing this? Without context, it probably looks a bit unusual. Maybe something like:

```suggestion
// To determine the UUIDs of article fragments recently removed from a collection in a way that
// preserves the overall ordering lost during normalisation (when fragments are assigned to a Group)
// we need to compare both the pre-normalised and normalised versions of the same collection.
const createPreviouslyArticleFragmentIds = (
```

ajwl: sure thing

client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
```diff
@@ -18,7 +18,7 @@ import {
   createSelectCollectionStageGroups,
   createSelectCollectionEditWarning,
   selectSharedState,
-  createSelectPreviouslyLiveArticlesInCollection
+  createSelectFivePreviouslyLiveArticlesInCollection
```
jonathonherbert: This name is very specific! It'd be good to parameterise this in the selector -- if we change the count, we have to update the code and the naming, and it's possible for developers to forget

client-v2/src/shared/selectors/shared.ts
```diff
@@ -222,15 +222,23 @@ const createSelectCollectionStageGroups = () => {
...
     selectCollection,
-    selectGroups,
-    (collection: Collection | void, groups: { [id: string]: Group }): Group[] =>
-      ((collection && collection[collectionItemSets.previously]) || []).map(
-        id => groups[id]
-      )
+    // All components that display articles do so using Groups. As a result, we have to create
+    // a Group here, to be able to render the previously removed articles.
+    // This will return the UUIDs for the 5 most recently removed articles.
+    // TODO: consider how we could change this interface, so we don't need to create this.
```
jonathonherbert: 👍

client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
ajwl: Oh sure, can amend now

ajwl:
@jennygrahamjones please do the honour of merging this tomorrow

PR #884, opened 2019-07-31
Edition version

akash1810:
## What's changed?

The main aim of this PR is to add a `version` field to the top level of a published issue. This field only exists when it is a published issue. Preview issues will not have a `version` field. 

The `version` field should be treated as opaque but is currently implemented as the timestamp of the time the edition is published. In the long run we will corellate this with successfully published issued from the archiver to provide a status dashboard for users.

## Implementation notes

This has simplified the bucket code somewhat as we now use `version` to form the edition key in the bucket and fallback to `preview` when it is absent.

We've added a number of tests that ensure that the serialisation logic does the right thing and won't be accidentally changed breaking downstream systems.

### General
- [X] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


app/services/editions/publishing/PublishedIssueFormatters.scala
```diff
@@ -4,10 +4,10 @@ import model.editions._
 import play.api.libs.json.Json
 
 object PublishedIssueFormatters {
-  implicit val publishedImageFormat = Json.format[PublishedImage]
-  implicit val publishedFurnitureFormat = Json.format[PublishedFurniture]
-  implicit val publishedArticleFormat = Json.format[PublishedArticle]
-  implicit val publishedCollectionsFormat = Json.format[PublishedCollection]
-  implicit val publishedFrontsFormat = Json.format[PublishedFront]
-  implicit val publishedIssueFormat = Json.format[PublishedIssue]
+  implicit val publishedImageWrites = Json.writes[PublishedImage]
```
akash1810: why?

sihil: We don't use them for reading in. We can make it explicit. If we want to be able to use them for reading in we need to add more tests to ensure that the format doesn't change.

PR #890, opened 2019-08-02
Rename front backend working

itsibitzi:
Added the backend code for handling front-level metadata fields such as name overrides and swatches.

Basically adds a small case class for `EditionsFrontMetadata` with only the name override stuff for now.

Putting this out there so that other front-level metadata can build on top of this.


itsibitzi:
Small aside, there's currently 2 flags in the templates, `hidden` and `canRename`. I think in all cases these are actually going to be the same, aka both flags are `true` when it's a "Special" front, both flags are false otherwise.

We might want to fold both these flags into a single `isSpecial` field, but might want to ask Blishen about it before?

sihil:
You'll want to unhide the special front and still be able to rename it. Don't think you can do that with a single flag and reckon it's best to keep the concerns separate.

sihil:
Would be nice to have some tests to ensure that the metadata can be stored and retrieved from the DB correctly.

PR #892, opened 2019-08-05
Front Renaming - Frontend

itsibitzi:
## What's changed?

Add a button to rename a front!

image

This only appears on fronts with the `canRename` flag set to true.

It works by replacing the title with an input box and focusing that box. Any blur or pressing enter will submit the new title. To go back to the old one just enter nothing.

## Implementation notes

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/services/editions/db/FrontsQueries.scala
+++ b/app/services/editions/db/FrontsQueries.scala
@@ -5,7 +5,7 @@ import scalikejdbc._
 import play.api.libs.json._
 
 trait FrontsQueries {
-  def updateFrontMetadata(id: String, metadata: EditionsFrontMetadata): Boolean = DB localTx { implicit session =>
+  def updateFrontMetadata(id: String, metadata: EditionsFrontMetadata): Option[EditionsFrontMetadata] = DB localTx { implicit session =>
     sql"""
           UPDATE fronts
           SET metadata = ${metadata.toPGobject}
@@ -36,16 +36,4 @@ trait FrontsQueries {
       }).single().apply().get
     Json.fromJson[EditionsFrontMetadata](Json.parse(rawJson)).get
   }
-
-  def mergeFrontMetadata(id: String, metadata: EditionsFrontMetadata): Boolean = DB localTx { implicit session =>
-    val original = getFrontMetadata(id)
-    val mergedMetadata = mergeMetadatas(metadata, original)
-    updateFrontMetadata(id, mergedMetadata)
-
-  }
-
-  def mergeMetadatas(additionalMetadata: EditionsFrontMetadata, originalMetadata: EditionsFrontMetadata): EditionsFrontMetadata = {
-    Json.fromJson[EditionsFrontMetadata](Json.toJson(originalMetadata).as[JsObject] ++ Json.toJson(additionalMetadata).as[JsObject]).get
-  }
-
 }
--- a/app/services/editions/db/FrontsQueries.scala
+++ b/app/services/editions/db/FrontsQueries.scala
@@ -11,6 +11,15 @@ trait FrontsQueries {
           SET metadata = ${metadata.toPGobject}
           WHERE id = $id
       """.execute().apply()
+
+    sql"""
+        SELECT metadata FROM fronts WHERE id = $id
+      """.map { rs =>
+      rs.stringOpt("metadata").map { metadataString =>
+        // Throw if we can't parse the metadata to signal to the user that something is broken
+        Json.parse(metadataString).validate[EditionsFrontMetadata].get
+      }
+    }.single().apply().flatten
   }
 
   def getFrontMetadata(id: String): EditionsFrontMetadata = DB localTx { implicit session =>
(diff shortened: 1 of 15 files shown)


client-v2/src/components/FrontsEdit/FrontContainer.tsx
```diff
@@ -132,13 +152,58 @@ class Fronts extends React.Component<FrontsComponentProps, ComponentState> {
     this.props.frontsActions.editorCloseFront(this.props.frontId);
   };
 
+  public renameFront = () => {
```
jonathonherbert: There are lots of `public` methods on this class -- if we're not intending parent components to use them via refs or whatnot, they might be better as `private`.

client-v2/src/reducers/frontsReducer.ts
```diff
@@ -42,6 +42,13 @@ const reducer = (
     };
   }
   switch (action.type) {
+    case 'FETCH_UPDATE_METADATA_SUCCESS': {
+      return set(
+        ['frontsConfig', 'data', 'fronts', action.payload.frontId, 'metadata'],
+        action.payload.metadata,
+        newState
+      );
+    }
```
jonathonherbert: Re: our conversation yesterday, this is fine but I understand why it might make you feel a little uneasy! To mitigate, we could always create a new reducer that extended the `frontsConfigReducer` and add our action there to keep our state management on the same 'level', e.g.

```
const newConfigReducer = (state, action) => {
  newState = frontsConfigReducer(state, action)
  switch (action.type) {
    case 'FETCH_UPDATE_METADATA_SUCCESS': {
       // ur logic here omg
    }
   // ...etc
}
```

... which should `just work` when eventually imported into the parent `frontsReducer`.

client-v2/src/components/FrontsEdit/FrontContainer.tsx
```diff
@@ -132,13 +152,58 @@ class Fronts extends React.Component<FrontsComponentProps, ComponentState> {
...
-      startCase(
-        this.props.selectedFront.displayName || this.props.selectedFront.id
+    const title = this.getTitle();
+
+    const { frontNameValue, editingFrontName } = this.state;
+    const canRename = this.props.selectedFront
+      ? this.props.selectedFront.canRename
+      : false;
+
+    const setName = () => {
```
jonathonherbert: This probably belongs in a method outside of `render`? Or you could write the component with hooks 😁

client-v2/src/reducers/frontsReducer.ts
itsibitzi: Alrighty, if you're ok with the slight code smell then I'll create a client side integration test incase we change how bundles work and call it a day.

itsibitzi:
Added integration testing - this is ready 4 review!

app/services/editions/db/FrontsQueries.scala
```diff
@@ -11,6 +11,15 @@ trait FrontsQueries {
           SET metadata = ${metadata.toPGobject}
           WHERE id = $id
       """.execute().apply()
+
+    sql"""
+        SELECT metadata FROM fronts WHERE id = $id
+      """.map { rs =>
+      rs.stringOpt("metadata").map { metadataString =>
+        // Throw if we can't parse the metadata to signal to the user that something is broken
+        Json.parse(metadataString).validate[EditionsFrontMetadata].get
```
akash1810: `metadata` is already of type `EditionsFrontMetadata`, so how would validation fail?

itsibitzi: If the JSON in the database doesn't reflect the case class we expect. For example if a required field is missing after a migration or whatevs.

client-v2/src/components/FrontsEdit/FrontContainer.tsx
```diff
@@ -196,6 +246,49 @@ class Fronts extends React.Component<FrontsComponentProps, ComponentState> {
...
+
+  private getTitle = () => {
+    const { selectedFront } = this.props;
+
+    if (selectedFront) {
+      if (selectedFront.metadata && selectedFront.metadata.nameOverride) {
+        return selectedFront.metadata.nameOverride;
+      } else {
+        return startCase(
+          this.props.selectedFront.displayName || this.props.selectedFront.id
```
akash1810: minor: we've destructured above, so can reuse `selectedFront`

```suggestion
          selectedFront.displayName || selectedFront.id
```

client-v2/src/components/FrontsEdit/FrontContainer.tsx
```diff
@@ -196,6 +246,49 @@ class Fronts extends React.Component<FrontsComponentProps, ComponentState> {
...
+    }
+
+    return;
+  };
+
+  private setName = () => {
+    const metadata =
+      this.state.frontNameValue !== ''
+        ? {
+            ...this.props.selectedFront.metadata,
```
akash1810: minor: we could use lodash set here, for example https://github.com/guardian/facia-tool/commit/27bb9c4a0e9caa61cf83013bf920f90a7feeadc9

akash1810: minor: we could use lodash set here, for example https://github.com/guardian/facia-tool/commit/27bb9c4a0e9caa61cf83013bf920f90a7feeadc9

PR #903, opened 2019-08-08
allow a container to be hidden/shown

akash1810:
## What's changed?
Allow a container to be hidden within an Edition's Front.

**Hidden**
image

**Unhidden**
image

## Implementation notes
Pairing on this was a lot of fun!

Possible TODOs for followup PRs:
- Cleaner UX
- Only allow specific containers to be hidden (we don't have a model for this yet)

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/bundles/collectionsBundle.ts
+++ b/client-v2/src/shared/bundles/collectionsBundle.ts
@@ -1,6 +1,8 @@
 import { State as SharedState } from '../types/State';
-import createAsyncResourceBundle from 'lib/createAsyncResourceBundle';
+import createAsyncResourceBundle, {State, Actions} from 'lib/createAsyncResourceBundle';
 import { Collection } from 'shared/types/Collection';
+
+const collectionsEntityName = 'collections';
 
 const {
   actions,
@@ -8,7 +10,7 @@ const {
   reducer,
   selectors,
   initialState
-} = createAsyncResourceBundle<Collection>('collections', { indexById: true });
+} = createAsyncResourceBundle<Collection>(collectionsEntityName, { indexById: true });
 
 const collectionSelectors = {
   ...selectors,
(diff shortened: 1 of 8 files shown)


client-v2/src/shared/bundles/collectionsBundle.ts
```diff
@@ -49,10 +57,65 @@ const collectionSelectors = {
...
+  action: CollectionActions
+): State<Collection> => {
+  const updatedState = reducer(state, action);
+  switch (action.type) {
+    case SET_HIDDEN: {
+      if (!updatedState.data[action.payload.collectionId]) {
+        return updatedState;
+      }
+
+      const freshState = {
```
jonathonherbert: We could use 'lodash/fp/set' to save some lines here -- see e.g frontsReducer.ts.

akash1810: Done.

client-v2/src/shared/bundles/collectionsBundle.ts
```diff
@@ -49,10 +58,59 @@ const collectionSelectors = {
   }
 };
 
+const SET_HIDDEN = 'SET_HIDDEN' as 'SET_HIDDEN';
```
itsibitzi: whuh?

sihil: Weird hey? @jonathonherbert says this is some magic to make a string token type.

itsibitzi: Ohhh - kinda like `const SET_HIDDEN: 'SET_HIDDEN' = 'SET_HIDDEN'` I gotcha.

sihil: https://basarat.gitbooks.io/typescript/docs/types/literal-types.html

sihil:
The card doesn't have any requirement that only special fronts should be hidable/unhidable. This will make unhiding be possible which will deliver immediate value and we can refine it later if required.

client-v2/src/shared/bundles/collectionsBundle.ts
jonathonherbert: @sihil @itsibitzi Haha, yes, it's odd. I learned last week can now do this as of TS 3.4 with 'const assertions':

```const SET_HIDDEN = 'SET_HIDDEN' as const```ts

which at least removes the repetition.

PR #908, opened 2019-08-09
Add sparklines

jennygrahamjones:
## What's changed?

This introduces "spark lines" or page view data graphs to the tool. 

**From a user perspective - these graphs should appear if:**
a) a feature flag for sparklines is switched on
b) the collection which you are viewing is able to show these graphs (for eg, they don't show on clipboard or in sublinks)
c) the story has had page views greater than 0 in the last hour. 

The graph displays alongside the thumbnail, if there is space...

Screenshot 2019-08-09 at 13 21 30

...or beneath the thumbnail when there are multiple fronts open.

Screenshot 2019-08-09 at 13 22 17

**What is the data and where does it come from** 

This data comes from Ophan on an endpoint called `histogram`, it returns the total hits on the story in a set time period (set to 1hr) and 6 datapoints  - the pageview at 10minute intervals. 

The request contains the Front the story is on, and the story's `urlPath`. It would be possible for a story to have two different sets of page view data - for example if it is hosted on 2 different fronts. Ophan is not aware of collections. 

The clientside fires off requests to a route on the backend every 10 seconds. The scala backend makes the request to Ophan. 

We only request data for open collections on open fronts. 

**Displaying the data**

We are using a new React friendly charting library [reCharts](https://github.com/recharts/recharts). It is open-source and easy to use. 

**What if Ophan falls over**

We don't do anything smart here. If Ophan falls over, we continue trying to poll it every 10 seconds. We also will continue to show the old data which means that it is possible charts will be stale for the duration of the time that Ophan is down. 

## Implementation notes

* We used the new Redux modules folder structure here - putting all Redux files related to this page view feature in one folder. As pioneered by Jon and feature flags. 

* A lot of the code is to make the call to Ophan - getting all the pieces together in the same place, from bits of state scattered around. 

* A small amount of code is used to actually control the display and generation of the graph.

* A fair amount of the code is used to manipulate the json into the correct shape so that it can be incorporated correctly into the components. This includes reintroducing the idea of collections, which we need from a rendering perspective. 

* Unfortunately Ophan does not have the concept of the `article uuid` or the `internal page code` so we need to run through several transforms to link the article's urlPath (used by Ophan) to the article's id (used by facia-tool for rendering)

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/components/FrontsEdit/Collection.tsx
```diff
@@ -109,11 +114,27 @@ type ConnectedCollectionContextProps = CollectionContextProps & {
   handleBlur: () => void;
   lastDesktopArticle?: string;
   lastMobileArticle?: string;
-};
+  pageViewData: PageViewDataPerCollection | undefined;
```
akash1810: could this be an optional type instead?

client-v2/src/redux/modules/pageViewData/actions.ts
```diff
@@ -0,0 +1,127 @@
...
+  };
+};
+
+const fetchPageViewData = (
+  front: string,
+  articlesInFront: ArticlePathAndId[]
+): Promise<any> => {
+  return getPageViewDataFromOphan(buildRequestUrl(front, articlesInFront)).then(
+    res => {
+      return res;
```
akash1810: given we're not doing anything with `res` we could just `return getPageViewDataFromOphan(buildRequestUrl(front, articlesInFront))`

ajwl:
Wrt to the article identifiers: unfortunately Ophan does not have the concept of the article uuid or the internal page code. And they don't plan to change that. So we do need to run through several transforms to link the article's urlPath (used by Ophan) to the article's id (used by facia-tool for rendering)

PR #909, opened 2019-08-09
Remove unnecessary form fields for e-mail priorities.

jonathonherbert:
## What's changed?

Remove unnecessary fields when we're in the `email` priority.

Screenshot 2019-08-09 at 14 13 29

Look at that lovely concise form!

## Implementation notes

I've taken the first step towards slightly more typesafe form code in this PR by adding the `FormFields` type, derived from the keys of the form output. As a result, incorrect strings in lists of fields elsewhere will now throw type errors.

Perhaps the next step is to store these in a map and key into them when we instantiate fields, to ensure our field names in the form component match across our codebase.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/selectors/__tests__/formSelectors.spec.ts
+++ b/client-v2/src/selectors/__tests__/formSelectors.spec.ts
@@ -2,30 +2,28 @@ import cloneDeep from 'lodash/cloneDeep';
 import {
   createSelectFormFieldsForCollectionItem,
   defaultFields,
-  supportingFields
+  supportingFields,
+  emailFieldsToExclude
 } from '../formSelectors';
 import { state, stateWithVideoArticle } from 'fixtures/form';
+import without from 'lodash/without';
 
 describe('Form utils', () => {
-  describe('getFormFieldsForCollectionItem', () => {
+  describe('selectFormFieldsForCollectionItem', () => {
     it("should handle articles that don't exist in the state", () => {
       const selectFormFields = createSelectFormFieldsForCollectionItem();
-      expect(selectFormFields(state as any, 'who-are-you')).toEqual([]);
+      expect(selectFormFields(state, 'who-are-you')).toEqual([]);
     });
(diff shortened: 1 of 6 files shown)


client-v2/src/selectors/__tests__/formSelectors.spec.ts
```diff
@@ -2,30 +2,28 @@ import cloneDeep from 'lodash/cloneDeep';
...
 import { state, stateWithVideoArticle } from 'fixtures/form';
+import without from 'lodash/without';
 
 describe('Form utils', () => {
-  describe('getFormFieldsForCollectionItem', () => {
+  describe('selectFormFieldsForCollectionItem', () => {
     it("should handle articles that don't exist in the state", () => {
       const selectFormFields = createSelectFormFieldsForCollectionItem();
-      expect(selectFormFields(state as any, 'who-are-you')).toEqual([]);
+      expect(selectFormFields(state, 'who-are-you')).toEqual([]);
```
jonathonherbert: I've boyscouted a bit here and removed the unnecessary `any` -- looks like we now model our state in a complete way 🎉

client-v2/src/util/pollingConfig.ts
```diff
@@ -14,12 +13,10 @@ export default (store: Store) =>
     if ((window as any).IS_INTEGRATION) {
       return;
     }
-    const path = selectV2SubPath(store.getState());
-    const match = matchFrontsEditPath(path) || matchIssuePath(path);
-    if (!match || !match.params.priority) {
+
+    const priority = selectPriority(store.getState());
```
jonathonherbert: A single selector is a little cleaner, and lets us re-use the logic.

jonathonherbert: @jennygrahamjones @ajwl this may affect your sparklines work (hopefully it's handy though!)

PR #914, opened 2019-08-09
always set customKicker aka deprecate showKickerTag and showKickerSection

akash1810:
## What's changed?
Updates the behaviour of the kicker suggestions - rather than setting the `showKickerTag` or `showKickerSection` flag, set the `customKicker` to the text value of the suggestion.

## Implementation notes
Change is applied to both Fronts and Editions.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
+++ b/client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
@@ -226,7 +226,13 @@ class FormComponent extends React.Component<Props, FormComponentState> {
 
     const isEditionsMode = editMode === 'editions';
 
+    // const assertion to prevent type widening, necessary to keep `type KickerField` below DRY
+    // see https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions
     const kickerFields = ['showKickerTag', 'showKickerSection'] as const;
+
+    // In an effort to be DRY, create a type of all values from the `kickerFields` array that can be accessed by a number.
+    // That is, a union of the values.
+    // i.e. KickerField = 'showKickerTag' | 'showKickerSection'
     type KickerField = typeof kickerFields[number];
 
     const setCustomKicker = (customKickerValue: string) => {
--- a/client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
+++ b/client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
@@ -226,6 +226,32 @@ class FormComponent extends React.Component<Props, FormComponentState> {
 
     const isEditionsMode = editMode === 'editions';
 
+    const kickerFields = ['showKickerTag', 'showKickerSection'] as const;
+    type KickerField = typeof kickerFields[number];
+
+    const setCustomKicker = (customKickerValue: string) => {
+      change('customKicker', customKickerValue);
+      change('showKickerCustom', true);
+      kickerFields.forEach(field => change(field, false));
+    };
+
+    const handleKickerChange = (
+      customKickerValue: string,
+      showField: boolean,
+      fieldName: KickerField
+    ) => {
+      if (isEditionsMode) {
+        setCustomKicker(customKickerValue);
+      } else {
+        if (!showField) {
+          kickerFields.forEach(field => change(field, field === fieldName));
+          change('showKickerCustom', false);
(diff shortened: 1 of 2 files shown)


client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
```diff
@@ -215,9 +217,42 @@ class FormComponent extends React.Component<Props, FormComponentState> {
...
+    const setCustomKicker = (customKickerValue: string) => {
+      change('customKicker', customKickerValue);
+      change('showKickerCustom', true);
+      kickerFields.forEach(field => change(field, false));
+    };
+
+    const handleKickerChange = (
+      customKickerValue: string,
+      showField: boolean,
+      fieldName: KickerField
```
sihil: These field names are less clear than they could be. I would perhaps suggest: `literalKickerValue`, `kickerFieldValue` and `kickerFieldName`.

client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
```diff
@@ -215,9 +217,42 @@ class FormComponent extends React.Component<Props, FormComponentState> {
...
+    ) => {
+      if (editMode === 'editions') {
+        setCustomKicker(customKickerValue);
+      } else {
+        if (!showField) {
+          kickerFields.forEach(field => change(field, field === fieldName));
+          change('showKickerCustom', false);
+        } else {
+          change(fieldName, false);
+        }
```
sihil: I'd switch the order of this if statement if it were me.

sihil: The logic seems to be the reverse of what I would expect. Why `!showField` => `change(fieldName, false)` ???

sihil: Ah - the method is a toggle and `showField` is the **old** value before the toggle. Can the `!` be moved to where handleKickerChange is called? Alternatively (at the very least) let's rename the params to make it clear that it's the `old` value.

client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
```diff
@@ -215,9 +217,42 @@ class FormComponent extends React.Component<Props, FormComponentState> {
...
+      change('showKickerCustom', true);
+      kickerFields.forEach(field => change(field, false));
+    };
+
+    const handleKickerChange = (
+      customKickerValue: string,
+      showField: boolean,
+      fieldName: KickerField
+    ) => {
+      if (editMode === 'editions') {
```
ajwl: And as mentioned - we want this behaviour in normal fronts too - so can take out this if statement. 
thanks

akash1810:
PR updated to reflect the fact that we want this behaviour for both Fronts and Editions - the diff is a lot smaller as a result!

akash1810:
This PR has the effect of deprecating `showKickerTag` and `showKickerSection` and potentially gives us the opportunity to simplify some frontend and apps templates.

cc @guardian/dotcom-platform @guardian/mobile-server-side-staff

alexduf:
So this change would break compatibility with our app right? can that be broken down in multiple steps so we have the time to upgrade the consumers?

akash1810:
> So this change would break compatibility with our app right? can that be broken down in multiple steps so we have the time to upgrade the consumers?

It doesn't break compatibility, but rather reduces complexity downstream. Take this scenario:

image

Current behaviour:
- Click `Plane crashes`
- Set flag `showKickerTag` to `true`
- Set `customKicker` field to empty string
- Rendering clients see `showKickerTag = true`, looks up the lead tag for the article, and renders the `webTitle` of that tag

Behaviour in this PR:
- Click `Plane crashes`
- Set `customKicker` to `Plane crashes`
- Rendering clients render value of `customKicker`

The `showKickerTag` and `showKickerSection` are intended to allow the lead tag (or section tag) of an article to change which in turn will change how the article's kicker appears on the front. That is, if this article went from `Plane crashes` to `King Kong`, the kicker would automagically change to `King Kong`. We've confirmed with CP that this behaviour isn't needed.

alexduf:
Makes perfect sense, thanks @akash1810

PR #915, opened 2019-08-12
Justinrowles/save content based settings

aug24:
## What's changed?

Include in our metadata persistence all the values which are default in the following link except for video as Editions will not support video

https://github.com/guardian/facia-scala-client/blob/8a65c105a11cc40cfd24ce4565fda6edf16dcc51/fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala#L58

## Implementation notes

JSON fragments deliberately only include the values if explicitly set.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


app/model/editions/client/ClientArticleMetadata.scala
```diff
@@ -31,8 +31,11 @@ case class ClientArticleMetadata (
   imageCutoutSrcOrigin: Option[String]
 ) {
   def toArticleMetadata: ArticleMetadata = {
+    println(imageCutoutSrcHeight, imageCutoutSrcWidth, imageCutoutSrc, imageCutoutSrcOrigin)
```
sihil: debug?

app/util/ContentUpgrade.scala
```diff
@@ -30,14 +30,21 @@ object ContentUpgrade {
...
-            case _ => None
+        val newValues:List[JField] = jsObject.values.map { case (key, _) =>
+          if (ContentFields.contains(key)) {
+            jsObject \ key match {
+              case JArray(items) => (key -> JArray(items.map(upgradeItem)))
+              case item: JObject => (key -> upgradeItem(item))
+              case value => (key, value)
+            }
+          } else {
+            (key, jsObject \ key)
```
sihil: I expect that `jsObject \ key` here is also the right handside of `(key, _)` in the pattern match?

client-v2/src/shared/util/articleFragment.ts
```diff
@@ -3,12 +3,25 @@ import { ArticleFragment, ArticleFragmentMeta } from 'shared/types/Collection';
...
+  showQuotedHeadline: boolean = false
+) => ({
   uuid: v4(),
   id,
   frontPublicationDate: Date.now(),
   meta: {
-    supporting
+    ...(imageHide ? { imageHide } : ''),
+    ...(imageReplace ? { imageReplace } : ''),
+    // ...(imageCutoutReplace ? { imageCutoutReplace } : ''),
```
sihil: delete commented line

client-v2/src/shared/util/articleFragment.ts
```diff
@@ -3,12 +3,25 @@ import { ArticleFragment, ArticleFragmentMeta } from 'shared/types/Collection';
...
+    ...(imageCutoutReplace ? { imageCutoutReplace, imageCutoutSrc } : ''),
```
sihil: Magic notation 🤯

itsibitzi: Would it be clearer to spread an empty object here rather than an empty string?

app/model/editions/client/ClientArticleMetadata.scala
aug24: Damn.

client-v2/src/shared/util/articleFragment.ts
sihil: That is my instinct but I'm aware my TS foo is still lacking.

app/util/ContentUpgrade.scala
```diff
@@ -48,20 +55,41 @@ object ContentUpgrade {
...
+          val (metaDataMap, maybeFirstContributorWithCutout) = initialMetaDataMap.get("imageCutoutReplace") match {
+            case Some(true) => {
+              getBylineLargeImageUrlOption(content)  match {
+                case bylineLargeImageUrlOption@Some(_) => (initialMetaDataMap, bylineLargeImageUrlOption)
+                // If we can't find a cutout value, turn off the cutout bool
+                case None => (initialMetaDataMap + ("imageCutoutReplace" -> false), None)
+              }
+            }
+            case _ => (initialMetaDataMap, None)
+          }
```
sihil: Would be great to extract this block and then factor out https://github.com/guardian/facia-tool/blob/master/app/services/Capi.scala#L180-L187

client-v2/src/shared/util/articleFragment.ts
aug24: It does the exact same thing, I think, but OK.

app/util/ContentUpgrade.scala
aug24: I don't have time today.  Tech debt card?

sihil: Yes.

PR #922, opened 2019-08-13
POP-UP: Pop on v1 directing users to v2

ajwl:
## What's changed?

We want users on V1 to get the message that they should switch over to V2. 

This adds a permanent change to the header that should run until v1 is deprecated at the end of Aug 2019.

We opted NOT to make this dismissable so that the link is always available to people wanting to switch.

It will not show on breaking news or config

Screenshot 2019-08-14 at 17 46 14

## Implementation notes

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


ajwl:
@jennygrahamjones i got some horrible prettifier changes stuck in the diff so just made a different branch.

app/views/templates/header.scala.html
```diff
@@ -11,6 +11,8 @@ <h1 class="header__item header__link">
         </a>
     </div>
 
+     <div class="header__group v2-popup">There's a new version of the fronts tool. Please&nbsp;<a href="https://fronts.gutools.co.uk/v2">try it by clicking here</a>.</div>
```
akash1810: Can we make this stage sensitive?

jennygrahamjones:
> Is that header shared by any other parts of the app, like config or breaking news? If not LGTM!

It looks like the header is used by all views in the app, yes. We could provide the template(s) with an extra parameter to determine when to display the message?

PR #928, opened 2019-08-15
Sparkline optimisations

jennygrahamjones:
## What's changed?

Instead of re-rendering the collection every time new page view data is received from Ophan, only the article graph itself is re-rendered. 

Additionally, the call to Ophan will now only be made if the user has toggled on the page view data feature. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/redux/modules/pageViewData/selectors.ts
```diff
@@ -1,34 +1,43 @@
...
-  );
+): PageViewStory | undefined => {
+  const dataForFront: PageViewDataPerFront | undefined = selectPageViewData(
+    state
+  ).find(front => front.frontId === frontId);
+
+  const allArticlesOnFront =
+    dataForFront &&
+    dataForFront.collections.reduce(
+      (acc: PageViewArticlesOnFront, curr: PageViewDataPerCollection) => {
```
jonathonherbert: We'll need to memoize this to ensure we're not doing expensive graph work unnecessarily -- happy to pair!

client-v2/src/shared/components/article/ArticleGraph.tsx
```diff
@@ -1,35 +1,47 @@
...
+          />
+        </AreaChart>
+      </>
     );
   }
 }
 
-export default ArticleGraph;
+const mapStateToProps = (state: State, props: ArticleGraphProps) => ({
+  data: selectPageViewDataForArticleId(state, props.articleId, props.frontId)
```
jonathonherbert: Awesome 👌

client-v2/src/util/pollingConfig.ts
```diff
@@ -24,6 +24,12 @@ export default (store: Store) =>
     }
     (store.dispatch as Dispatch)(fetchStaleOpenCollections(priority));
 
+    if (
```
jonathonherbert: 👍

ajwl:
@jonathonherbert thanks

PR #931, opened 2019-08-16
Don't display the v2 banner on e-mail priorities

jonathonherbert:
## What's changed?

Removes the v2 banner on the `email` priority. E-mail fronts aren't quite ready yet, and Celine is worried about non tech-savvy users making the switch and being ... troubled.

They've got the front bookmarked, so we can keep the message on the root route.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


jennygrahamjones:
I'm not sure how frequently email users navigate via the root URL to get to the email fronts, but if they do, it's probably worth updating `def priorities()` to make the same check. Otherwise, 👍

jonathonherbert:
@jennygrahamjones I had run this past Celine, as I was keen to keep messaging where possible, and she's happy 👍

PR #933, opened 2019-08-16
Apply media overrides - backend code only

aug24:
## What's changed?

Allows the UI code to push a value for 'apply media overrides' to the back and and then to the database.

Does not provide a way for the UI code to set the value, pending design work.

### General
- [X] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/editions/PublishedIssue.scala
+++ b/app/model/editions/PublishedIssue.scala
@@ -29,7 +29,7 @@ case class PublishedFurniture(
   mediaType: PublishedMediaType,
   imageSrcOverride: Option[PublishedImage],
   sportScore: Option[String],
-  overrideArticleMedia: Boolean
+  overrideArticleMainMedia: Boolean
 )
 
 case class PublishedArticle(internalPageCode: Long, furniture: PublishedFurniture)
--- a/app/model/editions/PublishedIssue.scala
+++ b/app/model/editions/PublishedIssue.scala
@@ -28,7 +28,8 @@ case class PublishedFurniture(
   showQuotedHeadline: Boolean,
   mediaType: PublishedMediaType,
   imageSrcOverride: Option[PublishedImage],
-  sportScore: Option[String]
+  sportScore: Option[String],
+  overrideArticleMedia: Boolean
 )
 
 case class PublishedArticle(internalPageCode: Long, furniture: PublishedFurniture)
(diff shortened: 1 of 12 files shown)


AWare:
Is there a difference in `Some(false)`s and `None`s meanings here?

sihil:
Not sure where but `None` typically means no explicit setting whilst `Some(false)` is explicitly set to `false`. We usually hydrate defaults when exporting so you don't have to concern yourselves with defaults.

app/model/editions/client/ClientArticleMetadata.scala
```diff
@@ -28,7 +28,9 @@ case class ClientArticleMetadata (
   imageCutoutSrc: Option[String],
   imageCutoutSrcHeight: Option[String],
   imageCutoutSrcWidth: Option[String],
-  imageCutoutSrcOrigin: Option[String]
+  imageCutoutSrcOrigin: Option[String],
+
+  overrideArticleMainMedia: Option[Boolean]
```
AWare: So I'd be expecting this as a `Boolean`?

sihil: The classes that are exported to you are the ones prefixed with `Published`.

app/model/editions/PublishedIssue.scala
```diff
@@ -28,7 +28,8 @@ case class PublishedFurniture(
   showQuotedHeadline: Boolean,
   mediaType: PublishedMediaType,
   imageSrcOverride: Option[PublishedImage],
-  sportScore: Option[String]
+  sportScore: Option[String],
+  overrideArticleMainMedia: Boolean
```
sihil: This is the field you see @AWare

test/services/editions/db/EditionsDBTest.scala
```diff
@@ -135,6 +136,13 @@ class EditionsDBTest extends FreeSpec with Matchers with EditionsDBService with
       val commentFront = retrievedIssue.fronts.tail.head
       commentFront.displayName shouldBe "comment"
       commentFront.collections.length shouldBe 3
+
+      val editionsArticle = newsPoliticsCollection.items.head
```
akash1810: sneaking an extra check in 👍

PR #936, opened 2019-08-16
REPLACE IMAGE TWEAKS: Fix grid text input, make toggle more logical

ajwl:
## What's changed?

Replace image fields in articles: remove need to hit enter on text input & change the behaviour of the toggle.

Trello: https://trello.com/c/szlvJ9R0/787-replace-image-with-grid-url

With no replacement image present, toggle is hidden:

Screenshot 2019-08-19 at 18 26 29

When replacement image is added, the toggle appears, set to true: 

Screenshot 2019-08-19 at 18 26 52

## Implementation notes

- Knowledge of image replace disappears on save, meaning that the toggle itself will disappear, although it's still possible to remove the image using the little bin icon. 
- Need UX test this flow to confirm whether users understand what to do when the toggle is absent. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
+++ b/client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
@@ -48,6 +48,7 @@ import { selectors as collectionSelectors } from 'shared/bundles/collectionsBund
 import { getContributorImage } from 'util/CAPIUtils';
 import { EditMode } from 'types/EditMode';
 import { selectEditMode } from 'selectors/pathSelectors';
+import { ValidationResponse } from 'shared/util/validateImageSrc';
 
 interface ComponentProps extends ContainerProps {
   articleExists: boolean;
@@ -183,7 +184,6 @@ const getInputId = (articleFragmentId: string, label: string) =>
 interface FormComponentState {
   lastKnownCollectionId: string | null;
   displayImageReplaceToggle: boolean;
-  isImageReplaced: boolean;
 }
 
 class FormComponent extends React.Component<Props, FormComponentState> {
@@ -195,8 +195,7 @@ class FormComponent extends React.Component<Props, FormComponentState> {
 
   public state: FormComponentState = {
     lastKnownCollectionId: null,
-    displayImageReplaceToggle: false,
-    isImageReplaced: false
+    displayImageReplaceToggle: false
   };
 
   private allImageFields = [
@@ -441,10 +440,6 @@ class FormComponent extends React.Component<Props, FormComponentState> {
                   useDefault={!imageCutoutReplace && !imageReplace}
                   message={imageCutoutReplace ? 'Add cutout' : 'Replace image'}
                   onChange={this.handleImageChange}
-                  setDisplayImageReplaceToggle={
-                    this.setDisplayImageReplaceToggle
-                  }
-                  setImageReplaceToggleValue={this.setImageReplaceToggleValue}
                 />
               </ImageCol>
               <Col flex={2}>
@@ -494,7 +489,7 @@ class FormComponent extends React.Component<Props, FormComponentState> {
                       label="Use replacement image"
                       id={getInputId(articleFragmentId, 'image-replace')}
                       type="checkbox"
-                      default={this.state.isImageReplaced}
+                      default={false}
                       onChange={_ => this.changeImageField('imageReplace')}
                     />
                   </InputGroup>
@@ -554,17 +549,8 @@ class FormComponent extends React.Component<Props, FormComponentState> {
     return 'primaryImage';
   };
 
-  private setDisplayImageReplaceToggle = (display: boolean) => {
-    this.setState({ displayImageReplaceToggle: display });
-  };
-
-  private setImageReplaceToggleValue = (value: boolean) => {
-    this.props.change('imageReplace', value);
-    this.setState({ isImageReplaced: value });
-  };
-
   private handleImageChange: EventWithDataHandler<React.ChangeEvent<any>> = (
-    e,
+    e: ValidationResponse | undefined,
     ...args: [any?, any?, string?]
   ) => {
     // If we don't already have an image override enabled, enable the default imageReplace property.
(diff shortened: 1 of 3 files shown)


ajwl:
@jennygrahamjones fyi!

client-v2/src/shared/components/input/InputImage.tsx
```diff
@@ -289,31 +293,31 @@ class InputImage extends React.Component<ComponentProps, ComponentState> {
 
   private handlePasteImgSrcChange = (e: React.FormEvent<HTMLInputElement>) => {
     e.preventDefault();
-    this.setState({ imageSrc: e.currentTarget.value });
+    e.persist();
```
ajwl: Check what this does. Do we need it here?

client-v2/src/shared/components/input/InputImage.tsx
```diff
@@ -142,6 +142,9 @@ export interface InputImageContainerProps {
   defaultImageUrl?: string;
   useDefault?: boolean;
   message?: string;
+  replaceImage: boolean;
+  setDisplayImageReplaceToggle: (display: boolean) => void;
```
jonathonherbert: This couples our image input to the idea of an `imageReplace` field or similar, which future image fields or usages of this component (I'm thinking editions, although there might be other cases) may not need. Could we move this code outside of this component and deal with this logic in the `onChange` handler instead? Happy to talk through or pair on this if needed 👍

ajwl: Hmm interested by your suggestion. The problem we found is that the imageInput needs to pass certain temporary information (has this image been selected) up to its parent (ArticleParentFormInline - where the toggle is) and this appears to be simplest way to do it. 

Do you mean adding this state into the store in `ImageInput` and then pulling it off in `ArticleParentFormInline`?
I guess we should be cautious about coding around future problems that don't currently exist?

client-v2/src/shared/components/input/InputImage.tsx
jonathonherbert: `persist()` ensures that the event is available outside of its immediate context. It's a bit of an implementation detail from React -- React uses its own synthetic event handlers under the hood, and re-uses event objects liberally, so unless you call `persist()` when passing events elsewhere you can end up in situations where your event has disappeared. That probably sounds like evil magic. It kind of is. See https://reactjs.org/docs/events.html

I can't see that this event is being passed anywhere outside of this function, so you're probably safe to remove. You'll see red ink in the console if there are problems, so I'd remove, test and if all looks good then you're 👍

jonathonherbert: Just as a pointless NB, other vdom libraries like Preact don't bother with this synthetic event stuff, which I think was largely to smooth out event handling irregularities between browsers. So it doesn't have to be this way! It just _happens_ to be this way for React.

client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
```diff
@@ -435,23 +439,15 @@ class FormComponent extends React.Component<Props, FormComponentState> {
                       : articleCapiFieldValues.thumbnail
                   }
                   useDefault={!imageCutoutReplace && !imageReplace}
-                  message={imageCutoutReplace ? 'Add cutout' : 'Add image'}
+                  message={imageCutoutReplace ? 'Add cutout' : 'Replace image'}
                   onChange={this.handleImageChange}
+                  setDisplayImageReplaceToggle={
+                    this.setDisplayImageReplaceToggle
+                  }
+                  setImageReplaceToggleValue={this.setImageReplaceToggleValue}
```
jonathonherbert: See below comment re: these new fields -- could we do this in `onChange`, a bit like line 498? (God this form is long)

client-v2/src/shared/components/input/InputImage.tsx
ajwl: cool 👍 thanks for the explanation

ajwl:
Thanks for pointers @jonathonherbert - changes added. 
NB: there is a lingering type error noted above. Should we ignore or try to rewire so it is fixed?

client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
```diff
@@ -545,14 +550,17 @@ class FormComponent extends React.Component<Props, FormComponentState> {
   };
 
   private handleImageChange: EventWithDataHandler<React.ChangeEvent<any>> = (
-    e,
+    e: ValidationResponse | undefined,
     ...args: [any?, any?, string?]
```
ajwl: What's being passed in here is a `ValidationResponse` object with a src property etc, not an event. What the ConditionalField component expects the signature of the onChange function to be is `EventWithDataHandler<React.ChangeEvent<any>>` : 
```
Type '(e: ValidationResponse | undefined, args_0?: any, args_1?: any, args_2?: string | undefined) => void' is not assignable to type 'EventWithDataHandler<ChangeEvent<any>>'.
  Types of parameters 'e' and 'event' are incompatible.
    Type 'ChangeEvent<any> | undefined' is not assignable to type 'ValidationResponse | undefined'.
      Type 'ChangeEvent<any>' is missing the following properties from type 'ValidationResponse': src, thumb, origin, height, width
```

jonathonherbert: For posterity -- after a chat IRL, this has been resolved by the `unknown` type.

As far as we can tell, redux-form's typings for its `Field` component aren't quite right -- they expect an event in an `onChange` handler, which doesn't match the [current documentation.](https://redux-form.com/7.4.0/docs/api/field.md/#-code-onchange-event-newvalue-previousvalue-name-gt-void-code-optional-) Until they're correct, we patch this by casting the incoming value.

client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
```diff
@@ -544,15 +549,18 @@ class FormComponent extends React.Component<Props, FormComponentState> {
     return 'primaryImage';
   };
 
+  // onChange: EventWithDataHandler<React.ChangeEvent<any>> & ((_: React.ChangeEvent<any> | undefined) => void)
```
jonathonherbert: We probably don't need this 😁

PR #940, opened 2019-08-19
Changes to 'Show video' toggle

jennygrahamjones:
## What's changed?

- moved toggle to be located with other media hide/display toggles
- automatically disable video toggle if other media toggles are turned on

Trello for moving toggle: https://trello.com/c/LNS3aV70/806-move-show-video-toggle-next-to-main-media
Trello for making toggle consistent: https://trello.com/c/TitshcuD/805-show-video-replace-media-bugs

NB: (unable to test Preview behaviour until this is on Prod) 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/components/article/ArticleBody.tsx
+++ b/client-v2/src/shared/components/article/ArticleBody.tsx
@@ -295,7 +295,7 @@ const articleBodyDefault = React.memo(
                   {cutoutThumbnail ? (
                     <ThumbnailCutout src={cutoutThumbnail} />
                   ) : null}
-                  {showMainVideo && (
+                  {hasMainVideo && (
                     <VideoIconContainer>
                       <VideoIcon />
                     </VideoIconContainer>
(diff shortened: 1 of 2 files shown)


client-v2/src/shared/components/article/ArticleBody.tsx
```diff
@@ -342,7 +342,7 @@ const articleBodyDefault = React.memo(
                   {cutoutThumbnail ? (
                     <ThumbnailCutout src={cutoutThumbnail} />
                   ) : null}
-                  {hasMainVideo && (
+                  {showMainVideo && (
```
jonathonherbert: Are we certain this is the behaviour we want? We already have an indicator for `showMainVideo` on line 355. I think this icon is there to reveal that video is available, not that it's been overridden -- but I might be wrong!

jennygrahamjones: If the purpose of the icon is to show that here is a video _available_, rather than that the video is toggled on, then we should definitely preserve the existing behaviour.

PR #943, opened 2019-08-19
[readme-cleanup] refining setup scripts for dev env and README

mkuzdowicz:
## What's changed?

README and dev env setup scripts


client-v2/README.md
```diff
@@ -18,30 +18,13 @@
...
-./scripts/client-dev.sh
-```
-
-The v2 application will be running here:
-[https://fronts.local.dev-gutools.co.uk/v2](https://fronts.local.dev-gutools.co.uk/v2)
-
-View [client-dev.sh](scripts/client-dev.sh).
-
-If you wish to use git grep to search the code base you can use [fronts-find.sh](scripts/fronts-find.sh)
+1. Get credentials for `composer` and `workflow` from [Janus](https://janus.gutools.co.uk/multi-credentials?&permissionIds=cmsFronts-dev,capi-api-gateway,frontend-dev).
```
ajwl: And CMS Fronts perms

README.md
```diff
@@ -21,45 +21,20 @@ Fronts Client V2 is in Active Development. You'll find it over at [client-v2](/c
...
-
 #### NVM
 
 We use [NVM](https://github.com/creationix/nvm) to ensure a given Node version is used and project setup scripts run correctly.
 
 Ubuntu/Mac: [Follow these instructions](https://github.com/nvm-sh/nvm#installation-and-update)
 
+#### Dev Setup
+
+1. Get credentials for `composer` and `workflow` from [Janus](https://janus.gutools.co.uk/multi-credentials?&permissionIds=cmsFronts-dev,capi-api-gateway,frontend-dev).
```
ajwl: and cms-fronts here i guess, might as well spell them out

client-v2/README.md
```diff
@@ -18,30 +18,13 @@
...
+2. From the project root, run `./scripts/setup.sh`.
+3a. From the project root, run `./scripts/start-dev.sh`.
+3b. From the project root, run without debug `./script/start-dev.sh --no-debug`.
+4. Open `https://fronts.local.dev-gutools.co.uk/v2`.
```
jennygrahamjones: Do we need to do each of these steps every time we run the tool locally? If not, could we be clear which steps are one-time steps and which are any time you want to work locally?

client-v2/README.md
mkuzdowicz: done

client-v2/README.md
mkuzdowicz: done

client-v2/README.md
adamnfish: These hard-coded links will have their expiry time be 1-hour off during BST because they don't include the timezone parameter.

If people are wanting to link directly to Janus permission sets then we should look at adding an interstitial page that redirects to credentials having asked the user's browser for its timezone.

I know some people have bookmarked Janus credentials which would have the same problem, so this probably is worth looking at.

Nothing fundamentally wrong with this PR but the expiry time might be mildly annoying as-is. You could use a BST link for now to address the common use-case for the majority of today's developers, and I'll try and make sure we address the issue in Janus ahead of the hour change at the end of October.

PR #944, opened 2019-08-19
Snap links

jennygrahamjones:
## What's changed?

Snap links now have additional buttons on hover, allowing the link to be viewed or copied to the clipboard. Additionally, the link itself is displayed on the body of the card, replicating v1 behaviour. 

Screenshot 2019-08-19 at 16 46 12

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/components/snapLink/SnapLink.tsx
+++ b/client-v2/src/shared/components/snapLink/SnapLink.tsx
@@ -8,7 +8,10 @@ import CollectionItemMetaContainer from '../collectionItem/CollectionItemMetaCon
 import CollectionItemMetaHeading from '../collectionItem/CollectionItemMetaHeading';
 import { ThumbnailSmall } from '../Thumbnail';
 import { HoverActionsButtonWrapper } from '../input/HoverActionButtonWrapper';
-import { HoverDeleteButton } from '../input/HoverActionButtons';
+import {
+  HoverDeleteButton,
+  HoverAddToClipboardButton
+} from '../input/HoverActionButtons';
 import { HoverActionsAreaOverlay } from '../CollectionHoverItems';
 import { ArticleFragment, CollectionItemSizes } from 'shared/types/Collection';
 import {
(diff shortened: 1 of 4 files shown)


client-v2/src/shared/components/snapLink/SnapLink.tsx
```diff
@@ -64,6 +74,7 @@ const SnapLink = ({
     (articleFragment.meta.customKicker
       ? `{ ${articleFragment.meta.customKicker} }`
       : 'No headline');
+  const urlPath = articleFragment.meta.href;
```
jonathonherbert: For some diabolical reason we store Guardian paths without the guardian domain. Presumably so if the domain changes the links are OK. (`30X`?) For this reason, you'll need to add the appropriate domain if there isn't one, and we can crib from v1 here -- 

```
const normaliseSnapUrl = (href: string) => {
  if (href && !/^https?:\/\//.test(href)) {
    return 'https://' + urls.guDomain + href;
  }

  return href;
};
```

where urls.guDomain can be set in e.g. `constants/urls`.

(If this sounds oddly specific, I touched this task this afternoon, but you were faster 🚀 👍)

jonathonherbert: To test this, drag e.g. https://www.theguardian.com/uk into the tool. You should see a lonely `/uk` pop up in the snap, that links, inappropriately, to the tool itself.

client-v2/src/shared/components/snapLink/SnapLink.tsx
```diff
@@ -80,16 +91,30 @@ const SnapLink = ({
             <CollectionItemMetaHeading>Snap link </CollectionItemMetaHeading>
           )}
           <CollectionItemHeading html>{headline}</CollectionItemHeading>
+          <SnapLinkURL>
+            snap link url: &nbsp;
```
jonathonherbert: Probably just `url` is sufficient here -- the snap link is labeled just to the left of this component.

client-v2/src/shared/components/snapLink/SnapLink.tsx
jennygrahamjones: Aha. Is that regex just validating whether we have a complete URL or a partial path, as would be the case with Guardian links?

jonathonherbert: precisely -- it's logic that exists in v1 but it's obviously encoding a big assumption about our model (incomplete paths belong to the guardian!)

PR #954, opened 2019-08-20
Make v2 default by redirecting

akash1810:
## What's changed?

Redirect `/` to `/v2` if:

`redirect` query string is not present or is `true`

AND

(we're in DEV OR it's after Christmas 2019)

If `redirect` query string is not present, we default to `true`, however as it's not after Christmas 2019, we will not redirect. That is, there should be no impact on non DEV stages.

If, for whatever reason, we needed to go back to the index page of V1, then `?redirect=false` will do it.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/ViewsController.scala
+++ b/app/controllers/ViewsController.scala
@@ -3,6 +3,7 @@ package controllers
 import scala.concurrent.ExecutionContext
 import com.gu.pandomainauth.action.UserRequest
 import model.Cached
+import org.joda.time.DateTime
 import permissions.ConfigPermissionCheck
 import play.api.mvc._
 import services.AssetsManager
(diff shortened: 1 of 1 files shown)


app/controllers/ViewsController.scala
```diff
@@ -12,9 +12,15 @@ class ViewsController(val acl: Acl, assetsManager: AssetsManager, isDev: Boolean
                       crypto: Encryption, val deps: BaseFaciaControllerComponents)(implicit ec: ExecutionContext) extends BaseFaciaController(deps) {
 
   def priorities() = AccessAuthAction { request =>
-    val identity = request.user
-    Cached(60) {
-      Ok(views.html.priority(Option(identity), config.facia.stage, isDev, true))
+    val shouldRedirect: Boolean = request.queryString.getOrElse("redirect", Seq("true")).contains("true")
```
akash1810: If, for whatever reason` we needed to go back to the index page of V1, then `?redirect=false` will do it. (Or really, value, other than `true`).

akash1810:
> Just to check that this has been tested with V1 on e.g. CODE?

Yes.

Will make the date a param now that the message has been updated.

PR #955, opened 2019-08-20
SNAPLINKS: Add images in the body

ajwl:
## What's changed?
Images can be added to snaplinks... but they don't turn up in the containers :( 

Screenshot 2019-08-20 at 18 19 13

This should add images to snaplink containers - and now they do!

Screenshot 2019-08-22 at 10 33 41

Trello: https://trello.com/c/MeJ8mGfH/757-snap-trails-in-containers-image-not-displayed-on-snap-in-container-when-an-image-is-added-to-a-snap-train-it-appears-in-the-edit

## Implementation notes

NB: I extracted some image related styled components into a folder called `images`. Makes the PR a bit bigger than it could be 

Following discussion with @jonathonherbert 
* these images will not be draggable / droppable
* main video will not be an option 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


ajwl:
@jennygrahamjones @jonathonherbert - have done a certain amount of work on this. Feel free to hack around redo. Back in a week!

client-v2/src/shared/components/article/ArticleBody.tsx
```diff
@@ -301,12 +297,11 @@ const articleBodyDefault = React.memo(
...
-                  {imageSlideshowReplace && 'Slideshow'}
-                  {imageReplace && 'Image replaced'}
-                  {imageCutoutReplace && 'Cutout replaced'}
-                  {showMainVideo && 'Show video'}
-                </ImageMetadataContainer>
+                <ImageMetadataContainer
+                  imageSlideshowReplace={imageSlideshowReplace}
+                  imageReplace={imageReplace}
+                  imageCutoutReplace={imageCutoutReplace}
+                ></ImageMetadataContainer>
```
jonathonherbert: Did we lose `showMainVideo` here?

client-v2/src/shared/components/image/ImageMetaDataContainer.tsx
```diff
@@ -0,0 +1,25 @@
...
+  imageCutoutReplace?: boolean;
+}
+
+export const ImageMetadataContainer = ({
+  imageSlideshowReplace,
+  imageReplace,
+  imageCutoutReplace
+}: ImageMetadataContainerProps) => (
+  <MetadataContainer>
+    {imageSlideshowReplace && 'Slideshow'}
```
jonathonherbert: 👍

client-v2/src/shared/components/image/ImagePlaceholder.tsx
```diff
@@ -0,0 +1,8 @@
+import BasePlaceholder from '../BasePlaceholder';
+import { styled } from 'shared/constants/theme';
+
+export default styled(BasePlaceholder)<{ width?: number; height?: number }>`
+  width: ${({ width }) => (width ? `${width}px` : '100%')};
```
jonathonherbert: double 👍

client-v2/src/shared/components/snapLink/SnapLink.tsx
```diff
@@ -121,7 +128,21 @@ const SnapLink = ({
...
           </CollectionItemHeadingContainer>
         </CollectionItemContent>
-        {size === 'default' && <ThumbnailSmall />}
+        <ImageWrapper>
+          <ThumbnailSmall
+            imageHide={article && article.imageHide}
+            url={article && article.thumbnail}
+          >
+            {article && article.thumbnail ? (
+              <ThumbnailCutout src={article && article.thumbnail} />
```
jonathonherbert: I don't think it's possible for a snap link to have a default cutout image -- these come from the `cutoutThumbnail` property, which is derived by `getContributorImage`.  So `ThumbnailCutout` here should be unnecessary.

client-v2/src/shared/components/snapLink/SnapLink.tsx
```diff
@@ -133,7 +140,17 @@ const SnapLink = ({
             </SnapLinkURL>
           </CollectionItemHeadingContainer>
         </CollectionItemContent>
-        {size === 'default' && <ThumbnailSmall />}
+        <ImageWrapper>
+          <ThumbnailSmall
+            imageHide={article && article.imageHide}
+            url={article && article.imageReplace ? article.imageSrcThumb : ''}
```
jennygrahamjones: `article.imageSrcThumb` seems to give the right src for the thumbnail, but without the check on `artice.imageReplace` the thumbnail image will persist even when the toggle is switched off. Not sure if there is a better way to handle this?

jonathonherbert: I think we're after `article.thumbnail` here, which is a property of `DerivedArticle` that displays whatever is current (taking into account overrides etc.)

jennygrahamjones: `article.thumbail` doesn't seem to contain any value, even after adding a new image from the Grid. Do we need to assign it somewhere?

jonathonherbert: Discussed IRL -- it turns out that at the point at which a `DerivedArticle` is created, `getThumbnail` doesn't get called if an external article isn't present. We should amend accordingly 👍

jennygrahamjones: I've had a stab at that, let me know what you think.

PR #956, opened 2019-08-20
fix isLive field calculation for external articles

akash1810:
## What's changed?

Related to https://github.com/guardian/facia-tool/pull/942.

**Before**
image

**After**
image

This also ensures the [`isLive` buttonProp](https://github.com/guardian/facia-tool/blob/580aab29cd21715131ca6a0bfcfe84bb4b98752e/client-v2/src/components/FrontsCAPIInterface/FeedItem.tsx#L214-L224) is correct for draft articles in the feed, meaning the [preview link](https://github.com/guardian/facia-tool/blob/63b1591f6aa95c4d48e58a1f1e514294a00d2be5/client-v2/src/shared/components/input/HoverActionButtons.tsx#L90-L96) goes to the correct place.

## Implementation notes

img

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/util/CAPIUtils.ts
+++ b/client-v2/src/util/CAPIUtils.ts
@@ -98,9 +98,9 @@ const getPrimaryTag = (externalArticle: ExternalArticle): Tag | null =>
 
 const isLive = (article: CapiArticle) => {
   // `isLive` is `undefined` if item is from Live CAPI, so return `true`
-  // if we're from Preview CAPI, coerce `isLive` (string) field into a Boolean
+  // `CapiBool` is either a `string` or `boolean`, so cast it as a string and compare 🤢
   const isLiveField = article.fields.isLive;
-  return isLiveField === undefined || Boolean(isLiveField);
+  return isLiveField === undefined || isLiveField.toString() === 'true';
 };
 
 const getArticleLabel = (article: CapiArticle) => {
--- a/client-v2/src/util/__tests__/CAPIUtils.spec.ts
+++ b/client-v2/src/util/__tests__/CAPIUtils.spec.ts
@@ -6,8 +6,10 @@ import {
 import {
   getIdFromURL,
   getContributorImage,
-  getThumbnail
+  getThumbnail,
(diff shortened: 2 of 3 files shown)


client-v2/src/util/CAPIUtils.ts
```diff
@@ -98,9 +98,9 @@ const getPrimaryTag = (externalArticle: ExternalArticle): Tag | null =>
 
 const isLive = (article: CapiArticle) => {
   // `isLive` is `undefined` if item is from Live CAPI, so return `true`
-  // if we're from Preview CAPI, coerce `isLive` (string) field into a Boolean
+  // `CapiBool` is either a `string` or `boolean`, so cast it as a string and compare 🤢
   const isLiveField = article.fields.isLive;
-  return isLiveField === undefined || Boolean(isLiveField);
+  return isLiveField === undefined || isLiveField.toString() === 'true';
```
sihil: What case does this catch that the other did and do we need to write a unit test for this given we can't get it right ??!?!

akash1810: The case where `isLiveField = 'false'` as `Boolean("false") === true` 🙄

client-v2/src/util/__tests__/CAPIUtils.spec.ts
```diff
@@ -203,4 +205,64 @@ describe('CAPIUtils', () => {
       ).toEqual('imageSrcThumb');
     });
   });
+
```
jonathonherbert: 👍

client-v2/src/util/CAPIUtils.ts
jonathonherbert: Incredible scenes

sihil: I totally thought we tested that :(

akash1810: Yeah... we're explicitly testing it now though!

PR #959, opened 2019-08-21
Add card image overrides

itsibitzi:
## What's changed?
Added new image fields to the form for cover card overrides, one for mobile and one for tablet.

Currently won't let you save if you've set only one of the two images.

Screenshot: 
image

And if you've got a missing image:
image

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/editions/EditionsArticle.scala
+++ b/app/model/editions/EditionsArticle.scala
@@ -18,11 +18,7 @@ object Image {
   implicit val format = Json.format[Image]
 }
 
-case class CoverCardImages(mobile: Option[Image], tablet: Option[Image]) {
-  def toPublishedCardImage: Option[PublishedCardImage] = {
-
-  }
-}
+case class CoverCardImages(mobile: Option[Image], tablet: Option[Image])
 
 object CoverCardImages {
   implicit val format = Json.format[CoverCardImages]
--- a/app/model/editions/EditionsArticle.scala
+++ b/app/model/editions/EditionsArticle.scala
@@ -69,7 +69,7 @@ case class EditionsArticle(pageCode: String, addedOn: Long, metadata: Option[Art
 
     val coverCardImages = metadata.flatMap { meta => {
       meta.mediaType match {
-        case Some(MediaType.OverrideCoverCard) => meta.coverCardImages.flatMap(_.toPublishedCardImage)
+        case Some(MediaType.CoverCard) => meta.coverCardImages.flatMap(_.toPublishedCardImage)
         case _ => None
       }
     }}
--- a/app/model/editions/EditionsArticle.scala
+++ b/app/model/editions/EditionsArticle.scala
@@ -1,5 +1,6 @@
 package model.editions
 
+import logging.Logging
 import play.api.libs.json.Json
 import scalikejdbc.WrappedResultSet
 
@@ -19,12 +20,7 @@ object Image {
 
 case class CoverCardImages(mobile: Option[Image], tablet: Option[Image]) {
   def toPublishedCardImage: Option[PublishedCardImage] = {
-    (mobile, tablet) match {
-      case (Some(m), Some(t)) => Some(PublishedCardImage(m.toPublishedImage, t.toPublishedImage))
-      case (None, Some(_)) => throw new IllegalStateException("Must define both mobile and tablet images")
-      case (Some(_), None) => throw new IllegalStateException("Must define both mobile and tablet images")
-      case _ => None
-    }
+
   }
 }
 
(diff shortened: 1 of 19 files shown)


build.sbt
```diff
@@ -105,6 +105,7 @@ libraryDependencies ++= Seq(
     "org.json4s" %% "json4s-native" % json4sVersion,
     "org.json4s" %% "json4s-jackson" % json4sVersion,
     "com.typesafe.play" %% "play-json-joda" % "2.6.9",
+    "ai.x" %% "play-json-extensions" % "0.40.2",
```
sihil: This fixes the 22 field limit?

itsibitzi: Yep!

test/model/editions/PublishedIssueTest.scala
```diff
@@ -86,40 +86,90 @@ class PublishedIssueTest extends FreeSpec with Matchers with OptionValues {
...
+
+      List(
+        EditionsArticle("123456", 0, Some(coverCardFurniture)),
+        EditionsArticle("123456", 0, Some(swapped))
+      ).foreach { a =>
+        assertThrows[IllegalStateException] {
+          a.toPublishedArticle
+        }
+      }
+    }
```
sihil: Instinctively I think we should not prevent exporting in this scenario. I would much prefer to log a `WARN` and then flip `MediaType` to `UseArticleTrail`. Thoughts @akash1810 ?

itsibitzi: We do validation on the client to prevent this from happening. If this breaks we should probably revert to the default so 24/7 don't have to get called.

akash1810: If we're guarding against this on save (on both client and server) I don't think we'd need to cater for this scenario.

Defaulting to `UseArticleTrail` is the simplest way to ensure we can still publish - we'd want to make this obvious on the client too (the warning message from the PR description goes some way to achieving this).

client-v2/src/util/form.ts
```diff
@@ -151,13 +157,20 @@ const formToMetaFieldMap: { [fieldName: string]: string } = {
...
 };
 
-export const getImageMetaFromValidationResponse = (image: ImageData) => ({
-  imageSrc: image.src,
-  imageSrcThumb: image.thumb,
-  imageSrcWidth: intToStr(image.width),
-  imageSrcHeight: intToStr(image.height),
-  imageSrcOrigin: image.origin
-});
+export const getImageMetaFromValidationResponse = (
```
akash1810: what's this diff for?

app/model/editions/EditionsArticle.scala
```diff
@@ -32,19 +39,39 @@ case class ArticleMetadata (
...
 
-  val default = ArticleMetadata(None, None, None, None, None, None, None, None, None, None, None)
+  val default = ArticleMetadata(None, None, None, None, None, None, None, None, None, None, None, None)
 }
 
-case class EditionsArticle(pageCode: String, addedOn: Long, metadata: Option[ArticleMetadata]) {
+case class EditionsArticle(pageCode: String, addedOn: Long, metadata: Option[ArticleMetadata]) extends Logging {
   def toPublishedArticle: PublishedArticle = {
-    val imageSrcOverride: Option[PublishedImage] = metadata.flatMap(meta => {
+    var mediaType: Option[MediaType] = metadata.flatMap(_.mediaType)
```
sihil: Been trying to figure out how to get rid of this `var`.

```
  def toPublishedArticle: PublishedArticle = {
    val (maybeCoverCard, publishedMediaType): (Option[PublishedCardImage], PublishedMediaType) = metadata.map { meta =>
      (meta.mediaType, meta.coverCardImages.flatMap(_.mobile), meta.coverCardImages.flatMap(_.tablet)) match {
        case (Some(MediaType.CoverCard), Some(mobileCoverCard), Some(tabletCoverCard)) =>
          val coverCard = PublishedCardImage(mobileCoverCard.toPublishedImage, tabletCoverCard.toPublishedImage)
          meta.copy()
          (Some(coverCard), PublishedMediaType.CoverCard)
        case (Some(MediaType.CoverCard), _, _) =>
          logger.warn(s"Failed to convert card images, must define both mobile and tablet images, falling back to use article trail")
          (None, PublishedMediaType.UseArticleTrail)
        case (Some(otherMediaType), _, _) =>
          (None, otherMediaType.toPublishedMediaType)
      }
    } getOrElse (None, PublishedMediaType.UseArticleTrail)

    val imageSrcOverride: Option[PublishedImage] = metadata.flatMap(meta => {
      publishedMediaType match {
        case PublishedMediaType.Image => meta.replaceImage.map(_.toPublishedImage)
        case PublishedMediaType.Cutout => meta.cutoutImage.map(_.toPublishedImage)
        case _ => None
      }
    })

    PublishedArticle(
      pageCode.toLong,
      PublishedFurniture(
        kicker = metadata.flatMap(_.customKicker),
        headlineOverride = metadata.flatMap(_.headline),
        trailTextOverride = metadata.flatMap(_.trailText),
        bylineOverride = metadata.flatMap(_.byline),
        showByline = metadata.flatMap(_.showByline).getOrElse(false),
        showQuotedHeadline = metadata.flatMap(_.showQuotedHeadline).getOrElse(false),
        mediaType = publishedMediaType,
        imageSrcOverride = imageSrcOverride,
        sportScore = metadata.flatMap(_.sportScore),
        overrideArticleMainMedia = metadata.flatMap(_.overrideArticleMainMedia).getOrElse(true),
        coverCardImages = maybeCoverCard
      )
    )
  }
```

PR #960, opened 2019-08-21
Display sublinks when form is open

jennygrahamjones:
## What's changed?

Previously, when an inline form was selected/opened, it would hide any sub-links that had been added to the article. Now, the sub-links are visible and can be interacted in the same way whether the form is open or closed. 

The sub-links attached to an article displaying beneath it, while the form is open:
Screenshot 2019-08-21 at 11 40 01

Same as above, but the sub-links component has been expanded:
Screenshot 2019-08-21 at 11 40 19

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/CollectionComponents/CollectionItem.tsx
+++ b/client-v2/src/components/FrontsEdit/CollectionComponents/CollectionItem.tsx
@@ -114,6 +114,17 @@ class CollectionItem extends React.Component<ArticleContainerProps> {
       canShowPageViewData = false
     } = this.props;
 
+    const getSublinks = () => {
+      return (
+        <Sublinks
+          numSupportingArticles={numSupportingArticles}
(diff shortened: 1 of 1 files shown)


client-v2/src/components/FrontsEdit/CollectionComponents/CollectionItem.tsx
```diff
@@ -113,6 +113,17 @@ class CollectionItem extends React.Component<ArticleContainerProps> {
       canShowPageViewData = false
     } = this.props;
 
+    const getSublinks = () => {
```
jonathonherbert: This can be an assignment, no need for a lambda -- 

```
const sublinks = <Sublinks ...etc />

... later, e.g.

<div>{sublinks}</div>
```

PR #961, opened 2019-08-21
make modal more generic

itsibitzi:
## What's changed?
Make the modal accept call back functions instead of `Action[]` - this will allow us to do things with the confirmation modal which don't affect redux at all.

### DIFF
--- a/client-v2/src/actions/ArticleFragments.ts
+++ b/client-v2/src/actions/ArticleFragments.ts
@@ -38,6 +38,7 @@ import { ValidationResponse } from 'shared/util/validateImageSrc';
 import { MappableDropType } from 'util/collectionUtils';
 import { selectWillCollectionHitCollectionCap } from 'selectors/collectionSelectors';
 import { batchActions } from 'redux-batched-actions';
+import noop from 'lodash/noop';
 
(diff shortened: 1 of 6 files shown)


client-v2/src/actions/ArticleFragments.ts
```diff
@@ -124,19 +125,29 @@ const maybeInsertGroupArticleFragment = (
...
+              .concat([
+                insertGroupArticleFragment(id, index, articleFragmentId),
+                maybeAddFrontPublicationDate(articleFragmentId),
+                addPersistMetaToAction(capGroupSiblings, {
+                  id: articleFragmentId,
+                  persistTo,
+                  applyBeforeReducer: true
+                })(id, collectionCap)
+              ])
+              .forEach(action => dispatch(action));
```
akash1810: why are we dispatching now?

itsibitzi: We're moving away from sending arrays of actions for the modal to dispatch on completion to using just a call back.

The downside of this is time travelling through redux wont work very well but following a conversation with Jon we intend to move over to thread's way of handling modals.

PR #963, opened 2019-08-21
Switch issue_date to be a LocalDate instead of DateTime

sihil:
## What's changed?
This PR changes the issueDate in the model to be a date instead of a date time. This is easier to reason about as it is a better fit and means that there are fewer conversions from and to date times in the code base.

## Implementation notes
We added tests for the DB evolution in question (`6.sql`). These tests use raw SQL as we are changing the model so can't use the models in this instance.

The implemetation fell out of the model changes nicely and eliminated several conversions.

The type on the client has changed from epoch to `YYYY-MM-dd` and the client just seems to have carried on working since Date(x) works the same for both.

## Checklist
 - [ ] Talk with @AWare to see how happy he is with the new export format

### General
- [X] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/editions/EditionsIssue.scala
+++ b/app/model/editions/EditionsIssue.scala
@@ -1,6 +1,6 @@
 package model.editions
 
-import java.time.{Instant, OffsetDateTime, ZoneId}
+import java.time.{Instant, LocalDate, OffsetDateTime, ZoneId}
 
 import play.api.libs.json.Json
 import scalikejdbc.WrappedResultSet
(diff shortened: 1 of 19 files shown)


RichieAHB:
Just been through this IRL with @sihil and it seems that `issueDate` is only being used in `renderIssueDate` which calls `new Date(issueDate)`, which should handle this change 👌

itsibitzi:
github being flakey D:

sihil:
I've just rebased this and pushed another tiny client side change which I forgot to push yesterday. GH seems to be borked this morning though...

PR #970, opened 2019-08-22
Bug: display boost toggle if story is boosted

jennygrahamjones:
## What's changed?

When stories have been boosted and are then moved into a collection that doesn't support boosting, the story still shows the Boost label but provides no way for users to toggle this off.

This change will ensure that the toggle still appears, regardless of the stories location, until the toggle has been switched off. At that time, it will disappear, unless the story is moved back into a collection that supports boosting.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


ajwl:
The collection can be allowed to have boost if you go into config and set the layout to dynamic.

PR #977, opened 2019-08-23
Unbreak breaking news

jonathonherbert:
## What's changed?

Breaking news would still redirect to `v2` as the code currently stands. But there's no breaking news in v2! This PR makes an exception of the `breaking-news` vanity url and the breaking-news front, which incidentally is the way the fronts tool handles breaking news.

It also amends the handover time to next Tuesday, 9:30PM

Tested with different times and modes in DEV.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/ViewsController.scala
+++ b/app/controllers/ViewsController.scala
@@ -12,15 +12,13 @@ import util.{Acl, Encryption}
 class ViewsController(val acl: Acl, assetsManager: AssetsManager, isDev: Boolean,
                       crypto: Encryption, val deps: BaseFaciaControllerComponents)(implicit ec: ExecutionContext) extends BaseFaciaController(deps) {
 
-  private def shouldRedirectToV2(request: UserRequest[AnyContent]): Boolean = {
-    val shouldRedirectByQs: Boolean = (request.queryString.getOrElse("redirect", Seq("true")).contains("true")
-      && !request.queryString.getOrElse("layout", Seq("")).exists(_.contains("breaking-news")))
-
-    // TODO set correct date
-    val redirectDay = DateTime.parse("2019-08-27T09:30")
-    val shouldRedirectByDate: Boolean = DateTime.now().isAfter(redirectDay)
-
-    shouldRedirectByQs && (isDev || shouldRedirectByDate)
+  private def shouldRedirectToV2(request: UserRequest[AnyContent], priority: Option[String] = None): Boolean = {
+    val isBreakingNews = priority.getOrElse("") == "breaking-news" || request.queryString.getOrElse("layout", Seq("")).exists(_.contains("breaking-news"))
+    if (isBreakingNews) {
+      false
+    } else {
+      request.queryString.getOrElse("redirect", Seq("true")).contains("true")
+    }
   }
 
   def priorities() = AccessAuthAction { request =>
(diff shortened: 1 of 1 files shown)


app/controllers/ViewsController.scala
```diff
@@ -13,10 +13,11 @@ class ViewsController(val acl: Acl, assetsManager: AssetsManager, isDev: Boolean
...
-    val shouldRedirectByQs: Boolean = request.queryString.getOrElse("redirect", Seq("true")).contains("true")
+    val shouldRedirectByQs: Boolean = (request.queryString.getOrElse("redirect", Seq("true")).contains("true")
+      && !request.queryString.getOrElse("layout", Seq("")).exists(_.contains("breaking-news")))
 
     // TODO set correct date
-    val redirectDay = DateTime.parse("2019-12-25")
+    val redirectDay = DateTime.parse("2019-08-27T09:30")
     val shouldRedirectByDate: Boolean = DateTime.now().isAfter(redirectDay)
 
     shouldRedirectByQs && (isDev || shouldRedirectByDate)
```
akash1810: `isDev` can go now.

app/controllers/ViewsController.scala
```diff
@@ -34,7 +35,7 @@ class ViewsController(val acl: Acl, assetsManager: AssetsManager, isDev: Boolean
   }
 
   def collectionEditor(priority: String) = getCollectionPermissionFilterByPriority(priority, acl)(ec) { request =>
-    if (shouldRedirectToV2(request)) {
+    if (priority != "breaking-news" && shouldRedirectToV2(request)) {
```
aug24: I'd like the priority check moved into the shouldRedirect function.

app/controllers/ViewsController.scala
jonathonherbert: 👍

app/controllers/ViewsController.scala
jonathonherbert: 👍

app/controllers/ViewsController.scala
```diff
@@ -12,14 +12,13 @@ import util.{Acl, Encryption}
...
-  private def shouldRedirectToV2(request: UserRequest[AnyContent]): Boolean = {
-    val shouldRedirectByQs: Boolean = request.queryString.getOrElse("redirect", Seq("true")).contains("true")
-
-    // TODO set correct date
-    val redirectDay = DateTime.parse("2019-12-25")
-    val shouldRedirectByDate: Boolean = DateTime.now().isAfter(redirectDay)
-
-    shouldRedirectByQs && (isDev || shouldRedirectByDate)
+  private def shouldRedirectToV2(request: UserRequest[AnyContent], priority: Option[String] = None): Boolean = {
+    val isBreakingNews = priority.getOrElse("") == "breaking-news" || request.queryString.getOrElse("layout", Seq("")).exists(_.contains("breaking-news"))
```
akash1810: curious to know how `/breaking-news` works w/out an explicit play route...

jonathonherbert: it does have a route in `routes` -- see `VanityRedirects.scala`

PR #983, opened 2019-08-28
Path type

mkuzdowicz:
## What's changed?
Introducing PathType enum and prop in Postgres that will enable us to switch between different paths for prefill queries
This PR includes switching to `/search` from `/content/print-sent` path while getting prefill for crosswords

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### DIFF
--- a/app/conf/Configuration.scala
+++ b/app/conf/Configuration.scala
@@ -202,8 +202,6 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
 
     val contentApiLiveHost: String = getMandatoryString("content.api.host")
     def contentApiDraftHost: String = getMandatoryString("content.api.draft.iam-host")
-
-    val editionsPrefillHost: String = getMandatoryString("content.api.draft.iam-host")
     lazy val editionsKey: String = getMandatoryString("content.api.editions.apiKey")
 
     lazy val key: Option[String] = getString("content.api.key")
--- a/app/conf/Configuration.scala
+++ b/app/conf/Configuration.scala
@@ -203,7 +203,7 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
     val contentApiLiveHost: String = getMandatoryString("content.api.host")
     def contentApiDraftHost: String = getMandatoryString("content.api.draft.iam-host")
 
-    val editionsPrefillHost: String = getMandatoryString("content.api.editions.prefillHost")
+    val editionsPrefillHost: String = getMandatoryString("content.api.draft.iam-host")
     lazy val editionsKey: String = getMandatoryString("content.api.editions.apiKey")
 
     lazy val key: Option[String] = getString("content.api.key")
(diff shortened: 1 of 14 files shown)


akash1810:
This looks sensible to me, however I can't approve.

app/conf/Configuration.scala
```diff
@@ -202,8 +202,6 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
 
     val contentApiLiveHost: String = getMandatoryString("content.api.host")
     def contentApiDraftHost: String = getMandatoryString("content.api.draft.iam-host")
-
-    val editionsPrefillHost: String = getMandatoryString("content.api.editions.prefillHost")
     lazy val editionsKey: String = getMandatoryString("content.api.editions.apiKey")
```
itsibitzi: This can go too I think

app/services/Capi.scala
```diff
@@ -122,6 +120,10 @@ class GuardianCapi(config: ApplicationConfiguration)(implicit ex: ExecutionConte
       .showBlocks("main")
       .showAtoms("media")
 
+    import play.api.Logger
+
+    Logger.info(s"prefill query URL generated:\n${query.getUrl(targetUrl)}")
```
itsibitzi: Was this intended to be left here?

app/conf/Configuration.scala
mkuzdowicz: i tried to remove the apiKey and it was not working locally without it

app/services/Capi.scala
mkuzdowicz: a sorry that should go away :) 
Debugging leftover

mkuzdowicz: done

app/conf/Configuration.scala
mkuzdowicz: Regarding JSONB question, i think it will make sense to go with JSONB if we will need to include more params for the query object in the future, for that one the model is quite simple so i don't think we paid a lot of attention to decide if we should go with JSONB or additional field

One of the advantage of not using JSONB is not having issue with serialisation of past state and not being forced to use Option[] fields as default for non optional params ;)

We can go with JSONB if that will be better but that will require some changes

itsibitzi: Huh. What was it that broke?

itsibitzi: Had a chat IRL - it seems like the CAPI Scala client requires a key to be provided so we'll leave this piece of config in for now.

conf/evolutions/default/7.sql
```diff
@@ -0,0 +1,10 @@
+# --- !Ups
+
+ALTER TABLE collections ADD COLUMN path_type TEXT;
+UPDATE  collections  SET path_type = 'printSent' where prefill IS NOT NULL;
```
sihil: Should this be `where prefill IS NULL`? 
Are there tests for this evolution?
Should the `path_type` column be altered to be NOT NULL?

PR #986, opened 2019-08-28
Limit textarea auto-expansion to a given height, but allow users to expand beyond it

jonathonherbert:
## What's changed?

This PR limits  the textarea auto-expansion to a given height, but allow users to expand beyond it if they'd like.

textarea

At the moment, users can't expand the textarea for trail text beyond its max height. This is useful when inspecting lots of content for e.g. e-mail fronts, but we'd still like to keep the old behaviour of auto-expansion as users type, with a limit so the area doesn't grow too large. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


jonathonherbert:
@akash1810 you are correct, I've just rebased 👍

PR #989, opened 2019-08-31
Add swatch and fronts for 'Big Serious Things'

blishen:
## What's changed?
https://trello.com/c/S5azk9Il/452-front-tool-support-for-big-news

This PR extends swatches to introduce "Special" as a swatch. This is a new palette of colours that will be introduced in the app for BIG SERIOUS THINGS.

In addition it adds two new special fronts, and reconfigures the special fronts that straddle "Top Stories" to use the "Special" swatch.

## Implementation notes
To be honest I thought we defined both the pillar value and the swatch for the fronts, but it seems like I was wrong about that. I don't think this is problematic, but I will check in with Alex W & Co.

### DIFF
--- a/app/model/editions/templates/DailyEdition.scala
+++ b/app/model/editions/templates/DailyEdition.scala
@@ -10,6 +10,7 @@ object DailyEdition {
     List(
       FrontSpecialSpecial1.front -> Daily(),
       FrontTopStories.front -> Daily(),
+      FrontSpecialSpecial2.front -> Daily(),
       FrontNewsUkGuardian.front -> WeekDays(List(WeekDay.Mon, WeekDay.Tues, WeekDay.Wed, WeekDay.Thurs, WeekDay.Fri)),
       FrontNewsUkGuardianSaturday.front -> WeekDays(List(WeekDay.Sat)),
       FrontNewsWorldGuardian.front -> WeekDays(List(WeekDay.Mon, WeekDay.Tues, WeekDay.Wed, WeekDay.Thurs, WeekDay.Fri, WeekDay.Sat)),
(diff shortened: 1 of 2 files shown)


app/model/editions/templates/DailyEdition.scala
```diff
@@ -580,6 +598,22 @@ object FrontFoodObserver {
...
+  val collectionSpecialSpecial3 = CollectionTemplate(
+    name = "Special",
+    prefill = None,
+    presentation = TemplateDefaults.defaultCollectionPresentation
+  )
+
+  val front = FrontTemplate(
+    name = "Special 3",
+    collections = List(collectionSpecialSpecial3),
+    presentation = TemplateDefaults.defaultFrontPresentation,
```
sihil: Is this right? A special front with default formatting?

app/model/editions/templates/DailyEdition.scala
```diff
@@ -618,16 +652,16 @@ object FrontSportObserver {
...
     prefill = Some(CapiPrefillQuery("?tag=theguardian/special-supplement/special-supplement|theobserver/special-supplement/special-supplement", PathType.PrintSent)),
     presentation = TemplateDefaults.defaultCollectionPresentation
   )
 
   val front = FrontTemplate(
-    name = "Special 2",
-    collections = List(collectionSpecialSpecial2),
+    name = "Special 4",
+    collections = List(collectionSpecialSpecial4),
     presentation = TemplateDefaults.defaultFrontPresentation,
```
sihil: Ditto - should this be the default formatting?

blishen: I think it's okay - default formatting gives the neutral palette (I think!) - it's only the two specials that straddle Top stores that want the new "special" palette. Perhaps I should name them differently?

app/model/editions/templates/DailyEdition.scala
blishen: Yup - only the two that straddle top stories get the "special" swatch. I suppose it is a bit confusing, but I'd rather not change the special names.

PR #994, opened 2019-09-05
Copy instructions to press locally over to the readme

Reettaphant:
## What's changed?

The guide to fronts file described fronts v1 and wasn't linked to from anywhere so I've copied over a section on how to check that fronts are still pressing after making changes and how to troubleshoot if fronts are failing to press.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


ajwl:
Thank you @Reettaphant. One issue I have is that I always jump over the v1 documentation to the v2 documentation so I missed the `m.code` trick - which is very useful. I guess this will be deleted when v1 goes but ideally we could split out what is v1 specific and highlight what is general to both parts of the tool?

ajwl:
I mean - this can also be done in another PR.. so 👍  can approve for now

Reettaphant:
Good point, I've moved over everything to do with pressing fronts to the readme so people can have a full understanding of how it works, everything else in that file is related to v1 so I think once v1 code is gone it can go too. I'll leave anything else that needs to be cleaned up for another pr.

PR #996, opened 2019-09-06
Add polling for articles

jonathonherbert:
## What's changed?

Adds polling for articles. The tool will only ask for changes to articles that are in open collections. As we re-fetch articles when their parent collections are opened, nothing on screen should be staler than the polling window, which is currently 30s.

Incidentally, this also lengthens the Ophan polling window to 300s, as it was in v1. Redrawing Ophan graphs is quite expensive and the previous window of 10s was arguably too eager!

## Implementation notes

There are quite a few changes here, as I felt there was a lot of logic in the polling config that better belonged elsewhere -

- Selector logic has been refactored into selectors
- Untested logic has yummy tests 🎉 
- I've removed the dependency on `priority` props, as we can derive that from the application state.

As a result the changelist is quite large -- I'd be happy to work through the PR or split it into separate PRs if anything is unclear.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/constants/polling.ts
```diff
@@ -0,0 +1,3 @@
+export const collectionsPollInterval = 10000;
+export const articlesPollInterval = 30000;
+export const ophanPollInterval = 300000;
```
jennygrahamjones: Could we do an initial poll for data once collections have loaded, and then default to this longer polling interval?

client-v2/src/redux/modules/pageViewData/actions.ts
```diff
@@ -61,12 +61,12 @@ const convertToStoriesData = (
...
   ophanData: PageViewDataFromOphan,
-  articles: ArticlePathAndId[]
+  articles: DerivedArticle[]
 ): string => {
   // the path from ophan has a slash at the front, removing below
   const ophanPathClean = ophanData.path.substr(1);
-  const matchingArticle = articles.find(a => a.articlePath === ophanPathClean)!;
-  return matchingArticle.articleId;
+  const matchingArticle = articles.find(a => a.urlPath === ophanPathClean)!;
+  return matchingArticle.id;
```
jennygrahamjones: This appears to be returning as `undefined` and leading to no graph data displaying

client-v2/src/constants/polling.ts
jonathonherbert: Thanks for noticing, this is really important to address 👍

client-v2/src/redux/modules/pageViewData/actions.ts
jonathonherbert: Gotcha, it's likely a result of me removing the intermediate type `ArticleWithPath` or similar.

PR #999, opened 2019-09-09
Add edition field to published issue in preparation to switch from name

sihil:
## What's changed?

The field `name` is not very descriptive so lets make it better. (This previously said `id` which just goes to prove the point as that was completely the wrong field).

## Implementation notes

This also introduces an edition enum to make it slightly typier to work with. Finally a regression test was introduced as the enum change modified the behaviour of the location to which the file was written in S3. Unfortunately the editions code relies on the path layout.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/editions/PublishedIssue.scala
+++ b/app/model/editions/PublishedIssue.scala
@@ -50,8 +50,8 @@ case class PublishedFront(id: String, name: String, collections: List[PublishedC
 
 case class PublishedIssue(
   id: String, // TODO: Not sure we should be leaking our internal ID here...
-  name: String, // TODO: remove this once downstream is consuming 'edition'
-  edition: String,
+  name: Edition, // TODO: remove this once downstream is consuming 'edition'
+  edition: Edition,
   issueDate: LocalDate,
   version: String,
   fronts: List[PublishedFront]
--- a/app/model/editions/PublishedIssue.scala
+++ b/app/model/editions/PublishedIssue.scala
@@ -49,9 +49,9 @@ case class PublishedCollection(id: String, name: String, items: List[PublishedAr
 case class PublishedFront(id: String, name: String, collections: List[PublishedCollection], swatch: Swatch)
 
 case class PublishedIssue(
-  id: String, // TODO: remove this downstream is consuming 'edition'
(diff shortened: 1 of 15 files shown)


sihil:
Oh bah! I thought this was `daily-edition`, but it is not! This makes me sad and I will rework this PR so that edition does indeed do this.

sihil:
How about something like this @akash1810. I'd really like to get this out before #1025 is merged @aug24 so that the edition names are correct in the database and I don't have to do anything smarter than setting the template name to `daily-edition` for all existing lines in the DB.

app/model/editions/PublishedIssue.scala
```diff
@@ -49,8 +49,9 @@ case class PublishedCollection(id: String, name: String, items: List[PublishedAr
 case class PublishedFront(id: String, name: String, collections: List[PublishedCollection], swatch: Swatch)
 
 case class PublishedIssue(
-  id: String,
-  name: String,
+  id: String, // TODO: Not sure we should be leaking our internal ID here...
+  name: String, // TODO: remove this once downstream is consuming 'edition'
+  edition: String,
```
akash1810: minor: feels like we can make `edition` an enum to be stricter.

sihil: Done. This will definitely have conflicts with #1025. Maybe get that in first and then merge this second.

build.sbt
```diff
@@ -57,6 +57,9 @@ TwirlKeys.templateImports ++= Seq(
     "play.api.Play.current"
 )
 
+// include the enum path bindables
+routesImport += "model.editions._"
```
akash1810: ooh, tell me more!

sihil: This enables you to use enums as bindables in routes: https://github.com/guardian/facia-tool/pull/999/files#diff-93c1ebbcb555c5f7d8ee08519974efa6R102 
PlayEnum generates them for free. Then you don't even need to check that the use has provided a sane value as the framework rejects anything that isn't with the appropriate 4xx error.

sihil:
I'll push this out first thing tomorrow rather than put tonights shift at risk.

PR #1001, opened 2019-09-11
add issue deletion endpoint

akash1810:
## What's changed?

When an issue is deleted, we'll log the issue id, creation date and who is performing the deletion.

## Implementation notes

There's no UI... unless you count `fetch`:

```javascript
await fetch("/editions-api/issues/<ID>", {"credentials": "include", "method": "DELETE"});
```

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/test/services/editions/db/EditionsDBTest.scala
+++ b/test/services/editions/db/EditionsDBTest.scala
@@ -393,7 +393,7 @@ class EditionsDBTest extends FreeSpec with Matchers with EditionsDBService with 
       specialFront2.isHidden shouldBe false
     }
 
-    "should delete an issue" taggedAs UsesDatabase in {
+    "should delete an issue and cascade the deletion to related entities" taggedAs UsesDatabase in {
       val issue = insertSkeletonIssue(2019, 9, 1,
         front("news/uk",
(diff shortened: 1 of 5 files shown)


test/services/editions/db/EditionsDBTest.scala
```diff
@@ -393,6 +393,30 @@ class EditionsDBTest extends FreeSpec with Matchers with EditionsDBService with
...
+            article("23456")
+          )
+        )
+      )
+
+      val dbIssue: EditionsIssue = editionsDB.getIssue(issue).value
+      val frontIds = dbIssue.fronts.map(_.id)
+      frontIds.length shouldBe 1
+
+      editionsDB.getFrontFromIssueId(dbIssue.id) shouldNot be
```
sihil: This is a thing that you only use in a test right? In which case I'd move the SQL into the test file.

sihil: Possibly inline...

akash1810: Updated now.

PR #1009, opened 2019-09-16
PR for pulling messages from SQS for issue publish events

mkuzdowicz:
## What's changed?
PR for pulling messages from SQS for issue publish events
With test coverage for parsing the SNS messages from SQS to relevant data that may be use-full
and
deleting received messages from queue if they were successfully updated in DB

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### DIFF
--- a/app/conf/Configuration.scala
+++ b/app/conf/Configuration.scala
@@ -229,6 +229,8 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
   object faciatool {
     lazy val breakingNewsFront = "breaking-news"
     lazy val frontPressToolQueue = getString("frontpress.sqs.tool_queue_url")
+    lazy val issuePublishedEventsQueue = getString("publish_events.queue_url")
+      .getOrElse(throw new BadConfigurationException(s"property of type string not configured for `publish_events.queue_url` for stage $stageFromProperties"))
     lazy val showTestContainers = getBoolean("faciatool.show_test_containers").getOrElse(false)
     lazy val stsRoleToAssume = getString("faciatool.sts.role.to.assume").getOrElse(stsRoleToAssumeFromProperties)
     lazy val frontPressUpdateTable = frontPressedDynamoTable
--- a/app/conf/Configuration.scala
+++ b/app/conf/Configuration.scala
@@ -229,8 +229,7 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
   object faciatool {
     lazy val breakingNewsFront = "breaking-news"
     lazy val frontPressToolQueue = getString("frontpress.sqs.tool_queue_url")
-    lazy val issuePublishedEventsQueue = getString("publish_events.queue_url")
-      .getOrElse(throw new BadConfigurationException(s"property of type string not configured for `publish_events.queue_url` for stage $stageFromProperties"))
+    lazy val publishEventsQueue = getMandatoryString("publish_events.queue_url")
     lazy val showTestContainers = getBoolean("faciatool.show_test_containers").getOrElse(false)
     lazy val stsRoleToAssume = getString("faciatool.sts.role.to.assume").getOrElse(stsRoleToAssumeFromProperties)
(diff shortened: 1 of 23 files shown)


app/conf/Configuration.scala
```diff
@@ -229,6 +229,7 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
   object faciatool {
     lazy val breakingNewsFront = "breaking-news"
     lazy val frontPressToolQueue = getString("frontpress.sqs.tool_queue_url")
+    lazy val issuePublishedEventsQueue = getString("issue_published_events.queue_url").getOrElse("")
```
akash1810: Should we throw if we're unable to get this value from config?

app/services/editions/SQSFacade.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.amazonaws.regions.Regions
+import com.amazonaws.services.sqs.AmazonSQSAsyncClientBuilder
+import com.amazonaws.services.sqs.model.{Message, ReceiveMessageRequest}
+import conf.ApplicationConfiguration
+import logging.Logging
+
+import scala.collection.JavaConverters._
+import scala.util.{Failure, Success, Try}
+
+object SQSFacade {
```
akash1810: Would `SQSClient` be more intention revealing?

app/services/editions/publishing/IssuePublishedSNSMessageParser.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import scala.util.Try
+
+object IssuePublishedSNSMessageParser extends Logging {
+
+  import IssuePublishedSQSMsgFormatter._
+
+  def parseMsg(msgWithSNSNotification: Message): Try[IssuePublishedEventWrapper] = {
+    logger.info("read new publish events")
+    import Json.{fromJson, parse}
```
akash1810: any reason to have a local import here, given we have `import play.api.libs.json.Json` up top?

app/conf/Configuration.scala
mkuzdowicz: yeah we can use .get that will throw exception

app/services/editions/SQSFacade.scala
mkuzdowicz: i changed name of that class to SQSIssuePublishEventsQueueFacade because it will be specific for messages from issue events queue only

mkuzdowicz: that is why it is facade because it will hide all the complexity related to reading from that specific queue not generic one like max messages or wait-time

app/services/editions/publishing/IssuePublishedSNSMessageParser.scala
mkuzdowicz: sure

app/conf/Configuration.scala
mkuzdowicz: done

app/services/editions/publishing/IssuePublishedSNSMessageParser.scala
mkuzdowicz: done

app/conf/Configuration.scala
```diff
@@ -229,6 +229,8 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
   object faciatool {
     lazy val breakingNewsFront = "breaking-news"
     lazy val frontPressToolQueue = getString("frontpress.sqs.tool_queue_url")
+    lazy val issuePublishedEventsQueue = getString("publish_events.queue_url")
+      .getOrElse(throw new BadConfigurationException(s"property of type string not configured for `publish_events.queue_url` for stage $stageFromProperties"))
```
akash1810: minor: we can reuse [`getMandatoryString`](https://github.com/guardian/facia-tool/blob/master/app/conf/Configuration.scala#L41-L42) here

mkuzdowicz: good spot, thanks

PR #1011, opened 2019-09-16
new table to track publication status of an issue

akash1810:
## What's changed?

Track the publication status of an issue in a new table.

When we publish an issue, insert a new row into a new table. The record will get [updated in the future](https://github.com/guardian/facia-tool/pull/1009) by an event from the Archiver lambda.

Also add an endpoint to view the status. This will allow us to display the progress of publication in the UI.

## Implementation notes

We probably want to move to a GUID rather than a date string, however the downstream services rely on it being a date string. To promote this move, however, I've pushed the problem next to the query to make it simpler to update later.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/editions/EditionsIssue.scala
+++ b/app/model/editions/EditionsIssue.scala
@@ -18,7 +18,9 @@ case class EditionsIssue(
     launchedEmail: Option[String],
     fronts: List[EditionsFront]
 ) {
-  def toPublishedIssue(publicationEventId: Option[String] = None): PublishedIssue = PublishedIssue(
+  def toPreviewIssue(): PublishedIssue = toPublishedIssue("preview")
+
+  def toPublishedIssue(publicationEventId: String): PublishedIssue = PublishedIssue(
     id,
     displayName,
     issueDate,
--- a/app/model/editions/EditionsIssue.scala
+++ b/app/model/editions/EditionsIssue.scala
@@ -18,11 +18,11 @@ case class EditionsIssue(
     launchedEmail: Option[String],
     fronts: List[EditionsFront]
 ) {
-  def toPublishedIssue(version: Option[String] = None): PublishedIssue = PublishedIssue(
+  def toPublishedIssue(publicationEventId: Option[String] = None): PublishedIssue = PublishedIssue(
     id,
     displayName,
     issueDate,
-    version,
+    publicationEventId,
     fronts
       .filterNot(_.isHidden) // drop hidden fronts
       .map(_.toPublishedFront) // convert
--- a/app/model/editions/EditionsIssue.scala
+++ b/app/model/editions/EditionsIssue.scala
@@ -20,11 +20,11 @@ case class EditionsIssue(
 ) {
   def toPreviewIssue(): PublishedIssue = toPublishedIssue("preview")
 
-  def toPublishedIssue(publicationEventId: String): PublishedIssue = PublishedIssue(
+  def toPublishedIssue(version: String): PublishedIssue = PublishedIssue(
     id,
     displayName,
     issueDate,
-    publicationEventId,
+    version,
     fronts
       .filterNot(_.isHidden) // drop hidden fronts
       .map(_.toPublishedFront) // convert
--- a/app/model/editions/EditionsPublicationHistory.scala
+++ b/app/model/editions/EditionsPublicationHistory.scala
@@ -1,0 +1,28 @@
+package model.editions
+
+import play.api.libs.json.Json
+import scalikejdbc.WrappedResultSet
+
+case class EditionsPublicationHistory(
+  id: String,
+  status: PublicationStatus,
+  launchedOn: Long,
+  launchedBy: String,
+  launchedEmail: String,
+  publishedOn: Option[Long],
+  message: Option[String]
+)
+
+object EditionsPublicationHistory {
+  implicit val writes = Json.writes[EditionsPublicationHistory]
+
+  def fromRow(rs: WrappedResultSet): EditionsPublicationHistory = EditionsPublicationHistory(
+    rs.string("id"),
+    PublicationStatus.withName(rs.string("status")),
+    rs.zonedDateTime("launched_on").toInstant.toEpochMilli,
+    rs.string("launched_by"),
+    rs.string("launched_email"),
+    rs.zonedDateTimeOpt("published_on").map(_.toInstant.toEpochMilli),
+    rs.stringOpt("message"),
+  )
+}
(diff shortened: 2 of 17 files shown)


conf/evolutions/default/8.sql
```diff
@@ -0,0 +1,15 @@
+# --- !Ups
+
+CREATE TABLE edition_issues_publication_history (
```
akash1810: `publication_history` might be a simpler name for this table

mkuzdowicz: if issue is the only entity that can be published

app/model/editions/EditionsPublicationHistory.scala
```diff
@@ -0,0 +1,28 @@
...
+case class EditionsPublicationHistory(
+  id: String,
+  status: PublicationStatus,
+  launchedOn: Long,
+  launchedBy: String,
+  launchedEmail: String,
+  publishedOn: Option[Long],
+  message: Option[String]
+)
+
```
mkuzdowicz: is the id event_id ? I think we will need both event_id and issue_id

app/services/editions/publishing/EditionsPublishing.scala
```diff
@@ -15,13 +15,12 @@ class EditionsPublishing(publishedBucket: EditionsBucket, previewBucket: Edition
   }
 
   def publish(issue: EditionsIssue, user: User, now: OffsetDateTime) = {
-    val publishedIssue = issue.toPublishedIssue(Some(now.format(DateTimeFormatter.ISO_DATE_TIME)))
+    // Bump the recently published counters
+    val publishId = db.publishIssue(issue.id, user, now)
+
```
mkuzdowicz: do we want to have some Try and logging while talking to DB ?

app/services/editions/publishing/EditionsBucket.scala
```diff
@@ -11,14 +11,15 @@ class EditionsBucket(s3Client: AmazonS3, bucketName: String) {
...
     s"$keyname.json"
   }
 
   def putIssue(issue: PublishedIssue) = {
     val issueJson = Json.stringify(Json.toJson(issue))
     val metadata = new ObjectMetadata()
     metadata.setContentType("application/json")
+    issue.publicationEventId.foreach(metadata.addUserMetadata("publication-event-id", _))
     val key = s"${createIssuePrefix(issue)}/${createIssueFilename(issue)}"
     val request = new PutObjectRequest(bucketName, key, new StringInputStream(issueJson), metadata)
```
mkuzdowicz: i think for custom metadata in s3 there is a convention to start with `x-`: https://docs.aws.amazon.com/AmazonS3/latest/user-guide/add-object-metadata.html

app/model/editions/EditionsIssue.scala
```diff
@@ -18,11 +18,11 @@ case class EditionsIssue(
     launchedEmail: Option[String],
     fronts: List[EditionsFront]
 ) {
-  def toPublishedIssue(version: Option[String] = None): PublishedIssue = PublishedIssue(
+  def toPublishedIssue(publicationEventId: Option[String] = None): PublishedIssue = PublishedIssue(
     id,
     displayName,
```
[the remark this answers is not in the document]
akash1810: It's an Option because we hard code it to `preview` for preview. However we can improve here a little.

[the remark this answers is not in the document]
akash1810: Done [here](https://github.com/guardian/facia-tool/pull/1011/commits/c988de0e7c2c7bea6d46088620d68cd5adaaa11c).

app/services/editions/publishing/EditionsBucket.scala
```diff
@@ -10,15 +10,13 @@ import PublishedIssueFormatters._
 class EditionsBucket(s3Client: AmazonS3, bucketName: String) {
   def createIssuePrefix(issue: PublishedIssue): String = s"${issue.name}/${issue.issueDate.toString}"
 
-  def createIssueFilename(issue: PublishedIssue): String = {
-    val keyname = issue.version.getOrElse("preview")
-    s"$keyname.json"
-  }
```
mkuzdowicz: you prefer the change the filename to `event-id` anyway ? we are passing that to object metadata now, what was the version before? the date ?

app/model/editions/EditionsIssue.scala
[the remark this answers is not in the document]
mkuzdowicz: 👍

app/services/editions/publishing/EditionsBucket.scala
akash1810: I think It'll keep things simple to more to a guid everywhere we can.

app/services/editions/db/IssueQueries.scala
```diff
@@ -223,7 +223,7 @@ trait IssueQueries {
       .apply()
   }
 
-  def publishIssue(issueId: String, user: User, now: OffsetDateTime) = DB localTx { implicit session =>
+  def publishIssue(issueId: String, user: User, now: OffsetDateTime): String = DB localTx { implicit session =>
     val userName = user.firstName + " " + user.lastName
     val truncatedNow = EditionsDB.truncateDateTime(now)
```
mkuzdowicz: we will need as well a function that will do the update of `PublicationEvent ` by updating the `published_on` and `status` columns and return some indication if ti was successful (base on the outcome we decide if we delete the message form SQS)

`published_on` is nullable ?

app/services/editions/publishing/EditionsBucket.scala
mkuzdowicz: 👍

app/services/editions/db/IssueQueries.scala
akash1810: > we will need as well a function that will do the update of `PublicationEvent` by event id, which will update the `published_on` and `status` columns and return some indication if it was successful (base on the outcome, we decide if we delete the message form SQS, now it is boolean)

Agreed. Let's do that in another PR though to avoid this one getting any larger.

> `published_on` is nullable ?

Yes - when we first publish an issue, we write to the events table. As we have not processed a message from SQS at this point, we do not know the `published_on` value (as that represents the time we received the SQS message).

mkuzdowicz: 👍

akash1810:
@mkuzdowicz @sihil are we happy for this to go now? We're back to a field called `version` and it being a date string.

akash1810:
Closing in favour of https://github.com/guardian/facia-tool/pull/1039.

PR #1016, opened 2019-09-18
Update incoming CAPI articles more efficiently

jonathonherbert:
## What's changed?

We ask for lots of CAPI articles in the Fronts tool. Opening collections, polling for collections, fetching the feed -- etc.

This PR changes the tool's behaviour: only update articles in the application state if the incoming article has a `lastModified` date that's more recent than the one in the store.

This should make polling in particular much more efficient, but will also be handy wherever else we're fetching articles.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/types/Capi.ts
+++ b/client-v2/src/types/Capi.ts
@@ -93,6 +93,7 @@ interface CapiArticleFields {
   isLive?: CapiBool;
   firstPublicationDate?: string;
   scheduledPublicationDate?: CapiDate;
+  lastModified?: string;
   secureThumbnail?: string;
   thumbnail?: string | void;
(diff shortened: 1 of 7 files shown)


client-v2/src/types/Capi.ts
```diff
@@ -93,6 +93,7 @@ interface CapiArticleFields {
   isLive?: CapiBool;
   firstPublicationDate?: string;
   scheduledPublicationDate?: CapiDate;
+  lastModified?: string;
```
akash1810: Why not use the `Date` type here?

jonathonherbert: 👍

PR #1017, opened 2019-09-19
Fix: external links on snap links should be allowed query params

ajwl:
## What's changed?
Celine couldn't add external links that had query params eg: 
https://www.salesforce.org/events/higher-ed-summit-horizons-2019/?utm_source=the-guardian&utm_medium=banner

This only checks out query params if they are guardian links otherwise lets the url through as normal 

Have added a test to check that the query params are preserved on external urls. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/actions/ArticleFragments.ts
+++ b/client-v2/src/shared/actions/ArticleFragments.ts
@@ -15,7 +15,7 @@ import {
 import { createFragment } from 'shared/util/articleFragment';
 import { createSnap, createLatestSnap } from 'shared/util/snap';
 import { getIdFromURL } from 'util/CAPIUtils';
-import { isValidURL } from 'shared/util/url';
+import { isValidURL, isGuardianUrl } from 'shared/util/url';
 import { MappableDropType } from 'util/collectionUtils';
 import { ExternalArticle } from 'shared/types/ExternalArticle';
 import { CapiArticle } from 'types/Capi';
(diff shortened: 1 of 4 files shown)


client-v2/src/shared/actions/ArticleFragments.ts
```diff
@@ -56,8 +56,8 @@ function updateArticleFragmentMeta(
 function articleFragmentsReceived(
   articleFragments:
     | {
-        [uuid: string]: ArticleFragment;
-      }
+      [uuid: string]: ArticleFragment;
```
ajwl: Sorry about these whitespace changes. VSCode getting aggressive

client-v2/src/shared/actions/ArticleFragments.ts
```diff
@@ -176,16 +176,16 @@ const getArticleEntitiesFromDrop = async (
   const resourceIdOrUrl = drop.data;
   const isURL = isValidURL(resourceIdOrUrl);
   const id = isURL ? getIdFromURL(resourceIdOrUrl) : resourceIdOrUrl;
-  const meta = getArticleFragmentMetaFromUrlParams(resourceIdOrUrl);
-  const isPlainUrl = isURL && !id && !meta;
+  const guMeta = isGuardianUrl(resourceIdOrUrl) ? getArticleFragmentMetaFromUrlParams(resourceIdOrUrl) : false;
```
ajwl: Because this now only applies to guardian links, i have renamed this

ajwl:
@jennygrahamjones - i added a test 🥇 
@jonathonherbert - tests now pass - is there any external link behaviour that i'm not aware of that this may break?

PR #1020, opened 2019-09-20
Reduce rerenders in article feed

jonathonherbert:
## What's changed?

This PR reduces the amount of work we do rerendering the article feed when we poll for updates:

- articles are only added to the state when their modified date has changed, so they don't rerender unnecessarily
- the list only rerenders when its order changes (value comparison)
- the feed controls only rerender when pagination data changes (value comparison)

Scripting time comparison, in ms, across average of three measurements:

| Before | After |
|--------|-------|
| 678ms  | 7ms   |

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/bundles/capiFeedBundle.ts
+++ b/client-v2/src/bundles/capiFeedBundle.ts
@@ -58,7 +58,7 @@ const fetchResourceOrResults = async (
   return {
     results: checkIsContent(response) ? [response.content] : response.results,
     pagination: checkIsContent(response)
-      ? null
+      ? undefined
       : {
           totalPages: response.pages,
           currentPage: response.currentPage,
--- a/client-v2/src/bundles/capiFeedBundle.ts
+++ b/client-v2/src/bundles/capiFeedBundle.ts
@@ -58,7 +58,7 @@ const fetchResourceOrResults = async (
   return {
     results: checkIsContent(response) ? [response.content] : response.results,
     pagination: checkIsContent(response)
-      ? undefined
+      ? null
       : {
           totalPages: response.pages,
           currentPage: response.currentPage,
@@ -96,7 +96,7 @@ export const createFetch = (
     );
     dispatch(
       actions.fetchSuccess(updatedResults, {
-        pagination: resultData.pagination,
+        pagination: resultData.pagination || undefined,
         order: nonCommercialResults.map(_ => _.id)
       })
     );
--- a/client-v2/src/bundles/capiFeedBundle.ts
+++ b/client-v2/src/bundles/capiFeedBundle.ts
@@ -5,17 +5,18 @@ import { previewCapi, liveCapi } from 'services/frontsCapi';
 import { checkIsContent } from 'services/capiQuery';
 import { getPrefills } from 'services/editionsApi';
 import { Dispatch } from 'redux';
+import { State } from 'types/State';
+import { createSelectIsArticleStale } from 'shared/util/externalArticle';
 
-type FeedState = CapiArticle[];
+type FeedState = CapiArticle;
 
 const {
   actions: liveActions,
   reducer: capiLiveFeed,
   selectors: liveSelectors
 } = createAsyncResourceBundle<FeedState>('capiLiveFeed', {
-  indexById: false,
   selectLocalState: state => state.feed.capiLiveFeed,
-  initialData: []
+  indexById: true
 });
 
 const isNonCommercialArticle = (article: CapiArticle | undefined): boolean => {
@@ -39,9 +40,8 @@ const {
   reducer: capiPreviewFeed,
   selectors: previewSelectors
 } = createAsyncResourceBundle<FeedState>('capiPreviewFeed', {
-  indexById: false,
   selectLocalState: state => state.feed.capiPreviewFeed,
-  initialData: []
+  indexById: true
 });
 
 const fetchResourceOrResults = async (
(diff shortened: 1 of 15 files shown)


client-v2/src/fixtures/initialState.ts
```diff
@@ -296,10 +296,10 @@ const shared = {
         type: 'fixed/small/slow-IV'
       }
     },
-    pagination: null,
-    lastError: null,
-    error: null,
-    lastFetch: null,
+    pagination: undefined,
```
jonathonherbert: We don't use null types much elsewhere in the project, so I updated them here for consistency's sake -- sorry for the noise!

client-v2/src/bundles/capiFeedBundle.ts
```diff
@@ -5,17 +5,18 @@ import { previewCapi, liveCapi } from 'services/frontsCapi';
...
 
 const {
   actions: liveActions,
   reducer: capiLiveFeed,
   selectors: liveSelectors
 } = createAsyncResourceBundle<FeedState>('capiLiveFeed', {
-  indexById: false,
   selectLocalState: state => state.feed.capiLiveFeed,
-  initialData: []
+  indexById: true
```
jonathonherbert: Instead of storing our feed articles in an array, we store them in an object and keep the incoming order elsewhere. This lets us update articles individually when we poll for changes, rather than having to update all of them at once.

client-v2/src/bundles/capiFeedBundle.ts
```diff
@@ -58,7 +58,7 @@ const fetchResourceOrResults = async (
   return {
     results: checkIsContent(response) ? [response.content] : response.results,
     pagination: checkIsContent(response)
-      ? null
+      ? undefined
```
akash1810: Best. Diff. Ever.

client-v2/src/components/FrontsCAPIInterface/Feed.tsx
```diff
@@ -1,35 +1,81 @@
...
+import { State } from 'types/State';
+import { connect } from 'react-redux';
 
 interface ErrorDisplayProps {
-  error: string | null;
+  error?: string;
   children: React.ReactNode;
 }
 
+interface FeedContainerProps {
```
akash1810: This type is very broad. Will we ever have both `liveArticleIds` and `previewArticleIds` and `prefillArticleIds`?

Would something like this be stricter:

```typescript
type Mode = 'prefill' | 'preview' | 'live';

interface FeedContainerProps {
  mode: Mode;
  articleIds: string[];
  error: string | undefined;
}
```

client-v2/src/components/util/RefreshPeriodically.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { useEffect, useState } from 'react';
+
+const refreshPeriodically = ({
+  timeMs,
```
akash1810: minor: would `rate` is more intention revealing here?

client-v2/src/shared/util/externalArticle.ts
```diff
@@ -48,3 +50,22 @@ export function hasMainMediaVideoAtom(
...
   );
 }
+
+/**
+ * Create a selector to answer the question -- is the article last modified field older than the given date?
+ * This function is liberal in what it accepts -- if either date is invalid/missing, it returns `true`.
+ */
+export const createSelectIsArticleStale = <State>(
+  selectArticleById: (state: State, id: string) => ExternalArticle | undefined
+) => (state: State, id: string, dateStr: string | undefined): boolean => {
```
akash1810: Could we type `dateStr` as a `Date`?

client-v2/src/components/FrontsCAPIInterface/Feed.tsx
jonathonherbert: Yes, this is a hangover from code previously in `FeedContainer`. I moved it without touching but it's very obviously due a refactor! I'll take a look, I suspect at the time I was keen to avoid expanding the scope of the PR.

client-v2/src/shared/util/externalArticle.ts
jonathonherbert: We can, but that would be mean every piece of consuming code having to roll their own date creation, along with the invalid case. I'm not sure that's preferable -- what do you think? An alternative might be to detail the format of the date string in a comment -- it's the same format we expect from CAPI.

client-v2/src/bundles/capiFeedBundle.ts
jonathonherbert: Couldn't be clearer 🤣

client-v2/src/components/FrontsCAPIInterface/Feed.tsx
akash1810: Ah hadn't acknowledged it was a just a move. Let's keep as is then to keep a small diff.

client-v2/src/components/util/RefreshPeriodically.tsx
jonathonherbert: 👍

ajwl:
+1 for splitting this out. Sounds useful though finding it hard to follow

jonathonherbert:
@ajwl that's fair, it touches a few different parts of the system in non-obvious ways! I've removed the null types but I'll split this into separate PRs for the change to the bundle creator state and the rest of the code, which should make it simpler.

client-v2/src/components/FrontsCAPIInterface/FeedItem.tsx
```diff
@@ -173,22 +187,29 @@ class FeedItem extends React.Component<FeedItemProps> {
...
-            )}
-            {article.webPublicationDate && (
-              <FirstPublished>
-                {distanceInWordsStrict(
-                  Date.now(),
-                  new Date(article.webPublicationDate)
-                )}
-              </FirstPublished>
-            )}
+            <RefreshPeriodically rateMs={collectionArticlesPollInterval}>
```
jonathonherbert: Because we can't rely on this feed updating itself as before, the component keeps these times fresh by itself.

PR #1021, opened 2019-09-23
Reduce length of Ophan calls to under 2048 chars

jennygrahamjones:
## What's changed?
When we make a call to Ophan for page view data, the length of the request might - for fronts with large collections - exceed the maximum number of characters (2048) that the request URI can contain. To avoid receiving a 414 from Ophan, we can split the request into multiple calls and combine the resulting responses together. 

## Implementation notes
This PR merges the `buildRequestUrl` and `getPageViewDataFromOphan` functions into the existing `fetchPageViewData` function: a single function that prepares, sends, and evaluates the Ophan query. As a result, the test on `buildRequestUrl` is no longer relevant.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/redux/modules/pageViewData/__tests__/actions.spec.ts
+++ b/client-v2/src/redux/modules/pageViewData/__tests__/actions.spec.ts
@@ -1,30 +1,105 @@
-import { buildRequestUrl } from '..';
-import { DerivedArticle } from 'shared/types/Article';
+import fetchMock from 'fetch-mock';
+import { fetchPageViewData } from '../actions';
+import { PageViewDataFromOphan } from 'shared/types/PageViewData';
 
-describe('buildRequestUrl', () => {
-  const correctOphanCall =
-    '?referring-path=/au/business&path=/film/2015/aug/13/dads-army-film-first-trailer-bill-nighy-toby-jones-catherine-zeta-jones&path=/commentisfree/2015/aug/13/intern-tent-david-hyde-un-internship-geneva&path=/uk-news/2018/mar/14/sharp-rise-in-number-of-eu-nationals-applying-for-uk-citizenship&hours=1&interval=10';
+describe('fetchPageViewData', () => {
+  it('should make a request to ophan to get ophan data', async () => {
+    const ophanQuery =
+      '?referring-path=/au/business&path=/film/2015/aug/13/dads-army-film-first-trailer-bill-nighy-toby-jones-catherine-zeta-jones' +
+      '&path=/commentisfree/2015/aug/13/intern-tent-david-hyde-un-internship-geneva' +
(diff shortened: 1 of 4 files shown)


client-v2/src/redux/modules/pageViewData/actions.ts
```diff
@@ -112,28 +113,44 @@ const pageViewDataRequestedAction = (
...
-};
+  const maxLength = 2048 - timePeriod.length - base.length;
+
+  const articlePaths = articlesInFront.map(article => `path=/${article}&`);
+
+  const urls: string[] = articlePaths.reduce(
+    (acc, path) => {
+      if (acc.length) {
+        const latestUrl = acc[acc.length - 1];
+        if (latestUrl.length + path.length < maxLength) {
```
jonathonherbert: Minor: could we move `latestUrl` out of the `if` statement, and then do both comparisons in one go? If that works, it saves us a nested `if`, which is _slightly_ easier to wrap my head around 😁

client-v2/src/redux/modules/pageViewData/__tests__/actions.spec.ts
```diff
@@ -1,30 +1,105 @@
...
+      'uk-news/2018/mar/14/sharp-rise-in-number-of-eu-nationals-applying-for-uk-citizenship'
+    ];
+
+    fetchMock.once(`/ophan/histogram${ophanQuery}`, []);
+
+    const result = await fetchPageViewData(frontIds, articleIds);
+    expect(result).toEqual([]);
+  });
+  it('should make multiple requests to ophan when given a number of articles that would exceed ophans request length', async () => {
+    const ophanQuery =
```
jonathonherbert: Minor: consider moving this to a fixture file 😁

client-v2/src/redux/modules/pageViewData/actions.ts
jennygrahamjones: Yes, hopefully it makes more sense now! 🤞

client-v2/src/redux/modules/pageViewData/actions.ts
```diff
@@ -112,28 +113,44 @@ const pageViewDataRequestedAction = (
...
+  );
 
-export {
-  fetchPageViewData,
-  buildRequestUrl,
-  getPageViewData,
-  pageViewDataReceivedAction
+  const response = await Promise.all(ophanCalls);
+  const parsedResponse = await Promise.all(
+    response.map(r => r.json().then((data: PageViewDataFromOphan) => data))
```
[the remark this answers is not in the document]
jennygrahamjones: You're right; it's not necessary at all in fact!

PR #1023, opened 2019-09-23
LOADING STATE: add loading state for launch button

ajwl:
## What's changed?
When users click the "Launch" button in the corner of a collection we want to show a loading state so that they know something is happening. 

ezgif com-video-to-gif (2)

We are reusing the loading gif from Composer as suggested by @akemitakagi 
Button also gets disabled and should stay the same width 

## Implementation notes
**Allowing the use of Gifs**
* The loading image is a gif. This is the first gif we're using in the new tool.
* To parse the gif we need to add .gif to the webpack config
* To allow typescript to recognise importing the gif, needed to edit the `modules.d.ts` so it accepts it. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
```diff
@@ -192,15 +194,20 @@ class Collection extends React.Component<CollectionProps> {
                 <Button
                   size="l"
                   priority="primary"
-                  onClick={() => publish(id, frontId)}
+                  onClick={() => {
```
jonathonherbert: The logic looks good here! But there are two things I should mention:

- this will cause a warning (and a memory leak, although it will be short lived!) if the component is unmounted before the operation completes -- see https://www.robinwieruch.de/react-warning-cant-call-setstate-on-an-unmounted-component for an example solution.
- Adding this as a class method rather than inline with the JSX might make the component easier to read.

client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
```diff
@@ -192,15 +194,20 @@ class Collection extends React.Component<CollectionProps> {
                 <Button
```
jonathonherbert: Could we have the same behaviour for Discard?

client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
ajwl: 👍

client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
ajwl: You mean so that the spinner runs if you are discarding a set of changes? 
Can do - do you think the delay there is long enough to merit the loading spinner?

client-v2/src/actions/Collections.ts
```diff
@@ -464,37 +464,47 @@ function publishCollection(
         );
 
         dispatch(getCollections([collectionId]));
+        dispatch(pollForCollectionPublished(collectionId, frontId));
```
ajwl: Extracting this side effect code and pulling into a new function to clarify a) what it does and b) that the main function doesn't have to wait for the side effect to complete before it returns

client-v2/src/components/FrontsEdit/CollectionComponents/Collection.tsx
ajwl: do you still want this..? Can add

jennygrahamjones: Would we ever have a delay when discarding?

jonathonherbert: This would be awesome to have. Discard involves a network call, so this button is latent for network + server time 👍

jonathonherbert: (If it's too much it can always be a separate PR)

PR #1025, opened 2019-09-24
Add training edition

aug24:
## What's changed?
Training editions.

## Implementation notes
None

### General
- [ ] 🤖 Relevant tests added
- [X ] ✅ CI checks / tests run locally
- [X] 🔍 Checked on CODE

### Client
- [ X] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ X] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)


jennygrahamjones:
Is this equivalent to the `/training` priority on Fronts? Are the American and Australian editions also for training?

aug24:
> Is this equivalent to the `/training` priority on Fronts? Are the American and Australian editions also for training?

The American and Australian editions are there ready for launch later, so that we can start getting ready for them.  The training one will never be launched, and is intended for 'live' user training.

akash1810:
Could we decouple this PR? That is, introduce the US/AUS editions separately?

aug24:
> Could we decouple this PR? That is, introduce the US/AUS editions separately?

We could, certainly.  Originally David asked me to tag them on, but now the apparent urgency has gone I'll split them back out.

client-v2/src/constants/priorities.ts
```diff
@@ -1,8 +1,27 @@
...
+    address: 'daily-edition'
+  },
+  americanEdition: {
+    description: 'American Edition',
+    address: 'american-edition'
+  },
+  australianEdition: {
+    description: 'Australian Edition',
+    address: 'australian-edition'
+  },
```
akash1810: Should this be here in this PR?

PR #1028, opened 2019-09-26
Allow updates to ophan on a per-article basis

jonathonherbert:
## What's changed?

This PR refactors the internal state of the page view module to make it easier to update page view data on a per-article basis. This is important when we're trying to make the tool more performant -- when we first drag an article into a collection and make a request to ophan for its analytics, we don't want to have to update an entire collection's worth of data in the state.

## Implementation notes

I've done this by indexing front, collection and article data by id, which has the additional benefit of reducing the amount of work our selectors have to do to fetch page view data. This does, however, mean threading a collection id all the way through from `CollectionItem` to `ArticleBody`, which is less elegant than I'd like. If there's another solution I'd be really keen to hear 👍 

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/redux/modules/pageViewData/actions.ts
+++ b/client-v2/src/redux/modules/pageViewData/actions.ts
@@ -47,7 +47,12 @@ const getPageViewData = (
       const data = await fetchPageViewData(frontId, urlPaths);
       const dataWithArticleIds = convertToStoriesData(data, articles);
       dispatch(
-        pageViewDataReceivedAction(dataWithArticleIds, frontId, collectionId)
+        pageViewDataReceivedAction(
+          dataWithArticleIds,
+          frontId,
+          collectionId,
+          true
+        )
       );
     } catch (e) {
       throw new Error(`API request to Ophan for page view data failed: ${e}`);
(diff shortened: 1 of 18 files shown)


client-v2/src/redux/modules/pageViewData/selectors.ts
```diff
@@ -26,37 +13,18 @@ const selectOpenCollectionsForFront = (
...
-      );
-    }
-  );
+const selectDataForArticle = (
+  state: State,
+  articleId: string,
+  collectionId: string,
+  frontId: string
+): PageViewStory | undefined =>
+  oc(state).shared.pageViewData[frontId].collections[collectionId].stories[
```
jennygrahamjones: What does `oc` do here?

client-v2/src/redux/modules/pageViewData/actions.ts
```diff
@@ -90,14 +95,18 @@ const getArticleIdFromOphanData = (
 const pageViewDataReceivedAction = (
   data: PageViewStory[],
   frontId: string,
-  collectionId: string
+  collectionId: string,
+  // Clear out the previous data for this collection. Useful when polling
+  // to prevent an endless accumulation of page view data.
+  clearPreviousData = false
```
jennygrahamjones: This is so sensible! I can't believe we didn't think of this to begin with 🤦‍♀

PR #1030, opened 2019-09-27
Kicker picker and Metadata creator

aug24:
## What's changed?
Unification of the Capi Prefil from multiple locations into a single logic class.
Provision of a kicker picker to suggest a kicker from the various tags and title information

## Implementation notes
This is all the server side code - there will need to be a matching ticket to expose the picked kicker to the user in the UI.

### General
- [X] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


app/logic/CapiPrefiller.scala
```diff
@@ -0,0 +1,96 @@
...
+      case (false, _) => (None, None)
+      // finally if a cutout is desired but nothing appropriate was found then explicitly set to use article trail
+      case (true, None) => (Some(MediaType.UseArticleTrail), None)
+    }
+    val newMetadata = metadata.copy(imageCutoutReplace = mediaType.contains(MediaType.Cutout))
+
+    val pickedKicker = pickKicker(content)
+
+    Prefill(
+      internalPageCode.toInt,
```
jennygrahamjones: Is it possible that the get on line 27 could return something that wouldn't parse as an Int?

aug24: Good spot, but no, because after this refactoring, internal page code is ... an int.

I'll remove the toInt :)

app/util/ContentUpgrade.scala
```diff
@@ -45,45 +44,34 @@ object ContentUpgrade {
...
-      val jsonString = JsonMethods.compact(JsonMethods.render(json))
-      val maybeParsedJson: Option[Json] = parser.parse(jsonString).right.toOption
-      val maybeCapiContent: Option[Content] = maybeParsedJson.flatMap(json => json.as[Content].right.toOption)
-      (json, maybeCapiContent) match {
-        case (jsObject: JObject, Some(content)) =>
-          val cardStyle = CardStyle(content, TrailMetaData.empty)
-          val initialMetaDataMap: Map[String, Boolean] = ResolvedMetaData.toMap(ResolvedMetaData.fromContent(content, cardStyle))
+    Try(getUpgradedItem(json)) match {
+      case Success(capiItem) => capiItem
+      case Failure(_) => json
```
sihil: We should probably log a warning here.

app/util/ContentUpgrade.scala
```diff
@@ -45,45 +44,34 @@ object ContentUpgrade {
...
+              ~ ("pickedKicker" -> prefill.pickedKicker)
+              ~ ("mediaType" -> prefill.mediaType.map(_.toString))
+            )
           )
-        case _ => json
       }
-    }) match {
-      case Success(capiItem) => capiItem
-      case Failure(_) => json
+      case _ => json
```
sihil: Again, should we log a warning here? This is an unexpected fallback IIUC?

app/logic/CapiPrefiller.scala
```diff
@@ -0,0 +1,96 @@
...
+      internalPageCode,
+      newMetadata,
+      cutoutImage,
+      cardStyle.toneString,
+      mediaType,
+      pickedKicker)
+  }
+
+  def getFirstContributorWithCutoutOption(content: Content, maybeImageCutoutReplace: Option[Boolean]): (Option[Boolean], Option[String]) =
+    maybeImageCutoutReplace match {
```
sihil: I would really prefer this function to have a single purpose - to get a cutout but not mix in any idea of the cutout replace.

The left hand side of the returned tuple is thrown away anyway.... and the logic to collapse it could be pulled into the calling function which has more context as to why we'd not look it up if a flag is set.

aug24: Done

app/util/ContentUpgrade.scala
aug24: Done

app/util/ContentUpgrade.scala
aug24: Done

PR #1033, opened 2019-09-27
Clean up Fronts component heirarchy

jonathonherbert:
## What's changed?

Part tech debt restructuring and part optimisation, this PR splits the component responsible for rendering the front overview and the front itself into two components, renaming and adding comments to indicate the responsibility of each component.

## Implementation notes

Because changes in highlighted collection ids no longer trigger a rerender of the front, scrolling through the front causes less work for the app -- see below.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/bundles/frontsUIBundle.ts
```diff
@@ -88,6 +91,27 @@ const editorCloseCollections = (
   payload: { collectionIds }
 });
 
+const editorOpenAllCollectionsForFront = (
```
jonathonherbert: Putting this logic in our bundle keeps business logic out of the component.

client-v2/src/components/util/WithDimensions.tsx
```diff
@@ -24,28 +23,26 @@ const initialState = { width: undefined, height: undefined } as Dimensions;
...
+    observer.observe(ref.current);
 
-      return () => observer.disconnect();
-    },
-    [element]
-  );
-  return children(dimensions);
+    return () => observer.disconnect();
+  }, []);
+  return <div ref={ref}>{children(dimensions)}</div>;
```
jonathonherbert: I've reluctantly added a container div here, as without a ref that this component owns, `useLayout` can't guarantee that refs of other components are present in the dom. This was effectively working by coincidence before this point, due to redundant rerenders.

client-v2/src/components/FrontsEdit/Edit.tsx
```diff
@@ -14,7 +14,7 @@ import {
 } from 'bundles/frontsUIBundle';
 import { State } from 'types/State';
 import { ActionError } from 'types/Action';
-import FrontContainer from './FrontContainer';
+import FrontContainer from './FrontSection';
```
jonathonherbert: I've split up the files responsible for rendering a front into

`FrontSection` --> `FrontContainer` --> `FrontContent` & `Overview`
(preview bar)     (expand btns, overview)   (collections etc.)

Hopefully this is clear, super happy to hear anyone's thoughts 👍

client-v2/src/components/FrontsEdit/FrontContent.tsx
```diff
@@ -0,0 +1,220 @@
+import React from 'react';
```
jonathonherbert: These files are large, but most of these lines haven't been changed -- they're the result of `FrontContainer.tsx` being split out into `FrontContent.tsx` and `FrontContainer.tsx`.

client-v2/src/components/util/WithDimensions.tsx
```diff
@@ -24,28 +23,26 @@ const initialState = { width: undefined, height: undefined } as Dimensions;
...
  * ```
  */
-const WithDimensions = ({ element, children }: Props) => {
+const WithDimensions = ({ children }: Props) => {
   const [dimensions, setDimensions] = useState(initialState);
-  useEffect(
-    () => {
-      if (!element) {
+  const ref: React.RefObject<HTMLDivElement> = React.createRef();
+  useLayoutEffect(() => {
```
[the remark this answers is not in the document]
jonathonherbert: Discussed IRL, but `useLayoutEffect` fires _after_ the component has been mounted and reconciled with the DOM, ensuring that the `ref` we've added in the returned JSX has an element to use. More info @ e.g. https://kentcdodds.com/blog/useeffect-vs-uselayouteffect/

PR #1034, opened 2019-09-27
Shrink vertical height of inline form

jennygrahamjones:
## What's changed?

This PR adds a new layout to the Fronts inline form. In an attempt to make more efficient use of space in the browser, when the form has a width of more than 700px the image toggles and inputs will display to the right of the other inputs and toggles. 

An example of the new layout:
Screenshot 2019-09-27 at 17 06 33

([Trello card](https://trello.com/c/RJFjyDvc/836-shrink-vertical-height-of-inline-form-akemi-designs))

## Implementation notes

This PR also makes a small change to the integration tests, removing a call to `maximizeWindow` from a number of tests. Previously, this call was necessary to allow the headless browser to find certain elements in the DOM. It's not clear when this stopped being necessary, but without removing this call, the tests now _would_ fail. This is because at the point the window is maximised, our collection items are re-rendered, due to the fact that they are moving from 'default' size to 'wide' size.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
+++ b/client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
@@ -89,32 +89,41 @@ const FormContainer = styled(ContentContainer.withComponent('form'))`
 `;
 
 const FormContent = styled.div`
-  flex: 1;
+  flex: 3;
   display: flex;
   flex-direction: ${(props: { size?: string }) =>
-    props.size === 'medium' ? 'column' : 'row'};
+    props.size !== 'wide' ? 'column' : 'row'};
 `;
 
 const RowContainer = styled.div`
   overflow: hidden;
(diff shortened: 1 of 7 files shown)


client-v2/src/components/FrontsEdit/ArticleFragmentFormInline.tsx
```diff
@@ -88,13 +90,44 @@ const FormContainer = styled(ContentContainer.withComponent('form'))`
 `;
 
 const FormContent = styled.div`
-  flex: 1;
+  flex: 3;
+  display: flex;
+  flex-direction: ${(props: { size?: string }) =>
+    props.size !== 'wide' ? 'column' : 'row'};
```
jonathonherbert: 👍

PR #1038, opened 2019-10-02
make version property of PublishedIssue non-optional

akash1810:
## What's changed?

Update `version` on `PublishedIssue` to a `String` rather than `Option[String]` and when we're talking about preview, `version` is set to `"preview"` via a helper method, which keeps things DRY. This makes for a cleaner interface.

## Implementation notes

This is part of my effort to split https://github.com/guardian/facia-tool/pull/1011 into smaller pieces of work.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/EditionsController.scala
+++ b/app/controllers/EditionsController.scala
@@ -110,7 +110,7 @@ class EditionsController(db: EditionsDB,
 
   def getPreviewEdition(id: String) = EditEditionsAuthAction { _ =>
     db.getIssue(id).map { issue =>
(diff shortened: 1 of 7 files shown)


akash1810:
> Can't see any reason it was ever optional anyway...?

The only thing I can see didn't send `version` in the json blob downstream. We are now - https://github.com/guardian/facia-tool/pull/1038/files#diff-2639180e96b5fa3abaea1c80c9833136R28-R36

PR #1039, opened 2019-10-02
new table to track publication status of an issue version

akash1810:
## What's changed?

When publishing, insert a row into `issue_versions` and `issue_versions_events`. This will allow us to record events fired by the Archiver lambda, providing the ability to see the progress of publication.

We can then render this on the client to provide confidence that the publishing action is doing something.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/services/editions/db/IssueQueries.scala
+++ b/app/services/editions/db/IssueQueries.scala
@@ -226,7 +226,7 @@ trait IssueQueries {
       .apply()
   }
 
-  def publishIssue(issueId: String, user: User, now: OffsetDateTime): String = DB localTx { implicit session =>
+  def createIssueVersion(issueId: String, user: User, now: OffsetDateTime): EditionIssueVersionId = DB localTx { implicit session =>
     val userName = user.firstName + " " + user.lastName
     val truncatedNow = EditionsDB.truncateDateTime(now)
 
--- a/app/services/editions/db/IssueQueries.scala
+++ b/app/services/editions/db/IssueQueries.scala
@@ -1,5 +1,6 @@
 package services.editions.db
 
+import java.time.format.DateTimeFormatter
 import java.time.{LocalDate, OffsetDateTime}
 
 import com.gu.pandomainauth.model.User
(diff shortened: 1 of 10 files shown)


app/services/editions/db/IssueQueries.scala
```diff
@@ -223,17 +226,50 @@ trait IssueQueries {
       .apply()
   }
 
-  def publishIssue(issueId: String, user: User, now: OffsetDateTime) = DB localTx { implicit session =>
+  def publishIssue(issueId: String, user: User, now: OffsetDateTime): String = DB localTx { implicit session =>
     val userName = user.firstName + " " + user.lastName
     val truncatedNow = EditionsDB.truncateDateTime(now)
```
mkuzdowicz: - in `def publishIssue(issueId: String, user: User, now: OffsetDateTime): String` can we type alias String to `VersionId` ?

like:

`def publishIssue(issueId: String, user: User, now: OffsetDateTime): VersionId` 

it is not clear that `publishIssue` will generate and return the id

- Or generate versionId outside and pass to `publishIssue`
- i am not sure if `publishIssue` is correct function name it `createIssueVersion` or `initialiseIssueVersion` and publishing is done by downstream stuff and function that calls this one `EditionsPublishing.publish`

app/services/editions/publishing/events/package.scala
```diff
@@ -1,10 +1,11 @@
 package services.editions.publishing
 
+import model.editions.IssueVersionStatus
 import play.api.libs.json.{Format, Json}
 
 package object events {
 
-  case class PublishEvent(status: String, message: String)
+  case class PublishEvent(status: IssueVersionStatus, message: String)
```
mkuzdowicz: PublishEvent will need `version` and `timestamp` now as well, but that can be done in next PR

current example JSON:

`\"edition\":\"daily-edition\",\"version\":\"2019-10-02T13:44:18.393Z\",\"issueDate\":\"2019-09-23\",\"status\":\"Processing\",\"message\":\"Publication stage: started\",\"timestamp\":\"2019-10-02T13:44:22+00:00\"`

fields like `edition` and `issueDate` are nice to have i guess

akash1810: Yes, let's push this to the next PR.

PR #1041, opened 2019-10-03
Add three containers to every special front

blishen:
To allow production to organise content within special fronts, we are adding three generic containers to each special front that was previously a single container.

## What's changed?
Sadly the single line descriptions for special sections now more resemble other fronts, since they now specify containers.

## Implementation notes
Written in github, treat with caution.

### DIFF
--- a/app/model/editions/templates/TemplateHelpers.scala
+++ b/app/model/editions/templates/TemplateHelpers.scala
@@ -1,6 +1,5 @@
 package model.editions.templates
 
-import model.editions.Swatch.Neutral
 import model.editions.{CapiPrefillQuery, CollectionPresentation, CollectionTemplate, FrontPresentation, FrontTemplate, Swatch}
 
 object TemplateHelpers {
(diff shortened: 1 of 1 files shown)


sihil:
The more this kind of thing happens the more I'm inclined to allow for arbitrary fronts and collections to be created on the fly. This is really cluttering the UI for the happy path.

blishen:
> The more this kind of thing happens the more I'm inclined to allow for arbitrary fronts and collections to be created on the fly. This is really cluttering the UI for the happy path.

Nooooo...

This is it. I promise. The problem was HUGE specials were hard to break up.

blishen:
> > The more this kind of thing happens the more I'm inclined to allow for arbitrary fronts and collections to be created on the fly. This is really cluttering the UI for the happy path.
> 
> Nooooo...
> 
> This is it. I promise. The problem was HUGE specials were hard to break up.

On reflection though, able to make containers on the fly within special fronts only would be a good thing. But let's not worry today.

PR #1045, opened 2019-10-04
Display page view data for snaplinks

jennygrahamjones:
## What's changed?

We can now show page view data for snaplinks. This is only applicable for non-article snaplinks, e.g. https://www.theguardian.com/news/series/todayinfocus.

Screenshot 2019-10-04 at 12 49 42

([Trello](https://trello.com/c/peQqzxZC/833-we-send-snaplinks-and-duplicate-articles-to-ophan-we-should-not-do-that))

## Implementation notes

As snaplinks don't have the `urlPath` that is used to look up article data, this PR instead uses the `href` of the snaplink to populate the Ophan request. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/redux/modules/pageViewData/actions.ts
+++ b/client-v2/src/redux/modules/pageViewData/actions.ts
@@ -54,8 +54,11 @@ const getPageViewData = (
   const articles = articleIds
     .map(_ => selectArticleFromArticleFragment(state, _))
     .filter(_ => _) as DerivedArticle[];
+  // the url path is either the urlPath value, for articles, or the href for snaplink
   const urlPaths: string[] = articles
-    .map(article => article.urlPath)
+    .map(article =>
+      article.urlPath ? article.urlPath : article.href && article.href.substr(1)
+    )
     .filter(_ => _) as string[];
   const data = await fetchPageViewData(frontId, urlPaths);
   const dataWithArticleIds = convertToStoriesData(data, articles);
@@ -90,9 +93,11 @@ const getArticleIdFromOphanData = (
   ophanData: PageViewDataFromOphan,
   articles: DerivedArticle[]
 ): string | undefined => {
-  // the path from ophan has a slash at the front, removing below
-  const ophanPathClean = ophanData.path.substr(1);
-  const matchingArticle = articles.find(a => a.urlPath === ophanPathClean);
+  // if we have a urlPath, we need to trim the Ophan data path to make the comparison
+  // if we have an href - because the article is a snap link - we don't need to trim
+  const matchingArticle = articles.find(
+    a => a.urlPath === ophanData.path.substr(1) || a.href === ophanData.path
+  );
   return matchingArticle ? matchingArticle.uuid : undefined;
 };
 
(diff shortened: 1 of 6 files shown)


client-v2/src/shared/components/snapLink/SnapLink.tsx
```diff
@@ -152,7 +180,17 @@ const SnapLink = ({
...
+          {featureFlagPageViewData && canShowPageViewData && collectionId && (
+            <PageViewDataWrapper data-testid="page-view-graph">
+              <ArticleGraph
+                articleId={id}
+                collectionId={collectionId}
+                frontId={frontId}
+              />
+            </PageViewDataWrapper>
+          )}
+          {/* )} */}
```
ajwl: leftover comment?

client-v2/src/shared/components/snapLink/SnapLink.tsx
```diff
@@ -47,7 +49,25 @@ const SnapLinkURL = styled.p`
...
+const PageViewDataWrapper = styled.div`
+  width: 45px;
+  display: flex;
+  flex-direction: column;
+  align-items: flex-end;
+  padding-right: 5px;
+  font-size: 12px;
+  height: 35px;
+  padding-bottom: 3px;
+`;
```
ajwl: Are these defined somewhere else as well? Could we pull them out into their own files and import them into both places?

jennygrahamjones: They are indeed. Good idea.

client-v2/src/redux/modules/pageViewData/actions.ts
```diff
@@ -55,7 +56,12 @@ const getPageViewData = (
     .map(_ => selectArticleFromArticleFragment(state, _))
     .filter(_ => _) as DerivedArticle[];
   const urlPaths: string[] = articles
-    .map(article => article.urlPath)
+    .map(
+      article =>
+        article.urlPath
+          ? article.urlPath // it's an article
+          : article.href && !isValidURL(article.href) && article.href.substr(1) // it's a snaplink
```
jennygrahamjones: The `!isValidURL(...)` check here ensures that we don't try to request Ophan data for a non-Guardian link. I'm definitely open to different / better ways of making this distinction.

ajwl: What about `isGuardianUrl` or `isGuardianWebsiteUrl` from https://github.com/guardian/facia-tool/blob/master/client-v2/src/shared/util/url.ts#L42

I guess so it's not a guardian article but it is a guardian link, so we may have data for it.

ajwl: The advantage of `isGuardianWebsiteUrl` is that we won't try to fetch data for commercial pages etc

jennygrahamjones: @ajwl I considered that, but I don't think that at this stage we have access to a complete URL containing a host name. The `href` is just something like "/news/series/todayinfocus" or "/uk" which would fail both the `isValidURL` and `isGuardianWebsiteUrl` checks. 

Hopefully this example illustrates my thought process:

Screenshot 2019-10-04 at 15 51 17

The third snap is not part of the Ophan query, because although it _is_ a valid URL, it is not one that Ophan would store data on (because the domain is not just the base Guardian domain). If all commercial pages are located on something like http://some-domain.theguardian.com then we would not, in this implementation, request data on them.

Happy to document this more explicitly in the code, if you think that would help?

ajwl: 👍  makes sense. And have just seen it working in local

PR #1052, opened 2019-10-07
Rename `ArticleFragment`/`CollectionItem` to `Card`

jonathonherbert:
## What's changed?

Rename `articleFragment` and `collectionItem` (which refer to the same entity) to `card`.

The name `cards` has been shopped around the team for a while, and the consensus seems to be that it's OK --

- It relates to a visual representation, which means that we can use this name across both Digital and Editorial -- in practice it's been easy to use in conversation
- It's a single name that's abstract enough to cover all of the things we might want to add to a collection now and in the future
- One name for one thing > Two names for one thing??

Any feedback would be really helpful, as we hope this change will make the tool easier to reason about for developers and users alike, and any evidence to the contrary is important here!

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


ajwl:
A few references to "fragment":
`utils.ts` - specToFragment
`card.spec.js`
`Card.ts`
`Clipboard.ts`
`KeyboardNavigation.ts`
`frontsUIBundle.ts`

jonathonherbert:
@ajwl good point, I forgot about the `fragment` fragments 👍 should be all done now.

jennygrahamjones:
You've got a couple of collection items in card.spec.ts, formSelectors.spec.ts and OpenFormsWarning.tsx - but otherwise, very happy to have this change!

PR #1055, opened 2019-10-08
update interface of Editions SQS reader

akash1810:
## What's changed?

As we're reading one message at a time; using an `Option` rather than a `List` makes the code easier to reason about.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/services/editions/publishing/events/PublishEventsQueueFacade.scala
+++ b/app/services/editions/publishing/events/PublishEventsQueueFacade.scala
@@ -11,7 +11,7 @@ import scala.collection.JavaConverters._
 import scala.util.{Failure, Success, Try}
 
 private[events] trait PublishEventsQueueFacade {
-  def getPublishEventsFromQueue: List[PublishEventMessage]
+  def getPublishEventFromQueue: Option[PublishEventMessage]
 
   def delete(receiptHandle: String)
 }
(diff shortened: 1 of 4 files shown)


app/services/editions/publishing/events/PublishEventsQueueFacade.scala
```diff
@@ -45,22 +45,22 @@ private[events] class PublishEventsSQSFacade(val config: ApplicationConfiguratio
     }
   }
 
-  private def receiveMessages: List[Message] = {
+  private def receiveMessage: Option[Message] = {
     Try {
       val receiveRequest = new ReceiveMessageRequest()
         .withQueueUrl(queueURL)
         .withMaxNumberOfMessages(maxNumberOfSQSMessagesPerReceiveReq)
         .withWaitTimeSeconds(sqsClientLongPoolingWaitTimeSec)
```
jonathonherbert: Just for my own curiosity -- `maxNumberOfSQSMessagesPerReceiveReq` looks like it doesn't have to be 1, but we seem to only be parsing the head of this list. Is there a possibility we could drop messages here?

akash1810: Great spot and I should have made the PR description more explicit to justify my reasoning for this change.

We're reading from the queue, parsing into case class then writing to database. Writing multiple records to postgres is totally fine, however the query is slightly more complex than writing a single record.

Given we're only ever reading one message from SQS, moving to an `Option[Message]` rather than a `List[Message]` massively simplifies the code and also forces us to carefully consider the impact of increasing the value of `maxNumberOfSQSMessagesPerReceiveReq` in the future.

mkuzdowicz: yeah, unfortunately long pooling is more confusing :) 
We are getting one message per req continuously
and when SQS become empty for given client it will wait 15 seconds until it will do next call
It's kind of like scheduling with 15 sec interval but i bit more reactive
and making the interface explicitly to get 1 message make sense

PR #1057, opened 2019-10-09
Ensure custom kicker is only set for Editions

aug24:
### General
- [X] 🤖 Relevant tests added
- [X] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/actions/__tests__/editionscards.spec.ts
+++ b/client-v2/src/shared/actions/__tests__/editionscards.spec.ts
@@ -1,0 +1,51 @@
+import configureMockStore from 'redux-mock-store';
+import fetchMock from 'fetch-mock';
+import thunk from 'redux-thunk';
+import { createArticleEntitiesFromDrop, cardsReceived } from '../Cards';
+import initialState from 'fixtures/initialStateForEditions';
+import { capiArticle } from '../../fixtures/shared';
+import { actionNames as externalArticleActionNames } from 'shared/bundles/externalArticlesBundle';
+import { createCard } from 'shared/util/card';
+import { RefDrop } from 'util/collectionUtils';
+
+jest.mock('uuid/v4', () => () => 'card1');
+const middlewares = [thunk];
+const mockStore = configureMockStore(middlewares);
+const idDrop = (id: string): RefDrop => ({ type: 'REF', data: id });
+
+describe('Editions cards actions', () => {
+  const { confirm } = window;
+  const localNow = Date.now;
+  const localGetTime = Date.prototype.getTime;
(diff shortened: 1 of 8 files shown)


client-v2/src/shared/actions/__tests__/editionscards.spec.ts
```diff
@@ -0,0 +1,51 @@
+import configureMockStore from 'redux-mock-store';
```
jonathonherbert: Thanks for the great work on the tests here, the scaffolding is non-trivial and the specs look good 👍

aug24: Most of it is refactored pre-existing tests  - I split them out for clarity so that I could add *one* cleanly.

client-v2/src/shared/actions/__tests__/editionscards.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      });
+      const store = mockStore(initialState);
+      await store.dispatch(createArticleEntitiesFromDrop(
+        idDrop('internal-code/page/5029528')
+      ) as any);
+      const actions = store.getActions();
+      expect(actions[0].type).toEqual(externalArticleActionNames.fetchSuccess);
+      expect(actions[1]).toEqual(
+        cardsReceived({
+          card1: createCard('internal-code/page/5029528', true)
```
jonathonherbert: One note -- asserting on the end state is arguably less fragile than asserting on the actions dispatched. (This is not a criticism -- you will have taken the approach here from other specs in the project, and it still has lots of value, so leaving this as it is is fine!) See e.g. `capiFeedBundle.spec.ts` for an example of this alternative approach.

client-v2/src/shared/util/card.ts
```diff
@@ -5,12 +5,15 @@ import cloneDeep from 'lodash/cloneDeep';
 
 const createCard = (
   id: string,
+  isEdition: boolean,
```
jonathonherbert: I'd love to refactor this into an options object at some point -- is it appropriate to do it in this PR, or is there another PR pending that will change things here?

aug24: No time, but I'll raise a tech debt card and add a comment?

client-v2/src/shared/actions/__tests__/editionscards.spec.ts
jonathonherbert: Ah, gotcha 👍

client-v2/src/shared/util/card.ts
jonathonherbert: Sounds good.

PR #1061, opened 2019-10-11
Be able to define time window rules for prefill queries in templates

mkuzdowicz:
## What's changed?

Be able to define time window rules for prefill queries in templates

## Implementation notes

Time window config is specified on Edition Template level

for example lets say for DailyEdition it will be:

```
capiQueryPrefillParams = CapiQueryPrefillParams(
      timeWindowConfig = TimeWindowConfigInDays(
        startDiff = -1,
        endDiff = 3
      )
    )
```

then it means that for issue-date is `2019-OCT-10` for window (-1, 3) the fromDate will be `2019-OCT-9` toDate will be `2019-OCT-13`

those values are propagated to Capi query params:

`to-date=2019-10-07T00%3A00%3A00Z`
`from-date=2019-10-04T00%3A00%3A00Z`

Now all templates have `startDiff = 0` and `endDiff = 0 `so it remain the same as before in runtime

Next steps would be persisting those window settings in DB

And having a new PathType such that those window rules will apply to  specified path type and to not disturb crosswords

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### DIFF
--- a/app/model/editions/EditionsTemplates.scala
+++ b/app/model/editions/EditionsTemplates.scala
@@ -146,7 +146,7 @@ case class FrontTemplate(
   def swatch(swatch: Swatch) = copy(presentation = FrontPresentation(swatch))
 }
 
-case class TimeWindowConfigInDays(startDiff: Int, endDiff: Int)
+case class TimeWindowConfigInDays(startOffset: Int, endOffset: Int)
 case class CapiQueryPrefillParams(timeWindowConfig: TimeWindowConfigInDays)
 
 case class EditionTemplate(
--- a/app/model/editions/EditionsTemplates.scala
+++ b/app/model/editions/EditionsTemplates.scala
@@ -146,8 +146,12 @@ case class FrontTemplate(
   def swatch(swatch: Swatch) = copy(presentation = FrontPresentation(swatch))
 }
 
+case class TimeWindowConfigInDays(startDiff: Int, endDiff: Int)
+case class CapiQueryPrefillParams(timeWindowConfig: TimeWindowConfigInDays)
+
 case class EditionTemplate(
   fronts: List[(FrontTemplate, Periodicity)],
+  capiQueryPrefillParams: CapiQueryPrefillParams,
   zoneId: ZoneId,
   availability: Periodicity,
 )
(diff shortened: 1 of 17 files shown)


app/model/editions/EditionsTemplates.scala
```diff
@@ -146,8 +146,12 @@ case class FrontTemplate(
   def swatch(swatch: Swatch) = copy(presentation = FrontPresentation(swatch))
 }
 
+case class TimeWindowConfigInDays(startDiff: Int, endDiff: Int)
+case class CapiQueryPrefillParams(timeWindowConfig: TimeWindowConfigInDays)
```
akash1810: Could use a default here to have a simpler interface:

```suggestion
case class CapiQueryPrefillParams(timeWindowConfig: TimeWindowConfigInDays = TimeWindowConfigInDays(0, 0))
```

app/services/editions/prefills/PrefillHelper.scala
```diff
@@ -0,0 +1,67 @@
+package services.editions.prefills
```
akash1810: What's the reason for this change? That is, why are we extracting `geneneratePrefillQuery` out from `Capi.scala`?

mkuzdowicz: I wanted to have a package related to prefills solely

app/model/editions/EditionsTemplates.scala
mkuzdowicz: I had that at the beginning, but then i decided that it is more clear to specify those values in templets config so it will be clear which values are used by which edition
Default values can cause that we may forget that those configs exists

app/model/editions/EditionsTemplates.scala
```diff
@@ -146,8 +146,12 @@ case class FrontTemplate(
   def swatch(swatch: Swatch) = copy(presentation = FrontPresentation(swatch))
 }
 
+case class TimeWindowConfigInDays(startDiff: Int, endDiff: Int)
```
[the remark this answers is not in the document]
mkuzdowicz: `startOffset` and `endOffset` ?

test/editions/EditionTemplates.scala
```diff
@@ -15,58 +13,61 @@ class editionTemplateTest extends FreeSpec with Matchers {
...
-    override def getPrefillArticleItems(issueDate: LocalDate, capiPrefillQuery: CapiPrefillQuery): Future[List[Prefill]] = Future.successful(Nil)
+
+    override def getPrefillArticles(prefillParams: PrefillParamsAdapter, currentPageCodes: List[String]): Future[SearchResponse] = Future.successful(null)
+
+    override def getPrefillArticleItems(prefillParams: PrefillParamsAdapter): Future[List[Prefill]] = Future.successful(Nil)
   }
+
   val templating = new EditionsTemplating(TestEdition.templates, TestCapi)
 
-//  "createEdition" - {
```
akash1810: Not for this PR, but this is of no value commented out. We should either fix the test, or delete this.

mkuzdowicz: agree, i was planning to do that later anyway

PR #1065, opened 2019-10-15
save Archiver messages into database

akash1810:
## What's changed?

Save messages read of the SQS queue, originating from Archiver and pertaining to the progress of issue publication, to the database.

This is getting us a step closer to showing this in the UI to provide confidence that publication is happening as currently we say "issue has been submitted to publication, check your device in 5 minutes".

## Implementation notes

If we encounter a foreign key constraint violation, because the `version_id` doesn't exist in the `issue_versions` table, proceed to remove the message from SQS (that is, mark it as successful). This is because we can never recover from this state and it is the only way to progress.

We're only likely to enter this state in DEV and CODE as we share the same Archiver stack. That is, the SQS queue will have messages for DEV and CODE.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/Components.scala
+++ b/app/Components.scala
@@ -19,6 +19,7 @@ import router.Routes
 import services._
 import services.editions.EditionsTemplating
 import services.editions.db.EditionsDB
+import services.editions.publishing.events.PublishEventsListener
 import services.editions.publishing.{EditionsBucket, EditionsPublishing}
 import slices.{Containers, FixedContainers}
 import thumbnails.ContainerThumbnails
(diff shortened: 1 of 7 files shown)


app/Components.scala
```diff
@@ -49,6 +50,7 @@ class AppComponents(context: Context, val config: ApplicationConfiguration)
   val publishingBucket = new EditionsBucket(s3Client, config.aws.publishedEditionsIssuesBucket)
   val previewBucket = new EditionsBucket(s3Client, config.aws.previewEditionsIssuesBucket)
   val editionsPublishing = new EditionsPublishing(publishingBucket, previewBucket, editionsDb)
+  PublishEventsListener.apply(config, editionsDb).start
```
jonathonherbert: Just out of interest -- why the move?

app/services/editions/publishing/events/package.scala
```diff
@@ -13,7 +15,20 @@ package object events {
     status: IssueVersionStatus,
     message: String,
     timestamp: LocalDateTime
-  )
+  ) {
+    def toLogMarker: LogstashMarker = {
```
jonathonherbert: Sweet pattern yo 👍

app/Components.scala
akash1810: `PublishEventsListener` needs an `EditionsDB` now; rather than construct a new one, moving the instantiation here means we can reused the same instance of `EditionsDB` throughout.

akash1810:
> "Pooling"? Polling?

Yes. Outside of the scope of this PR though 😄

PR #1067, opened 2019-10-15
ATOMS: Get an interactive atom from capi & create snap card for it

ajwl:
## What's changed?

This is a change to make adding snaps easier for the Interactives team. It makes use of an existing workflow to add interactives to CAPI as interactive atoms. 

**The old way:**
Users have to hand-craft a url like this, adding certain query params to make sure it does the right thing

```
https://gu.com?gu-snapType=json.html&gu-snapUri=https://interactive.guim.co.uk/atoms/2019/03/29/unmeaningful-vote/snap/snap.json
```

**The new way:** 
Users get a CAPI url for their interactive atom and drop that in, this pulls the title from the url

```
https://content.guardianapis.com/atom/interactive/interactives/2017/06/general-election
```
Screenshot 2019-10-16 at 17 13 39

CAPI links that are invalid or point to atoms of the wrong type will return invalid atom cards. 
```
https://content.guardianapis.com/atom/guide/166615a1-ed30-43e9-9201-083a41ce7713
```
Screenshot 2019-10-16 at 17 12 25

## Implementation notes
* Using the interactive atom as the canonical format for interactives - whether on articles, live blogs or the front gives us consistency of data structures. 

* The old method is still supported. Nothing is broken

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/services/faciaApi.ts
+++ b/client-v2/src/services/faciaApi.ts
@@ -18,11 +18,7 @@ import {
 import pandaFetch from './pandaFetch';
 import { CapiArticle } from 'types/Capi';
 import chunk from 'lodash/chunk';
-import {
-  CAPISearchQueryResponse,
-  checkIsResults,
-  CAPIInteractiveAtomResponse
-} from './capiQuery';
+import { CAPISearchQueryResponse, checkIsResults } from './capiQuery';
 import flatMap from 'lodash/flatMap';
 import { EditionsIssue, EditionsCollection } from 'types/Edition';
 import { EditionsRoutes } from 'routes/routes';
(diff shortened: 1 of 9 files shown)


ajwl:
cc @jennygrahamjones 
Would need to call this new func from inside Card.tsx I think

client-v2/src/services/faciaApi.ts
```diff
@@ -338,9 +342,18 @@ const getCapiUriForContentIds = (contentIds: string[]) => {
     .join('&')}`;
 };
 
+// getting interactive atoms from CAPI
+const getAtomFromCapi = async (path: string) => {
+  const response = await pandaFetch(`/api/live/${path}`, {
+    method: 'get',
+    credentials: 'same-origin'
+  });
+  const parsedResponse: CAPIInteractiveAtomResponse = await response.json();
```
ajwl: just return?

PR #1075, opened 2019-10-21
Revert "Remove the 'shared' state from the clipboard"

jennygrahamjones:
Reverts guardian/facia-tool#1063


ajwl:
Checked tests + integration tests and basic behaviour. Seems to be fine 👍

jennygrahamjones:
This revert was made to resolve a bug in production. Users adding sublinks to articles did not see a 'Launch' button, even though the contents of the collection had changed.

jonathonherbert:
Thanks for resolving @jennygrahamjones -- I'll take a look, and see if we can't get a test in there to spot this regression before it hits PROD 👍

PR #1077, opened 2019-10-21
More accurate logging and status code for /stories-visible endpoint

mkuzdowicz:
More accurate logging and status code for /stories-visible endpoint
## What's changed?

- logging added to StoriesVisibleController for `/stories-visible` endpoint
- if container id is invalid return bad-request not `Not-Found` status

### General
- [x] 🔍 Checked on CODE

### DIFF
--- a/app/controllers/StoriesVisibleController.scala
+++ b/app/controllers/StoriesVisibleController.scala
@@ -20,7 +20,7 @@ class StoriesVisibleController(val containerService: ContainerService, val deps:
     logger.info(s"got stories-visible=$storiesVisible for containerType=$containerType")
 
     storiesVisible.map { storiesVisibleResponse => Ok(Json.toJson(storiesVisibleResponse)) } getOrElse {
-      val errorMSG = s"$containerType is not a valid container id"
+      val errorMSG = s"'$containerType' is not a valid container id"
       logger.error(errorMSG)
       BadRequest(errorMSG)
     }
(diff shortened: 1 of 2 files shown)


app/controllers/StoriesVisibleController.scala
```diff
@@ -12,12 +13,16 @@ case class StoriesVisibleRequest(
...
-class StoriesVisibleController(val containerService: ContainerService, val deps: BaseFaciaControllerComponents) extends BaseFaciaController(deps) {
+class StoriesVisibleController(val containerService: ContainerService, val deps: BaseFaciaControllerComponents) extends BaseFaciaController(deps) with Logging {
   def storiesVisible(containerType: String) = AccessAPIAuthAction(parse.json[StoriesVisibleRequest]) { implicit request =>
     val storiesVisible = containerService.getStoriesVisible(containerType, request.body.stories)
 
+    logger.info(s"got stories-visible=$storiesVisible for containerType=$containerType")
+
     storiesVisible.map { storiesVisibleResponse => Ok(Json.toJson(storiesVisibleResponse)) } getOrElse {
-      NotFound(s"$containerType is not a valid container id")
+      val errorMSG = s"'$containerType' is not a valid container id"
```
aug24: Type or id?  I think these are different. Can you check this error message is correct please?

mkuzdowicz: will change :) that log was like that before, but it make sense

PR #1078, opened 2019-10-22
Saving content prefill time window into DB while issue creation, and getting that back while getPrefillForCollection request

mkuzdowicz:
Saving content prefill time window into DB while issue creation, and getting that back while getPrefillForCollection request

## Implementation notes

- code flow changes to be able to get content prefill time window in `EditionsController.createIssue` function
- adding 2 new columns to collections table
- inserting content prefill time window to collections table while issue creation
- getting content prefill time window from DB while getPrefillForCollection request

## Trello ticket:

[trello card](https://trello.com/c/mbbmfswB/144-insert-window-start-and-end-into-db-as-utc-on-issue-creation)

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


app/services/editions/EditionsTemplating.scala
```diff
@@ -5,51 +5,105 @@ import java.time.LocalDate
...
+import services.editions.prefills.PrefillHelper.defineContentQueryTimeWindow
+import services.editions.prefills.{CapiQueryTimeWindow, Prefill, PrefillParamsAdapter}
 import services.{Capi, Ophan}
 
 import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.language.postfixOps
 import scala.util.control.NonFatal
 
+case class GenEditionTemplateResult(issueSkeleton: EditionsIssueSkeleton, contentPrefillTimeWindow: CapiQueryTimeWindow)
```
aug24: "GenEditionTemplateResult" is not the clearest name.  What is it?

conf/evolutions/default/9.sql
```diff
@@ -0,0 +1,12 @@
+
+# --- !Ups
+
+ALTER TABLE collections ADD COLUMN content_prefill_window_start TIMESTAMPTZ;
+ALTER TABLE collections ADD COLUMN content_prefill_window_end TIMESTAMPTZ;
+UPDATE  collections  SET content_prefill_window_start = CONCAT(cast((select issue_date from edition_issues where edition_issues.id = (select issue_id from fronts where fronts.id = front_id)) as text), ' 01:00:00.00 UTC')::TIMESTAMP where prefill IS NOT NULL;
```
aug24: Why from 1am?

mkuzdowicz: i will add a comment on that, 
I think issue_date is in London time
that one is in UTC and effectively i was getting day before with 23:00:00 time
like issue_date was `2019-10-11` and then the start and end date as part fo migration were `2019-10-10T23:00:00:00`

So it is to keep the same day

app/services/editions/EditionsTemplating.scala
mkuzdowicz: yeah agree :)
The main idea is that this is a result of `generateEditionTemplate` function
where the skeleton is the actual data produced, additionally time-window is passed here as it is needed to persist that in DB

I can rename it to `GenerateEditionTemplateResult` ?

mkuzdowicz: done

conf/evolutions/default/9.sql
mkuzdowicz: done

app/services/Capi.scala
```diff
@@ -20,8 +20,6 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
 class GuardianCapi(config: ApplicationConfiguration)(implicit ex: ExecutionContext)
   extends GuardianContentClient(apiKey = config.contentApi.editionsKey) with Capi with Logging {
 
-  private def prefillHelper = PrefillHelper(EditionsTemplates.templates)
```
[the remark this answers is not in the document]
mkuzdowicz: it is used in `EditionsTemplating` as well , the `defineContentQueryTimeWindow` function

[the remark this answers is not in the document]
mkuzdowicz: `Capi` is used only for prefills here so i would put that into prefill package as well and have some main prefill service or prefill helper that have the main public functions like `getPrefillArticles`

Now it may be a bit confusing

We generate prefills while 2 requests: `EditionsController.createIssue` and `EditionsController.getPrefillForCollection` 

`EditionsController.createIssue` call `generateEditionTemplate` which call `capi.getUnsortedPrefillArticleItems`

`EditionsController.getPrefillForCollection`  call `capi.getPrefillArticles`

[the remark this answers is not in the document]
mkuzdowicz: i am keen to here your suggestions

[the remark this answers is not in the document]
mkuzdowicz: added changes related to our conversation

PR #1083, opened 2019-10-24
Fixes capi queries

blishen:
Fixes CAPI queries for Oz containers

### DIFF
--- a/app/model/editions/templates/AustralianEdition.scala
+++ b/app/model/editions/templates/AustralianEdition.scala
@@ -51,8 +51,8 @@ object AustralianEdition {
   def FrontNewsAu = front(
     "News",
     Some("au"),
-    collection("News").searchPrefill("?tag=(tone/news,australia-news/australia-news)-tone/comment"),
-    collection("World News").searchPrefill("?tag=(tone/news-australia-news/australia-news)-tone/comment")
+    collection("News").searchPrefill("?tag=tone/news,australia-news/australia-news,-tone/comment"),
+    collection("World News").searchPrefill("?tag=tone/news,-australia-news/australia-news,-tone/comment")
   )
   .swatch(News)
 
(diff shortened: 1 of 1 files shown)


app/model/editions/templates/AustralianEdition.scala
```diff
@@ -51,8 +51,8 @@ object AustralianEdition {
   def FrontNewsAu = front(
     "News",
     Some("au"),
-    collection("News").searchPrefill("?tag=(tone/news,australia-news/australia-news)-tone/comment"),
-    collection("World News").searchPrefill("?tag=(tone/news-australia-news/australia-news)-tone/comment")
+    collection("News").searchPrefill("?tag=tone/news,australia-news/australia-news,-tone/comment"),
+    collection("World News").searchPrefill("?tag=tone/news,-australia-news/australia-news,-tone/comment")
```
aug24: news, but not australia news, and not comment?  Right?

app/model/editions/templates/AustralianEdition.scala
```diff
@@ -61,8 +61,8 @@ object AustralianEdition {
   def FrontJournalAu = front(
     "Journal",
     Some("au"),
-    collection("Opinion").searchPrefill("?tag=(australia-news/australia-news,tone/comment)"),
-    collection("World Opinion").searchPrefill("?tag=(tone/comment-australia-news/australia-news)")
+    collection("Opinion").searchPrefill("?tag=(australia-news/australia-news,tone/comment)|tracking/commissioningdesk/australia-opinion"),
+    collection("World Opinion").searchPrefill("?tag=(tone/comment,-australia-news/australia-news)|(tone/comment,-tracking/commissioningdesk/australia-opinion)")
```
aug24: I think this could be simplified as both the OR elements have tone/comment?  Then I suspect the OR won't produce the effect you want as elements in the second section could be australia news, and elements in the first could be australian opinion?

tone/comment,(-australia-news/australia-news,-tracking/commissioningdesk/australia-opinion)

??

app/model/editions/templates/AustralianEdition.scala
```diff
@@ -82,14 +82,14 @@ object AustralianEdition {
   def FrontCultureAu = front(
     "Culture",
     Some("au"),
-    collection("Culture").searchPrefill("?tag=tone/features|tone/interviews,-tone/reviews+(music/music|books/books|stage/stage|music/classical-music-and-opera|artanddesign/artanddesign|games/games|tv-and-radio/tv-and-radio|film/film|culture/culture)")
+    collection("Culture").searchPrefill("?tag=((tone/features|tone/interviews),-tone/reviews),(music/music|books/books|stage/stage|music/classical-music-and-opera|artanddesign/artanddesign|games/games|tv-and-radio/tv-and-radio|film/film|culture/culture)")
```
aug24: Lose the outermost parentheses?  They appear to be around the whole thing.

PR #1094, opened 2019-11-12
Add an election topic to breaking news

jonathonherbert:
## What's changed?

Adds a new election topic to the breaking news part of the tool. Breaking news containers with the link name 'election' will send alerts with this topic.

In draft, as we'll need to merge https://github.com/guardian/mobile-n10n/pull/478 before we can build this PR.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/updates/BreakingNewsUpdate.scala
+++ b/app/updates/BreakingNewsUpdate.scala
@@ -89,7 +89,7 @@ class BreakingNewsUpdate(val config: ApplicationConfiguration, val ws: WSClient,
 
   private def createPayload(trail: ClientHydratedTrail, email: String): BreakingNewsPayload = {
     BreakingNewsPayload(
-      message = StringEscapeUtils.unescapeHtml4(trail.headline),
+      message = Some(StringEscapeUtils.unescapeHtml4(trail.headline)),
       thumbnailUrl = trail.thumb.map{new URI(_)},
       sender = email,
       link = createLinkDetails(trail),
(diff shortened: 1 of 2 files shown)


app/updates/BreakingNewsUpdate.scala
```diff
@@ -136,6 +136,7 @@ class BreakingNewsUpdate(val config: ApplicationConfiguration, val ws: WSClient,
       case Some("uk") => List(BreakingNewsUk)
       case Some("us") => List(BreakingNewsUs)
       case Some("sport") => List(BreakingNewsSport)
+      case Some("election") => List(BreakingNewsElection)
```
alexduf: So in my email that was the unknown part. How does the editions tool receive the "election" topic?
There must be a place where the container is mapped to this string but I fail to see it.

I've also changed that string on the mobile side to be `uk-general-election` just so that we can have other elections in future if necessary.

build.sbt
```diff
@@ -97,7 +97,7 @@ libraryDependencies ++= Seq(
     "com.gu" %% "editorial-permissions-client" % "2.0",
     "com.gu" %% "fapi-client-play26" % "3.0.11",
     "com.gu" % "kinesis-logback-appender" % "1.4.2",
-    "com.gu" %% "mobile-notifications-api-models" % "1.0.0",
+    "com.gu" %% "mobile-notifications-api-models" % "1.0.4-SNAPSHOT",
```
alexduf: I've now released `1.0.4`

app/updates/BreakingNewsUpdate.scala
jonathonherbert: Ah, righto. This comes straight from the client -- on alert send, the client sends a `ClientHydratedCollection` payload which looks like ```
{
  trails: [{
    alert: true
    group: "major"
    headline: "Waves in St Mark's Square as Venice flooded by highest tide in 50 years"
    image: "https://media.guim.co.uk/a09101c5ee771c78ff8d16bdf0f49b5309c34df9/0_131_3072_1842/500.jpg"
    topic: "election"
  }]
}
```

The topic is derived from the `meta.href` property of the collection, which is manually set by the user in the Config tool. If there's not a match, the server sends back an error and the tool tells the user that their alert has failed to send.

jonathonherbert: I'lll change the string accordingly 👍

alexduf: I see, make sense!

PR #1097, opened 2019-11-14
Add a way of adding free-text snap links without leaving the tool

jonathonherbert:
## What's changed?

This PR adds a way of adding free-text snap links without leaving the tool.

create-snap

## Implementation notes

This rather 𝓮𝓵𝓮𝓰𝓪𝓷𝓽 hack (YMMV 😁) depends on the way we handle domains and url params when ingesting snaplinks -- specifically, that we understand incoming links from gu.com to be snaplinks, and that we treat url params prefixed `gu-` in a special way. See the comments on `DragToAddSnap.tsx`, ln36.

Eventually, we should be able to migrate to a snap type of 'text' or 'html' or something and remove this workaround.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/integration/selectors/index.js
+++ b/client-v2/integration/selectors/index.js
@@ -36,6 +36,8 @@ const OPTIONS_MODAL = 'options-modal';
 const GUARDIAN_TAG_ANCHOR = 'guardian-tag';
 const EXTERNAL_LINK_ANCHOR = 'external-link';
 
+const DRAG_TO_ADD_SNAP = 'drag-to-add-snap';
+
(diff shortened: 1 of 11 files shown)


ajwl:
Ideally this would be different `snapType`: "Free text" or some such. But this may be part of the work on changing the card shape so happy for it to go in there.

jonathonherbert:
@ajwl agreed -- this is a bit of a stopgap. When we've got our nice new snap format, it shouldn't be much work to swap in.

PR #1102, opened 2019-11-15
Display print page numbers on suggested articles

mkuzdowicz:
## What's changed?

Be able to display print page numbers on suggested articles

## Screenshots

Screenshot 2019-11-15 at 17 38 28

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/util/ArticlePageNumberSection.tsx
+++ b/client-v2/src/components/util/ArticlePageNumberSection.tsx
@@ -1,0 +1,21 @@
+import CardMetaContent from '../../shared/components/card/CardMetaContent';
+import { CapiArticle } from 'types/Capi';
+import * as React from 'react';
+
+interface ComponentProps {
+  article?: CapiArticle;
+}
+
+const articlePageNumberSection = ({ article }: ComponentProps) => {
+  if (article && article.fields && article.fields.newspaperPageNumber) {
+    return (
+      <CardMetaContent title="The newspaper page number of this article">
+        Page {article.fields.newspaperPageNumber}
+      </CardMetaContent>
+    );
+  } else {
+    return null;
+  }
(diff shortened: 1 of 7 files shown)


client-v2/src/components/util/ArticlePageNumberSection.tsx
```diff
@@ -0,0 +1,21 @@
+import CardMetaContent from '../../shared/components/card/CardMetaContent';
+import { CapiArticle } from 'types/Capi';
+import * as React from 'react';
+
+interface ComponentProps {
+  article?: CapiArticle;
```
akash1810: Could we move this nullability up? That is, I think this component could be simplified if it has an article to render against.

mkuzdowicz: can you tel more details what you mean :) ?

I do nullability here as the only logic that this component does is: if article is not defined do not display  `CardMetaContent` with page number, display otherwise

It is very easy to unit test 

If i move it up i will have to test it by testing `FeedItem.tsx` which is much larger and does not have unit tests as far as i can see for now

mkuzdowicz: resolved in person :)

client-v2/src/types/Capi.ts
```diff
@@ -85,6 +85,7 @@ interface Tag {
 type CapiBool = 'true' | 'false' | boolean;
 
 interface CapiArticleFields {
+  newspaperPageNumber?: string;
```
akash1810: could this be the `number` type?

mkuzdowicz: done

PR #1106, opened 2019-11-18
Alter fields for HTML snaps

jonathonherbert:
_NB: Don't merge this just yet! It's missing tests, and I'd quite like to add some. But, gentle reader, you are of course more than welcome to review._

## What's changed?

Hot on the heels of #1105, this PR alters the fields for the `html` snap. Users of this snap type are allowed precisely three: 

Screenshot 2019-11-18 at 21 14 54

That's yer lot.

## Implementation notes

I've added some commenting, as we repurpose the `headline` field here for our own nefarious purposes (although, it was always being used this way -- we've simply called it something nice for our users that isn't `headline`). Once e-mail renderer has caught up, we should be able to use a proper `html` field for our proper `html` snap.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/FrontsEdit/CollectionComponents/DragToAddSnap.tsx
+++ b/client-v2/src/components/FrontsEdit/CollectionComponents/DragToAddSnap.tsx
@@ -64,7 +64,7 @@ const DragToAddTextSnap = () => {
         onDragStart={e => handleDragStart(e, ref.current)}
         draggable={true}
       >
-        <DragIcon /> Drag to add a text snap
+        <DragIcon /> Drag to add a text card
       </DragToAddSnapContainer>
     </>
   );
--- a/client-v2/src/components/FrontsEdit/CardFormInline.tsx
+++ b/client-v2/src/components/FrontsEdit/CardFormInline.tsx
@@ -408,7 +408,7 @@ class FormComponent extends React.Component<Props, FormComponentState> {
               <Field
                 name="headline"
                 label={this.getHeadlineLabel()}
-                rows={this.getHeadlineRows()}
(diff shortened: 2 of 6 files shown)


client-v2/src/components/FrontsEdit/CollectionComponents/DragToAddSnap.tsx
```diff
@@ -33,13 +32,8 @@ const handleDragStart = (
   event: React.DragEvent<HTMLDivElement>,
   dragImageElement: HTMLDivElement | null
 ) => {
-  // We must provide a unique URL for our snaplink, otherwise platforms might deduplicate
-  // this content when rendering it on a front. Because we know that in a free-text snaplink,
-  // the URL is not used, we supply a URL here that's a) guaranteed to be unique, and b)
-  // gives some indication why it's there. When snaplinks don't need URLs, possibly because
-  // there's a snap type for inserting arbitrary HTML, this will no longer be necessary.
```
ajwl: This comment seems quite useful. Leave it in?

ajwl:
Two suggested rewordings: 
"Drag to add a text snap" >> "Drag to add a free text box"   (snap wouldn't be meaningful to a new user)
"HTML Content" >> "Content"  (we know it's html, but to the user it isn't)

One gremlin:
Somehow this has got the weird shrinking text-box behaviour where the box closes in as you type. I didn't see this in the merged version on Code, so it may be fixed when it gets added to all the other changes. But since we're merging individually, may be good to fix here. 

Otherwise all looks good 👍 👍

PR #1107, opened 2019-11-20
Dependency cleanup: Bump Capi & facia-scala-client versions

ajwl:
## What's changed?
Various dependencies had got out of date which was preventing facia-scala-client upgrades being applied to the Fronts tool

This brings us almost up to date with facia-scala-client, and means that the capi version matches between Facia-scala-client and Facia tool. 

Several other Scala libraries are getting out of date (run `sbt eviction` to see what...) so we should return at some point to make these fixes. 

**Why `content-api-models-scala` instead of `content-api-models`?** 
This is now recommended practice - see: https://github.com/guardian/content-api-models#information-about-built-jars

**Why the deletion of this line?**  
`addSbtPlugin("io.get-coursier" % "sbt-coursier" % "2.0.0-RC2")`
The newer version of sbt includes a version of `coursier` that is incompatible with the trial version we had incorporated manually into the project. This was causing errors. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -61,8 +61,8 @@ TwirlKeys.templateImports ++= Seq(
 routesImport += "model.editions._"
 
 val awsVersion = "1.11.293"
-val capiModelsVersion = "14.1"
-val capiClientVersion = "14.3"
+val capiModelsVersion = "15.4"
+val capiClientVersion = "15.4"
 val json4sVersion = "3.6.0-M2"
 val enumeratumPlayVersion = "1.5.13"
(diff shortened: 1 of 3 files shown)


ajwl:
This is some preparatory work so we can add the atomid to the facia-scala-client

ajwl:
This has deployed successfully in CODE so when the tests go green again, I'm going to merge.. 🚢

PR #1111, opened 2019-11-22
ATOM ID: Add atomId to the data in an atom snaplink

ajwl:
## What's changed?
When snap links are created with links to atoms in the content api eg: 
`https://content.guardianapis.com/atom/interactive/interactives/2019/10/test-snap`. 

We want to extract the `atomId` - the path of the atom and send it on in a new optional `atomId` property.

This makes it easier for downstream teams like Dotcom and MAPI to fetch the atom from CAPI. 

This is follow-up work to https://github.com/guardian/facia-tool/pull/1067

It can be released after the new version of Facia-Scala-Client is updated: https://github.com/guardian/facia-scala-client/pull/231

The resulting json looks like this: 
 ```
{
	"id": "snap/1574361251494",
	"frontPublicationDate": 1574361316176,
	"meta": {
		"headline": "Test snap atom",
		"snapType": "interactive",
		"showByline": false,
		"atomId": "/atom/interactive/interactives/2019/10/test-snap",
		"snapUri": "https://content.guardianapis.com/atom/interactive/interactives/2019/10/test-snap",
		"byline": "Guardian Visuals",
		"href": "https://content.guardianapis.com/atom/interactive/interactives/2019/10/test-snap"
	}
}
```

**The version for Facia-Scala-Client will need to be updated in downstream apps.** 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -95,7 +95,7 @@ libraryDependencies ++= Seq(
     "com.gu" %% "content-api-client-aws" % "0.5",
     "com.gu" %% "content-api-client-default" % capiClientVersion,
     "com.gu" %% "editorial-permissions-client" % "2.0",
-    "com.gu" %% "fapi-client-play26" % "3.0.14",
+    "com.gu" %% "fapi-client-play26" % "3.0.16",
     "com.gu" % "kinesis-logback-appender" % "1.4.2",
     "com.gu" %% "mobile-notifications-api-models" % "1.0.5",
     "com.gu" %% "pan-domain-auth-play_2-6" % "0.7.2",
--- a/build.sbt
+++ b/build.sbt
@@ -95,7 +95,7 @@ libraryDependencies ++= Seq(
     "com.gu" %% "content-api-client-aws" % "0.5",
     "com.gu" %% "content-api-client-default" % capiClientVersion,
     "com.gu" %% "editorial-permissions-client" % "2.0",
-    "com.gu" %% "fapi-client-play26" % "3.0.16",
+    "com.gu" %% "fapi-client-play26" % "3.0.17",
     "com.gu" % "kinesis-logback-appender" % "1.4.2",
     "com.gu" %% "mobile-notifications-api-models" % "1.0.5",
     "com.gu" %% "pan-domain-auth-play_2-6" % "0.7.2",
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by ajwl** [comment]
@tjmw / @alexduf - adding you so you can check the json output above. 
--> what the fronts tool now returns for items created out of atom links. 
Does this look like what you wanted?

File: build.sbt [hunk 1]
```diff
@@ -95,7 +95,7 @@ libraryDependencies ++= Seq(
     "com.gu" %% "content-api-client-aws" % "0.5",
     "com.gu" %% "content-api-client-default" % capiClientVersion,
     "com.gu" %% "editorial-permissions-client" % "2.0",
-    "com.gu" %% "fapi-client-play26" % "3.0.14",
+    "com.gu" %% "fapi-client-play26" % "3.0.16",
```
ajwl: Update to the correct version once https://github.com/guardian/facia-scala-client/pull/231 is officially released

tjmw:
Hi @ajwl - great, yep that looks like what I was expecting! One question about the new `atomId` field - you mention in the description that it's optional. Should it always be there when the snap type is `interactive`? Are there times when we might still have to fall back on extracting from the URL?

ajwl:
@tjmw good question. If the link pasted into the tool is for a valid interactive atom, then a card of type `interactive` is created and it will have the `atomId`.
--> we do that validation and id extraction in the tool. You should never need to parse the url. 

It is optional because the Fronts Tool has several other types of snap links that do not involve atoms eg: to `https://www.bbc.co.uk/news` or to Guardian tag pages. In these cases `atomId` is not set. These have different `snapType` properties.

ajwl:
@jonathonherbert / @jennygrahamjones -> you should probably take a look at this before it gets merged. I'll be on holiday so just merge if you're happy

@alexduf & @tjmw -> when it does you'll need to update Dotcom and MAPI with "facia-scala-client" version `3.0.17` to get the benefits of the changes. 

NB: if you need to jump from below facia-scala-client 3.0.12 - a big change has been added in .12-.14 updating to a new version of capi models. So you will need to make sure you are on the right version of capi models. 
This change may be helpful: https://github.com/guardian/facia-tool/pull/1107

PR #1112, opened 2019-11-22
Improve delete functionality

aug24:
## What's changed?
Simplifies the process for deleting Issues. Very useful for testing.

## Implementation notes
Deliberately doesn't make it 'easy' for real users.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


client-v2/src/util/delete.ts
```diff
@@ -0,0 +1,17 @@
+export const deleteIssue = () => {
+  const issue = window.location.href.substring(
+    window.location.href.lastIndexOf('/') + 1
+  );
```
mkuzdowicz: `window.location.href` may not work on all browsers, you can add to README like use chrome or so

PR #1116, opened 2019-11-27
Handle Google redirect URLs

jennygrahamjones:
## What's changed?

When dragging or copying a link from Google (Hangouts, Mail, Drive...) into the Fronts tool, the link may be prepended with a Google 'redirect' that looks like `https://www.google.com/url?q=https://example.com/foobar&sa=D&source=hangouts...`

As a consequence, links copied from Google into the tool aren't handled correctly. For an example, see the below image. The first snap is how the link should be handled and the second is how it is currently handled.

Screenshot 2019-11-26 at 11 43 21

This PR introduces changes to ensure that links dragged or copied from Google are interpreted correctly by the tool.

## Implementation notes

When we use pandaFetch to determine the title and descriptions for a URL like `https://www.google.com/url?q=https://example.com/foobar&sa=D&source=hangouts...`, Google returns a 200 rather than a 3xx code that we can follow. 

To work around this, we can identify up-front if the link contains a Google redirect, then extract the relevant URL from the query parameters. The potential URL is extracted from a parameter called `q` and is then validated to confirm it is a URL. 
### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/actions/Cards.ts
+++ b/client-v2/src/shared/actions/Cards.ts
@@ -180,7 +180,10 @@ const getArticleEntitiesFromDrop = async (
   if (drop.type === 'CAPI') {
     return getArticleEntitiesFromFeedDrop(drop.data, isEdition);
   }
-  const resourceIdOrUrl = drop.data.trim();
+  const droppedDataURL = drop.data.trim();
+  const resourceIdOrUrl = isGoogleRedirectUrl(droppedDataURL)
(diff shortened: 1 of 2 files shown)


aug24:
FWIW, google are very very unlikely to rename the 'q' param.  It's a web standard that they suggested about fifteen years ago.

jennygrahamjones:
> FWIW, google are very very unlikely to rename the 'q' param. It's a web standard that they suggested about fifteen years ago.

Good to know!

PR #1072, opened 2019-10-17
RICH TEXT EDITOR: Add a prosemirror to enable rich text editing

ajwl:
## What's changed?
A new rich text editor for the Headline input on some cards. 

* Only available on Email fronts 

* Once other work is completed, this should only be available on a new type of "free text card" (see #1097 ).

Email editors often add free text sections representing "notes from the editor", or some introductory text. They often want to add links or bold / italics in here. Doing it in the current section involves handwriting html which gets tedious. The new editor should make this easier and quicker. 

Screenshot 2019-11-15 at 11 36 14

## Implementation notes

* This includes a lot of code lifted and shifted from https://github.com/guardian/flexible-content (specifically from the [prosemirror-setup](https://github.com/guardian/flexible-content/tree/master/composer/src/js/prosemirror-setup) directory) with minimal changes made, aside from to limit the types of styling that can be applied. 

* Elements that can form part of outputted HTML are `<strong>`,`<em>`, `<a>`, `<p>` and `&nbsp;`

* The final effect of the outputted html will be seen in the email renderer and should be checked. 

* Integration test added for bolding/unbolding and adding/removing a link 

### General
- [ ] 🤖 Relevant tests added
- [ ] ✅ CI checks / tests run locally
- [ ] 🔍 Checked on CODE

### Client
- [ ] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ ] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #1124, opened 2019-12-04
Rich text: Add hard line break on 'Enter'

jennygrahamjones:
## What's changed?

Our rich text editor for email fronts previously (#1072) created a new `<p>` tag whenever users pressed the enter key. In the Fronts tool, this could be used to create spacing between paragraphs. At the point of rendering in preview, the spacing was lost. 

Instead of adding a new paragraph, the tool now inserts a `<br>` tag. This ensures that there is parity between the styling in the Fronts tool and the styling in preview. It also means that no change is required to the styling applied during rendering.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/components/inputs/richtext/setup.ts
+++ b/client-v2/src/components/inputs/richtext/setup.ts
@@ -1,6 +1,5 @@
 import { buildKeymap } from './utils/keymap';
 import { keymap } from 'prosemirror-keymap';
-import { baseKeymap } from 'prosemirror-commands';
 import { Schema, DOMSerializer, DOMParser, NodeSpec } from 'prosemirror-model';
 import { schema } from 'prosemirror-schema-basic';
 import { addListNodes } from 'prosemirror-schema-list';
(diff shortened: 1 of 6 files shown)


client-v2/src/components/inputs/richtext/setup.ts
```diff
@@ -14,18 +13,13 @@ import { history } from 'prosemirror-history';
 const createBasePlugins = (s: Schema) => {
   const plugins = [
     keymap(buildKeymap(s, {}, {})),
-    keymap(baseKeymap),
```
ajwl: Was `baseKeymap` just not necessary?

jennygrahamjones: Yeah, it was importing (and potentially overriding) bindings that we were declaring ourselves.

ajwl:
Oh I guess <p> needs to be changed to <br> in the integration tests too...

PR #1131, opened 2019-12-10
Cap all pre-filled AUS containers

blishen:
## What's changed?
For all prefilled containers we now have a cap of 40 stories.
The new queries were returning more content than the old.


LATaylor-guardian:
👍

PR #1140, opened 2020-01-08
Allow renaming of all fronts and containers

aug24:
## What's changed?
See https://trello.com/c/ZU1l81Bq/159-all-containers-to-be-renamable

All containers and fronts are now editable.

### General
- [ ] 🤖 Relevant tests added
- [X ] ✅ CI checks / tests run locally
- [ ] 🔍 Checked on CODE

### Client
- [ X] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ X] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [X ] 📷 Screenshots / GIFs of relevant UI changes included
Screenshot 2020-01-09 at 14 05 39

### DIFF
--- a/client-v2/src/components/FrontsEdit/FrontSection.tsx
+++ b/client-v2/src/components/FrontsEdit/FrontSection.tsx
@@ -230,15 +230,15 @@ class FrontSection extends React.Component<
                   >
                     {isHidden ? 'Unhide' : 'Hide'}
                   </FrontHeaderButton>
-                  <FrontHeaderButton
-                    data-testid="rename-front-button"
-                    onClick={this.renameFront}
-                    size="l"
-                  >
-                    Rename
-                  </FrontHeaderButton>
                 </>
               )}
+              <FrontHeaderButton
+                data-testid="rename-front-button"
+                onClick={this.renameFront}
+                size="l"
+              >
+                Rename
+              </FrontHeaderButton>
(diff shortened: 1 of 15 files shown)


client-v2/src/components/FrontsEdit/FrontSection.tsx
```diff
@@ -119,6 +120,7 @@ type FrontsComponentProps = FrontsContainerProps & {
     ) => void;
     setFrontHiddenState: (id: string, hidden: boolean) => void;
   };
+  isEditions: boolean;
```
jonathonherbert: Is this used in the component? It's new to the props, but I can't see it in render().

client-v2/src/components/FrontsEdit/FrontSection.tsx
```diff
@@ -230,15 +232,16 @@ class FrontSection extends React.Component<
...
-                  <FrontHeaderButton
-                    data-testid="rename-front-button"
-                    onClick={this.renameFront}
-                    size="l"
-                  >
-                    Rename
-                  </FrontHeaderButton>
                 </>
               )}
+              <FrontHeaderButton
```
jonathonherbert: Nit: It would be more idiomatic to style in the component -- in this case at the point where FrontHeaderButton is created on L90.

PR #1142, opened 2020-01-15
Add clearer styling for breaking news error modal

jonathonherbert:
## What's changed?

The error modal for breaking-news looks much like the other modals. This has recently led editorial staff to dismiss it, unaware that a problem had occurred.

This PR makes it very clear that something has gone wrong should a breaking-news alert fail to publish.

Before:

Screenshot 2020-01-13 at 08 40 36

After:

Screenshot 2020-01-15 at 13 36 54

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/public/src/js/models/collections/collection.js
+++ b/public/src/js/models/collections/collection.js
@@ -268,10 +268,11 @@ export default class Collection extends BaseClass {
             reportErrors(new Error(message)); //report to sentry
         })
         .catch(() => {
+            const breakingNewsMsg = 'breaking news alert. The alert has not been sent. Please contact Central Production.';
             const isBreakingNewsAlert = this.front.confirmSendingAlert();
-            const message = `Failed ${action}ing the ${isBreakingNewsAlert ? 'breaking news alert' : 'collection'}`;
+            const message = `Failed ${action}ing the ${isBreakingNewsAlert ? breakingNewsMsg : 'collection'}`;
             this.setPending(false);
-            alert(message);
+            alert(message, 'error');
         });
     }
 
--- a/public/src/js/models/collections/collection.js
+++ b/public/src/js/models/collections/collection.js
@@ -268,7 +268,7 @@ export default class Collection extends BaseClass {
             reportErrors(new Error(message)); //report to sentry
         })
         .catch(() => {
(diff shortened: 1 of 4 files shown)


public/src/js/models/collections/collection.js
```diff
@@ -268,10 +268,11 @@ export default class Collection extends BaseClass {
             reportErrors(new Error(message)); //report to sentry
         })
         .catch(() => {
+            const breakingNewsMsg = 'breaking news alert. The alert may not have been sent. Please contact Central Production for more information.';
```
jennygrahamjones: `The alert may not have been sent.` - it would be nice to be more definite here. Do we have any way of knowing here whether it has been sent, or is there always going to be some uncertainty?

jonathonherbert: It definitely would -- but I don't think we can know without understanding all the different failure modes of our backend and the notification service. Our backend has validation, and the notification service will have its own checks to run before it sends the notification. In addition to this, the notification service will have different error states, too. 

I suspected this work would take a long time, so rather than give a categorical answer, I've left it ambiguous in the hopes that CP will let users know more, with the understanding that we could refine this approach if needed. Perhaps that isn't the right approach, though -- I'd be interested to know your thoughts.

jennygrahamjones: I think this is an improvement on the existing experience already, so happy to approve on that basis.

akash1810: When does this `catch` get exercised? Is it if we've failed to write to App's SNS topic?

jonathonherbert: It happens when our call to the collection action returns an error code. A failure to send a breaking news alert is one way this can happen -- but there are many others, too, I think, as this this code handles publish and discard actions and there are validation steps before each action, too.

jonathonherbert: Hmm, on closer inspection I think we can be pretty confident that an error would mean an alert wasn't sent -- I thought that a notifications client was handling the alert sending, but we're just posting to an endpoint. With that in mind I've changed the wording to remove the ambiguity @jennygrahamjones @akash1810

PR #1149, opened 2020-02-14
Don't allow creation of 'Invalid atom' cards

jennygrahamjones:
## What's changed? 

When a user drops a link into the Fronts tool that _looks_ like an interactive atom, we do one of two things: create a card for that atom, or create an 'Invalid Atom' card. The latter happens when the link appears to match the format for an interactive atom but CAPI has indicated that no atom of that ID can be found. In practice, this caused the press to fail if the invalid atom was left in a collection.

Instead of creating an 'Invalid Atom' card, I've added a modal, to let the user know that the atom they tried to add doesn't exist.
 
Screenshot 2020-02-14 at 16 00 45

Now when a user drops an interactive atom link into the tool, we either create a card for the atom if it is found in CAPI, or display the modal to let the user know they seem to have pasted/dragged in an invalid link.

Happy to take suggestions on alternate wording for the modal!

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/client-v2/src/shared/actions/Cards.ts
+++ b/client-v2/src/shared/actions/Cards.ts
@@ -217,8 +217,8 @@ const getArticleEntitiesFromDrop = async (
     } catch (e) {
       dispatch(
         startOptionsModal(
-          'Invalid atom link',
-          "It looks like you've tried to add an interactive atom that doesn't exist. Check the link and try again.",
+          'Invalid link',
+          'It looks like you’ve tried to add something from our content-api that we don’t accept. Only interactive atoms can be added as cards. Check the link is for an interactive atom and that the link is valid and try again.',
           [],
           noop,
           true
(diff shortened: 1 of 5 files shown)


ajwl:
Looking now! sorry for delay

client-v2/src/shared/actions/Cards.ts
```diff
@@ -198,8 +208,24 @@ const getArticleEntitiesFromDrop = async (
...
+      const atom = await getAtomFromCapi(
+        getAbsolutePath(resourceIdOrUrl, false)
+      );
+      const card = await createAtomSnap(resourceIdOrUrl, atom);
+      return [card];
+    } catch (e) {
+      dispatch(
+        startOptionsModal(
+          'Invalid atom link',
+          "It looks like you've tried to add an interactive atom that doesn't exist. Check the link and try again.",
```
ajwl: One thought on the message - all content-api urls that get pasted in here are funnelled into this modal. Perhaps we should clarify a bit more. 
```
"It looks like you've tried to add something from our content-api that we don't accept. Only interactive atoms can be added as cards. Check the link is for an interactive atom and that the link is valid and try again.
```

ajwl:
All working 👍  happy to approve with suggested message tweak

client-v2/src/shared/actions/Cards.ts
jennygrahamjones: Good point, I hadn't thought of that.

PR #1159, opened 2020-02-19
Removing the Shared Folder: Part 2

ajwl:
## Why we're changing 

The “shared” directory was added in the expectation that a future suite of tools would reuse some of the fronts tools styles and components. It means that some components are in directories called components, others are in a directory called shared. 

Since this future suite of tools doesn’t exist and isn’t planned - it’s not needed. 

The fact that there are two locations for components is confusing. The logic of what is where is not apparent to new user. 

## What's changed?

1. I merged two actions files called `Cards.ts`. 
`/src/shared/actions/Cards.ts` into `/src/actions/Cards.ts`.

2. I merged two types files called `Collections.ts`. This one `/src/shared/actions/Collections.ts` goes into this one: `/src/actions/Collections.ts`. 
The type file was required because there was a conflict in the imports in the new `Cards.ts`. 

3. I updated all the paths to the new file locations

4. I deleted the old files

✅ The Tool works 
✅ The integration tests pass
❌ 6 unit tests fail with this error: (WHY?)

Tests failing in: `src/actions/__tests__/Cards.spec.ts`

On this line: https://github.com/guardian/facia-tool/pull/1159/files#diff-db583ef82abf886090f0201f76870e8fL77

```
TypeError: action is not a function
```


ajwl:
@jonathonherbert - when you get back from holiday...

Neither me, Jenny nor Akash can understand why these tests are failing, after all staring at the code for several hours. 

==> Can you understand this? 

Also can we refactor the top of `actions/Cards.ts` from this point onwards? It feels like some of the function wrapping makes this more complicated than it needs to be. And if 3 devs can't follow it after several hours we probably need to simplify! 

```
const createInsertCardThunk = (action: InsertActionCreator) => (
  persistTo: 'collection' | 'clipboard'
) => (id: string, index: number, cardId: string, removeAction?: Action) => (
  dispatch: Dispatch
) => {
  if (removeAction) {
    dispatch(removeAction);
  }
  // This cast seems to be necessary to disambiguate the type fed to Dispatch,
  // whose call signature accepts either an Action or a ThunkResult. I'm not really
  // sure why.
  dispatch(action(id, index, cardId, persistTo) as Action);
};
```

PR #1182, opened 2020-02-25
Container renames for new navigation

blishen:
## What's changed?
The new navigation displays container names when they differ to the front name.
These changes preempt that and attempt to supply useful defaults to the production team.


blishen:
Starting again as I made a mess of this.

PR #1183, opened 2020-02-25
Removing Shared State: preparatory work 

ajwl:
## What's changed?
As part of the work to remove the shared folder - we also want to spread out the shared part of the state into the root. 

This lays some of the ground work for that change by creating a fixed type for the State, on the way to remove the "SharedState" type. 

Next piece of work is to replace all mentions of "SharedState" with `State['shared']` to be changed finally to `State['groups'] or whatever. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/actions/OptionsModal.ts
+++ b/fronts-client/src/actions/OptionsModal.ts
@@ -15,7 +15,8 @@ const startOptionsModal = (
     description,
     options,
     onCancel,
-    showCancelButton
+    showCancelButton,
+    isOpen: true
   }
 });
 
--- a/fronts-client/src/reducers/modalsReducer.ts
+++ b/fronts-client/src/reducers/modalsReducer.ts
@@ -1,16 +1,7 @@
 import { Action } from 'types/Action';
-import { OptionsModalChoices } from 'types/Modals';
-import { ReactNode } from 'react';
+import { OptionsModalProps } from 'types/Modals';
 
-type OptionsModalState = null | {
-  title: string;
-  description: string | ReactNode;
-  options: OptionsModalChoices[];
-  onCancel: () => void;
-  showCancelButton: boolean;
-};
-
-const optionsModal = (state: OptionsModalState, action: Action) => {
+const optionsModal = (state: OptionsModalProps, action: Action) => {
   switch (action.type) {
     case 'MODAL/START_OPTIONS_MODAL': {
       return action.payload;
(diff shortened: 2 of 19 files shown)


ajwl:
@jennygrahamjones i hope this is a sensible first step.. if you're well let's discuss tomorrow!

fronts-client/src/reducers/modalsReducer.ts
```diff
@@ -1,15 +1,16 @@
 import { Action } from 'types/Action';
 import { OptionsModalChoices } from 'types/Modals';
+import { ReactNode } from 'react';
 
 type OptionsModalState = null | {
   title: string;
-  description: string;
+  description: string | ReactNode;
```
jennygrahamjones: What prompted adding `ReactNode` here?

ajwl: ah yes interesting one.. 
It is defined with ReactNode in here: https://github.com/guardian/facia-tool/blob/master/fronts-client/src/types/Action.ts#L301

So it was type failing because it had two different definitions. A better fix would be to merge them in some way

ajwl: I'll try fix this

jennygrahamjones: ahhh yes, there should probably be a type for the modal, then the reducer and action could both use it?

fronts-client/src/actions/OptionsModal.ts
```diff
@@ -15,7 +15,8 @@ const startOptionsModal = (
     description,
     options,
     onCancel,
-    showCancelButton
+    showCancelButton,
+    isOpen: true
```
ajwl: Had to add this for the type to conform. Guessed it must be true if the modal is being shown?

fronts-client/src/reducers/rootReducer.ts
```diff
@@ -51,7 +52,7 @@ export interface State {
   editor: editorState;
   staleFronts: staleFrontsState;
   form: FormStateMap;
-  optionsModal: OptionsModalState;
+  optionsModal: OptionsModalState | null;
```
ajwl: This was needed for the types to compile. Makes sense as this is not present in the state when there is no modal open

jennygrahamjones: Seems sensible!

ajwl:
Seems to work... I will merge!

PR #1185, opened 2020-02-26
Navigation change: New front, many container renames

blishen:
## What's changed?
To support the new navigation we're renaming a lot of container names. NOTE: this now means that many container names now are the same within a front, and also in many cases, the same name as the front itself. This is because the app will not show the container name when it is the same as the front name. 

We've (Katy V and me) tried to make it so the default names, should they split through unchecked, wouldn't be a disaster in production. Production will typically rename containers as necessary on the night should changes be required.

## Implementation notes
Edited in Github by a product manager (👎 )


jennygrahamjones:
Screenshot 2020-02-26 at 13 36 57

If this is what you'd expect to see when managing these fronts, then this looks good?

blishen:
@jennygrahamjones - Yes, that's expected, and incredibly - even desired ;-)

PR #1186, opened 2020-02-26
Removing Shared: Remove shared state

ajwl:
## What's changed?
As part of the large piece of work to remove the Shared folder from this tool. We are are also removing the concept of `Shared` from the state object. 

Instead, all concepts that were on the Shared property are now on the root of the state object. This simplifies some code, allows for a flatter state structure, and removes this abstraction that is no longer useful. 

As part of the work to do this - we have provided a more static type definition for the state, instead of using one dynamically generated from reducers. 

**Why we were removing Shared:** 

The “shared” directory was added in the expectation that a future suite of tools would reuse some of the fronts tools styles and components. It means that some components are in directories called components, others are in a directory called shared.

Since this future suite of tools doesn’t exist and isn’t planned - it’s not needed.

The fact that there are two locations for components is confusing. The logic of what is where is not apparent to new user.

## Implementation notes

* Most of this work is amending Fixtures which now should reflect the different state shape. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fronts-client/src/bundles/collectionsBundle.ts
```diff
@@ -83,9 +83,9 @@ const collectionActions = {
 type CollectionActions = Actions<Collection> | SetHidden;
 
 const collectionReducer = (
-  state: State<Collection>,
+  state: LibState<Collection>,
```
jennygrahamjones: Not necessarily in this PR, but may be worth looking at why some of our slices `State` of state need this pattern.

fronts-client/src/selectors/shared.ts
```diff
@@ -17,33 +17,32 @@ import {
...
-import { SharedState, State } from '../types/State';
+import { State } from 'types/State';
 import { cardSets } from 'constants/fronts';
 import { createShallowEqualResultSelector } from 'util/selectorUtils';
 import { DerivedArticle } from 'types/Article';
 import { hasMainVideo } from 'util/externalArticle';
 
 // Selects the shared part of the application state mounted at its default point, '.shared'.
-const selectSharedState = (rootState: any): State['shared'] => rootState.shared;
+const selectSharedState = (rootState: any): State => rootState;
```
jennygrahamjones: Looking forward to getting rid of this next 😁

fronts-client/src/bundles/collectionsBundle.ts
ajwl: `asyncResourceBundles` could do with some documentation in here

PR #1187, opened 2020-02-27
Removing Shared: Delete selectSharedState selector

ajwl:
## What's changed?
As part of the bigger project to remove the Shared folder and the shared part of the state, we want to remove the `selectSharedState` selector which is now no longer needed. 

1. Deletes the `selectSharedState` selector
2. Removes all references to it. Properties are now on the state root so it is not needed
3. Lots of tests and fixtures needed to be updated to remove the concept of shared existing on the state. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/actions/__tests__/Cards.spec.ts
+++ b/fronts-client/src/actions/__tests__/Cards.spec.ts
@@ -6,9 +6,8 @@ import cardsReducer from 'reducers/cardsReducer';
 import {
   createSelectGroupArticles,
   createSelectSupportingArticles,
-  selectCard,
-  selectSharedState
-} from '../../selectors/shared';
+  selectCard
+} from 'selectors/shared';
 import { selectClipboard as innerClipboardSelector } from '../../selectors/frontsSelectors';
 import { createCardStateFromSpec, CardSpec, specToCard } from './utils';
 import {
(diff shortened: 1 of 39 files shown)


fronts-client/src/actions/__tests__/Cards.spec.ts
```diff
@@ -104,19 +104,15 @@ const insert = async (
...
-    afId => () =>
-      Promise.resolve(selectCard(selectSharedState(getState()), uuid))
+    afId => () => Promise.resolve(selectCard(stateHere, uuid))
   ) as any);
 
-  // TODO: use modal service to mock return from modal
-  // if (collectionCapInfo && collectionCapInfo.accept !== null) {
-  //   dispatch(endConfirmModal(collectionCapInfo.accept));
-  // }
-
```
ajwl: Removing this comment. This test is too complicated as it is, so any new modal test should  be in a separate file

fronts-client/src/actions/__tests__/Cards.spec.ts
```diff
@@ -104,19 +104,15 @@ const insert = async (
     [uuid, id, undefined],
     collectionCapInfo ? collectionCapInfo.cap : Infinity
   );
+
+  const stateHere: any = getState();
```
ajwl: Casting this as `any` allows the tests to run without throwing lots of type errors. We don't need exact type matching in the tests.

PR #1189, opened 2020-02-28
Consistent directory structure for Redux

ajwl:
## What's changed?
As part of the refactoring work to put this project into maintenance mode, we wanted to simplify a few areas of the codebase. 

Before this PR there were 3 different ways of storing Redux code in the file system. 
1) In directories called `actions`, `reducers` and `selectors`.
2) In the bundle style as [outlined here](https://reduxbundler.com/)
3) In the redux modules style

The redux module style definitely has its merits, and keeps feature code nice and encapsulated. But I chose to go with the simple `actions`, `reducers`, `selectors` directories at root level approach for one reason. 

==> Most of the code was done that way, so this was the quickest possible way to achieve consistency 

And the benefits of a consistent approach outweighs the benefits of any individual approach, as @akash1810 likes to say.* 

NB: the bundles method still exists - it would be more complicated to remove. But it should probably be documented.

*something like this.  

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fronts-client/src/reducers/rootReducer.ts
```diff
@@ -1,40 +1,40 @@
 import { reducer as form, FormStateMap } from 'redux-form';
```
AWare: Big can of worms here for the next refactor, but it'd be so nice to have consistent names so we don't need to rename on import.

ajwl:
oh cheers @AWare

fronts-client/src/reducers/rootReducer.ts
ajwl: Oh yeah - what importing this reducer directly from `redux-form`?

PR #1191, opened 2020-03-09
create a new endpoint for collection renaming

akash1810:
## What's changed?

In making renaming a collection a discrete action, we can avoid the scenario where:
- Person A renames container X from Y to Z
- Person B adds an item to container X
- On refresh, person A will see their rename reverted as person B pushed name Y even though they were not touching the container name

## Implementation notes

This is part of a series of PRs. Next in the series are:
- Update client to use this new endpoint
- Stop updating the name on the current endpoint (that is, only this endpoint can rename a collection)

### General
- [x] 🤖 Relevant tests added
- [x] ✅ CI checks / tests run locally
- [ ] 🔍 Checked on CODE

### Client
- [ ] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ ] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #1192, opened 2020-03-10
Renaming a collection calls the new rename route

akash1810:
## What's changed?

Following https://github.com/guardian/facia-tool/pull/1191 we now have a discrete route for updating a collection's name. This PR updates the client-side to use this new route.

## Implementation notes

The [new PATCH endpoint](https://github.com/guardian/facia-tool/pull/1191) has been updated to accept a payload of the same shape as the [`updateCollection`](https://github.com/guardian/facia-tool/blob/54e9dc6023ac7cd601ce54e2311d644acdcf37c8/app/controllers/EditionsController.scala#L128) route ([an `EditionsFrontendCollectionWrapper`](https://github.com/guardian/facia-tool/blob/a120a5382beb248bac2fb4b843ef38acaef79098/app/model/editions/EditionsClientCollection.scala#L39)).  This allows the changes on the client to be small - we just use a different endpoint. However it does mean the client is sending a lot of redundant information over the wire.

From a user's POV, this should be a no-op.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fronts-client/src/strategies/update-collection.ts
```diff
@@ -19,12 +22,19 @@ const collectionToEditionCollection = (
 const updateCollectionStrategy = (
   state: State,
   id: string,
-  collection: CollectionWithNestedArticles
+  collection: CollectionWithNestedArticles,
+  renamingCollection: boolean
```
jonathonherbert: This API is a little surprising to me – I was expecting a separate function! Perhaps it's something to address when we want to improve our update concurrency story more comprehensively.

akash1810: Yep, it's not the most obvious implementation. I was being super lazy. The main reason (though not especially obvious) is to make use of the actions post request. That is, I've kept the payload and response the same shape to reuse the sequence of actions we go through.

But mainly, I was being lazy...

jonathonherbert: @akash1810 I don't have a strong opinion here, and I think this PR is needed enough that we might be best off shipping this now and carding up this refactor as a health task – I'll leave that up to you 😁

PR #1201, opened 2020-03-19
Correct global topic name

jonathonherbert:
## What's changed?

#1198 had a mistake – by using `++` rather than `:+` in attempting to append a string to a list, it did the wrong thing. As a result, titles weren't present on global covid-19 notifications.

This PR corrects that mistake and adds a test to confirm.

Will be tested on CODE with normal and CV topics.

## Implementation notes

To enable easier testing, I've moved the relevant code – which is stateless – into a companion object.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/updates/BreakingNewsUpdate.scala
+++ b/app/updates/BreakingNewsUpdate.scala
@@ -24,6 +24,85 @@ import scala.util.{Failure, Success, Try}
 
 class InvalidNotificationContentType(msg: String) extends Throwable(msg) {}
 
+object BreakingNewsUpdate {
+  val CovidGlobalTopicName = "global-covid-19"
+  val CovidBreakingNewsTopics = List(
+    BreakingNewsCovid19Uk,
+    BreakingNewsCovid19Us,
+    BreakingNewsCovid19Au,
+    BreakingNewsCovid19International
+  )
+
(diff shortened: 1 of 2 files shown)


app/updates/BreakingNewsUpdate.scala
```diff
@@ -24,6 +24,85 @@ import scala.util.{Failure, Success, Try}
...
+    BreakingNewsCovid19Uk,
+    BreakingNewsCovid19Us,
+    BreakingNewsCovid19Au,
+    BreakingNewsCovid19International
+  )
+
+  def createPayload(trail: ClientHydratedTrail, email: String): BreakingNewsPayload = {
+    val title = trail.topic match {
+      case Some("uk-general-election") => Some("General election 2019")
+      case Some(topic) if (CovidBreakingNewsTopics.map(_.name) :+ CovidGlobalTopicName).contains(topic) => Some("Coronavirus")
```
jonathonherbert: `++` to `:+` is the only real change here – everything else is the move from class to object.

alexduf:
I haven't checked the code, but just so you know we've proved that it works on CODE with iOS and Android

jonathonherbert:
Thanks @alexduf 👍

PR #1204, opened 2020-03-26
Add state management for banner notification

jonathonherbert:
## What's changed?

We'd like to display banner notifications in the app. This adds state management for banner notifications, to be used by a component in a downstream PR.

## Implementation notes

I've added duplicate-management code, as there are many things that could display this banner, but we only want one of it. YAGNI?

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/bundles/notificationsBundle.ts
+++ b/fronts-client/src/bundles/notificationsBundle.ts
@@ -1,0 +1,50 @@
+import { Action } from 'types/Action';
+import set from 'lodash/fp/set';
+import { State } from 'types/State';
+
+interface BannerNotification {
+  message: string;
+}
+
+export interface NotificationState {
+  banner: BannerNotification | undefined;
+}
+
+// Actions
+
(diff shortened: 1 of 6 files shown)


fronts-client/src/bundles/notificationsBundle.ts
```diff
@@ -0,0 +1,82 @@
...
+        };
+      }
+
+      return {
+        banners: [
+          ...state.banners,
+          {
+            id: action.payload.id,
+            message: action.payload.message,
+            duplicates: 1
```
jennygrahamjones: If there is 1 duplicate, does that mean that only a single instance of the notification has been added?

jonathonherbert: Yes, the naming here isn't quite right. Perhaps `instances`?

jennygrahamjones: Perhaps stick with the name and default to `0` then. I think that would make sense!

jonathonherbert:
Closing as #1205 contained the changes already, and I mistakenly merged it first 🙄

PR #1207, opened 2020-03-26
Remove circular dependencies

jonathonherbert:
## What's changed?

This PR removes circular dependencies in the client-side app, and adds a check to ensure it's impossible to add new ones.

Circular dependencies can cause runtime errors that can't be detected by Typescript – they're a Javascript bundling issue. We can detect them at build time with a [webpack plugin](https://www.npmjs.com/package/circular-dependency-plugin).

There are many changes here, but all should boil down to either splitting code or adding `type` parameters to make it clear to compilers that no JS needs to be bundled. (This doesn't matter for `tsc`, Typescript's own compile, but [does matter for `babel`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html), which `jest` uses to run Typescript.

Before (reproduce with `npx madge --circular --extensions ts ./`)

```
✖ Found 64 circular dependencies!
```

After:

```
✔ No circular dependency found!
```

## Why have you done this? _why_

When trying to implement a notification for 419 errors – quite a basic change – I found that tests were failing unpredictably, and it was very difficult to find a working solution. Taking a look at the dependency tree, it was clear that we were in a tangle. thought it was better to deal with it up front.

## Implementation notes

I made a mistake in a previous PR in introducing Typescript 3.8 (required for this change) without introducing the new version of Prettier that supports it. So prettier comes with this PR, too, as it fixes the broken CI.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


fronts-client/src/actions/Config.ts
```diff
@@ -2,7 +2,7 @@ import { Config } from 'types/Config';
 
 const configReceived = (config: Config) => ({
   type: 'CONFIG_RECEIVED',
-  payload: config
+  payload: config,
```
jennygrahamjones: I've seen automated code formatting apply this change a few times and not really understood it. What is the benefit of ending with a comma here? Is there a particular lint/prettier rule that enforces this?

jonathonherbert: AFAIK Prettier has very few rules by design – instead it has opinions, and says, 'my way || highway' via reparsing the code, to [circumvent conversations about styling.](https://jlongster.com/A-Prettier-Formatter) (see 'Frictionless teams')

That said, I'm not sure why it has this particular opinion 😁

PR #1208, opened 2020-03-31
Notify users when auth fails

jonathonherbert:
## What's changed?

Notifies users when authentication fails. The number on the right hand side ticks up when things fail, with the reasoning – a) this deduplicates notifications which would otherwise be endless, and b) it provides some feedback that stateful operations aren't being saved as they happen.

auth

Let me know if the banner could be better worded!

## Implementation notes

We've added an event bus for notifications, as directly importing the store results in circular dependencies which are dangerous and break tests (see #1207 for details). This is the neatest way I and @akash1810 could think of to decouple the store from the code asking for notifications, but any feedback gratefully received.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/services/pandaFetch.ts
+++ b/fronts-client/src/services/pandaFetch.ts
@@ -2,7 +2,7 @@ import { reEstablishSession } from 'panda-session';
 import notifications from './notifications';
 
 const reauthUrl = '/login/status';
-const reauthErrorMessage = `We couldn't log you back in. Please log in again to reauthenticate.`;
+const reauthErrorMessage = `We couldn't log you back in. Your changes may not be saved. Please log in again to reauthenticate.`;
 
 /**
  * Make a fetch request with Panda authentication.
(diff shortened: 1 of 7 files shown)


fronts-client/src/services/pandaFetch.ts
```diff
@@ -1,6 +1,8 @@
 import { reEstablishSession } from 'panda-session';
+import notifications from './notifications';
 
 const reauthUrl = '/login/status';
+const reauthErrorMessage = `We couldn't log you back in. Your changes may not be saved. Please log in again to reauthenticate.`;
```
jennygrahamjones: Could we stick a link in here, so users can click through to log in again?

jonathonherbert: Yes, we can! Excellent idea, thanks for the suggestion @jennygrahamjones 👍

PR #1209, opened 2020-04-01
Move auth notification code to correct place

jonathonherbert:
## What's changed?

The auth notification code is in a place that will alert on any non-2XX response. This PR corrects that mistake.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/services/pandaFetch.ts
+++ b/fronts-client/src/services/pandaFetch.ts
@@ -22,16 +22,16 @@ const pandaFetch = (
         credentials: 'same-origin',
       });
 
-      if (res.status === 419 && count < 1) {
+      if ((res.status === 419 || res.status === 401) && count < 1) {
         try {
           await reEstablishSession(reauthUrl, 5000);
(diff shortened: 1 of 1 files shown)


fronts-client/src/services/pandaFetch.ts
```diff
@@ -22,16 +22,16 @@ const pandaFetch = (
         credentials: 'same-origin',
       });
 
-      if (res.status === 419 && count < 1) {
+      if ((res.status === 419 || res.status === 401) && count < 1) {
```
jonathonherbert, line 25: Our endpoints return `401` when authentication fails – I'm not sure why we used `419` here, [it's not an official HTTP status](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)

jonathonherbert: Ah, got it – panda issues that status when oauth has expired. So we catch 401s here from the app itself, and 419s from panda.

PR #1223, opened 2020-07-09
Begin adding metadata to templates

aug24:
## What's changed?

A single source of truth for metadata information about each Edition is required, so that we can generate information for other systems.

Currently this is hard coded in at least three places.

This work starts (there is more metadata to cover) placing metadata on templates and exposing it through an endpoint in JSON format.

https://trello.com/c/qFgFGr6y/265-parent-card-determine-where-new-editions-metadata-will-be-stored-and-how-they-will-be-sent-to-editions-app

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/EditionsController.scala
+++ b/app/controllers/EditionsController.scala
@@ -7,7 +7,8 @@ import com.gu.contentapi.json.CirceEncoders._
 import io.circe.syntax._
 import logging.Logging
 import logic.EditionsChecker
-import model.editions.{CapiTimeWindowConfigInDays, Edition, EditionsCollection, EditionsFrontMetadata, EditionsFrontendCollectionWrapper, EditionsTemplates}
+import model.editions._
+import model.editions.templates.EditionDefinition
 import model.forms._
 import net.logstash.logback.marker.Markers
 import play.api.Logger
@@ -176,7 +177,7 @@ class EditionsController(db: EditionsDB,
     db.getCollectionPrefill(id).map { prefillUpdate =>
       logger.info(s"getPrefillForCollection id=$id, prefillUpdate")
       import prefillUpdate._
-      val capiDateQueryParam = EditionsTemplates.templates(edition).capiDateQueryParam
+      val capiDateQueryParam = EditionsTemplates.templates(edition).template.capiDateQueryParam
       val capiPrefillTimeParams = CapiPrefillTimeParams(capiQueryTimeWindow, capiDateQueryParam)
       // TODO
       // when we click (suggest articles) for collection we are not using ophan metrics and we are not sorting on them
--- a/app/controllers/EditionsController.scala
+++ b/app/controllers/EditionsController.scala
@@ -8,7 +8,7 @@ import io.circe.syntax._
 import logging.Logging
 import logic.EditionsChecker
 import model.editions._
-import model.editions.templates.EditionDefinition
+import model.editions.templates.{EditionDefinition, EditionType}
 import model.forms._
 import net.logstash.logback.marker.Markers
 import play.api.Logger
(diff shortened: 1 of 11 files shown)


app/model/editions/templates/DailyEdition.scala
```diff
@@ -7,7 +7,14 @@ import model.editions._
 import model.editions.templates.TemplateHelpers._
 
 //noinspection TypeAnnotation
-object DailyEdition {
+object DailyEdition extends EditionDefinitionWithTemplate {
+  override val title = "The Daily"
+  override val subTitle = "Published every morning by 6am (GMT)"
+  override val edition = "daily-edition"
```
jonathonherbert, line 13: We already know this is an edition def. from the object name, but the name `edition` doesn't tell us much about the use of this field (although the string type gives you a hint) – for developers, is this a decorative `name`, or a uniquely referenced `id`? etc? Guessing the latter, in which case would `editionId` or just `id` be preferable?

app/model/editions/templates/AmericanEdition.scala
```diff
@@ -7,7 +7,15 @@ import model.editions._
 import model.editions.templates.TemplateHelpers._
 
 //noinspection TypeAnnotation
-object AmericanEdition {
+object AmericanEdition extends EditionDefinitionWithTemplate {
+
+  override val title = "US Weekender"
```
jonathonherbert, line 12: OMG spoilers 😁

app/model/editions/templates/DailyEdition.scala
aug24: It's the 'priority' or part of the 'key'... I don't quite know what to call it.

aug24:
> LGTM, one question about variable naming – I haven't seen this trait/object/case class pattern before, interesting!

It's only there to make apply/unapply work for json.

app/controllers/EditionsController.scala
```diff
@@ -85,9 +86,19 @@ class EditionsController(db: EditionsDB,
...
 
-  def getAvailableEditions = EditEditionsAuthAction { _ =>
-    Ok(Json.toJson(EditionsTemplates.getAvailableEditions))
-  }
+  def getAvailableEditions = EditEditionsAuthAction { _ => {
+    val allEditions = EditionsTemplates.getAvailableEditions
+    val regionalEditions = allEditions.filter(e => e.editionType == EditionType.Regional)
+    val specialEditions = allEditions.filter(e => e.editionType == EditionType.Special)
+    val trainingEditions = allEditions.filter(e => e.editionType == EditionType.Training)
+    val editions = Map(
```
akash1810, line 94: minor: would a `groupBy` here be more future proof?

app/model/editions/templates/TemplateHelpers.scala
```diff
@@ -39,3 +40,58 @@ object TemplateHelpers {
...
+case class Header (title: String, subTitle: Option[String] = None)
+object Header {
+  implicit val readsHeader: Reads[Header] = Json.reads[Header]
+  implicit val writesHeader: Writes[Header] = Json.writes[Header]
+}
+
+trait EditionDefinition {
+  val title: String
+  val subTitle: String
+  val edition: String
```
akash1810, line 61: Is there a possibility to make this an enum for a stricter type?

app/model/editions/templates/TemplateHelpers.scala
```diff
@@ -39,3 +40,58 @@ object TemplateHelpers {
...
+object EditionDefinition {
+  def apply(
+    title: String,
+    subTitle: String,
+    edition: String,
+    header: Header,
+    editionType: EditionType
+  ): EditionDefinition = EditionDefinitionRecord(title, subTitle, edition, header, editionType)
+
+  def unapply(x: EditionDefinition): Option[(String, String, String, Header, EditionType)]
```
akash1810, line 77: minor: could we use a more explicit variable name here?

app/model/editions/templates/TemplateHelpers.scala
```diff
@@ -39,3 +40,58 @@ object TemplateHelpers {
...
+    = Some(x.title, x.subTitle, x.edition, x.header, x.editionType)
+
+  implicit val formatEditionDefinition: OFormat[EditionDefinition] = Json.format[EditionDefinition]
+  implicit val writesEditionDefinition: OWrites[EditionDefinition] = Json.writes[EditionDefinition]
+  implicit val readsEditionDefinition: Reads[EditionDefinition] = Json.reads[EditionDefinition]
```
akash1810, line 82: Are reads and writes needed if we specify a format?

app/controllers/EditionsController.scala
aug24: It ought to be, but the strings (eg regionalEditions) need to match the existing values, and it doesn't actually end up any clearer.

app/model/editions/templates/TemplateHelpers.scala
aug24: I don't think so: the string is defined by the template, rather than the template having to match a list stored somewhere else.

app/model/editions/templates/TemplateHelpers.scala
aug24: Ah... apparently not.  Done.

app/model/editions/templates/TemplateHelpers.scala
aug24: Done.

PR #1224, opened 2020-07-13
Add notification offset in UTC for each edition

aug24:
## What's changed?

Each Edition definition now has a UTC offset.  This will be used, relative to the edition date in UTC, to decide when the download notification is sent.

## Implementation notes
This is the first part of the process: the offset has been added to the definition and exposed in the published issue json format.  It will be ignored by the step function.  The next step is to make the step function pass this value through to the notification service.

### General
- [ X] 🤖 Relevant tests added
- [ X] ✅ CI checks / tests run locally
- [ X] 🔍 Checked on CODE

### Client
- [ N/A] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ N/A] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ X] 📷 Screenshots / GIFs of relevant UI changes included

This is the file content now exposed via the published-editions-code s3 bucket.  The notification value is currently ignored.

image

### DIFF
--- a/app/model/editions/EditionsIssue.scala
+++ b/app/model/editions/EditionsIssue.scala
@@ -32,6 +32,11 @@ case class EditionsIssue(
     fronts: List[EditionsFront]
 ) {
 
+  // This is a no-op placeholder which matches the UK Daily Edition value.
+  // It should never be needed because the Editions Templates object should always
+  // contain a matching Edition.  It should (TODO) eventually go away.
+  private val defaultOffset = 3
+
   def toPreviewIssue: PublishableIssue = toPublishableIssue("preview", PublishAction.preview)
 
   def toPublishableIssue(version: String, action: PublishAction): PublishableIssue = PublishableIssue(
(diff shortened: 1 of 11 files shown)


app/model/editions/EditionsIssue.scala
```diff
@@ -48,7 +48,8 @@ case class EditionsIssue(
     else fronts
       .filterNot(_.isHidden) // drop hidden fronts
       .map(_.toPublishedFront) // convert
-      .filterNot(_.collections.isEmpty) // drop fronts that contain no collections
+      .filterNot(_.collections.isEmpty), // drop fronts that contain no collections
+    EditionsTemplates.templates.get(edition).map(_.notificationUTCOffset).getOrElse(3)
```
jennygrahamjones, line 52: Could we add a comment to clarify why `3` or use a constant/`val` declared elsewhere (perhaps `DailyEdition.notificationUTCOffset`) for this?

aug24: Yes, quite right, will do.  Magic numbers bad.

PR #1243, opened 2020-08-25
Some mods to the AU edition

blishen:
## What's changed?
Renamed some containers.
Created prefills and additional containers for Culture.
Removed one container from Recommended reads.

## Implementation notes
Edited in Github, beware my linting!


app/model/editions/templates/AustralianEdition.scala
```diff
@@ -56,9 +56,9 @@ object AustralianEdition extends EditionDefinitionWithTemplate {
 
   def FrontTopStoriesAu = front(
     "Top stories",
```
aug24, line 58:
```suggestion
    "Top Stories",
```

aug24 (approved):
See suggestion?

app/model/editions/templates/AustralianEdition.scala
blishen: Style is lower case on stories. But thanks!

PR #1247, opened 2020-09-02
Capi demo

aug24:
## What's changed?

I hate the capi filters
No, hang on, that hasn't changed.
You can now avoid the capi filters.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/editions/templates/Capi.scala
+++ b/app/model/editions/templates/Capi.scala
@@ -48,7 +48,18 @@ object Capi{
   def businessaustralia=Capi("australia-news/business-australia")
   def australiamedia=Capi("media/australia-media")
 
+  def sport=Capi("sport/sport")
   def culture=Capi("culture/culture")
   def lifestyle=Capi("lifestyle/lifestyle")
   def comment=Capi("tone/comment")
+
+  def world = Capi("world/world")
+  def usnews = Capi("|us-news/us-news")
+  def uk = Capi("uk/uk")
+  def europenews = Capi("world/europe-news")
+  def africa=Capi("world/africa")
+  def americas=Capi("world/americas")
+  def asiapacific=Capi("world/asia-pacific")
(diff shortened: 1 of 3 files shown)


app/model/editions/templates/Capi.scala
```diff
@@ -0,0 +1,65 @@
...
+  def build(c:Capi) = this.copy(next = Some(c))
+}
+
+object Capi{
+  def apply() = new Capi("?tag=")
+  def apply(s: String) = new Capi(s)
+  def article=Capi("type/article")
+  def lifeandstyle=Capi("lifeandstyle/lifeandstyle")
+
+  def features=Capi("tone/features")
```
juanvaccari, line 33: I was wondering how do you reify the tags, or sections, etc. I see you are mapping them manually. But there are more than 50,000 tags. And they are mutating all the time, I guess? At least more are added frequently. So how would you handle that?

juanvaccari: I must say I really like it, it is very elegant and safe to be able to write queries like this. I am just not sure about this practical issue.

aug24: Whoever is writing the client will define the tags they want.  This section of code would not be in a real client.

aug24: The `import model.editions.templates.Capi._` import at the top where the items are used is all that matters.

juanvaccari: OK, so the client needs to define at compile time which tags they will use, which will work well if they use a limited and known set of them, for what I understand.

aug24: You could easily make it dynamic.  Parse values out, wrap them into Capi class instances and include them.  That's what functional languages do best.

PR #1250, opened 2020-09-04
Am add locale field to editions template

aoifemcl15:
## What's changed?

In preparation to remove the US edition, we thought that the app shouldn't be dependant on the existence of any particular edition. 
This PR adds a new `locale` field which the editions app can use to decide which default edition to select for a new user based on their selected locale.
**Note** The locale field is only relevant for Regional Editions. 

I have deployed this branch to CODE to check that the field comes through as expected https://fronts.code.dev-gutools.co.uk/editions-api/editions

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/editions/templates/TheDummyEdition.scala
+++ b/app/model/editions/templates/TheDummyEdition.scala
@@ -14,6 +14,7 @@ object TheDummyEdition extends InternalEdition {
   override val header = Header("The Dummy", Some("Edition"))
   override val notificationUTCOffset = 3
   override val topic = "dm"
+  override val locale = Some("dm_DM")
 
   lazy val template = EditionTemplate(
     List(
--- a/app/model/editions/templates/TheDummyEdition.scala
+++ b/app/model/editions/templates/TheDummyEdition.scala
@@ -14,7 +14,7 @@ object TheDummyEdition extends InternalEdition {
   override val header = Header("The Dummy", Some("Edition"))
   override val notificationUTCOffset = 3
   override val topic = "dm"
-  override val locale = Some("dm_DM")
+  override val locale = Some("en_GB")
 
   lazy val template = EditionTemplate(
     List(
--- a/app/model/editions/templates/TrainingEdition.scala
+++ b/app/model/editions/templates/TrainingEdition.scala
@@ -14,7 +14,7 @@ object TrainingEdition extends InternalEdition {
   override val header = Header("Training Edition")
   override val notificationUTCOffset = 3
   override val topic = "tr"
-  override val locale = Some("tr_TR")
+  override val locale = Some("en_GB")
 
   lazy val template = EditionTemplate(
     List(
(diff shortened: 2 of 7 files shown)


app/model/editions/templates/TheDummyEdition.scala
```diff
@@ -14,6 +14,7 @@ object TheDummyEdition extends InternalEdition {
   override val header = Header("The Dummy", Some("Edition"))
   override val notificationUTCOffset = 3
   override val topic = "dm"
+  override val locale = Some("dm_DM")
```
aug24, line 17: I'm concerned this might prevent us viewing the Dummy Edition on devices.  May I suggest en_GB instead?

app/model/editions/templates/TrainingEdition.scala
```diff
@@ -14,6 +14,7 @@ object TrainingEdition extends InternalEdition {
   override val header = Header("Training Edition")
   override val notificationUTCOffset = 3
   override val topic = "tr"
+  override val locale = Some("tr_TR")
```
aug24, line 17: I'm concerned this might prevent us viewing the Training Edition on devices.  May I suggest en_GB instead?

test/fixtures/TestEdition.scala
```diff
@@ -41,6 +41,7 @@ object TestEdition extends RegionalEdition {
   override val editionType: EditionType = EditionType.Regional
   override val notificationUTCOffset: Int = 2
   override val topic: String = "tt"
+  override val locale: Option[String] = Some("tt_TT")
```
aug24, line 44: I doubt if we will ever want the Test Edition, but this isn't a valid Locale.  I'd suggest using some valid one.

Do the tests correctly check this value?

aug24 (changes requested):

jennygrahamjones:
I don't have a lot of context for this, but if it's something that's only relevant to editions that extend `RegionalEdition`, is it possible to make this field part of that trait rather than the base `EditionDefinition`?

aoifemcl15:
> I don't have a lot of context for this, but if it's something that's only relevant to editions that extend `RegionalEdition`, is it possible to make this field part of that trait rather than the base `EditionDefinition`?

Sounds like a good idea, however, my scala is very limited so I'm not sure how I would go about adding the `locale` to the `RegionalEdition` since it is currently defined as an abstract class instead of a trait, any guidance would be appreciated 🙂

jennygrahamjones:
> > I don't have a lot of context for this, but if it's something that's only relevant to editions that extend `RegionalEdition`, is it possible to make this field part of that trait rather than the base `EditionDefinition`?
> 
> Sounds like a good idea, however, my scala is very limited so I'm not sure how I would go about adding the `locale` to the `RegionalEdition` since it is currently defined as an abstract class instead of a trait, any guidance would be appreciated 🙂

Reading @aug24's comments, it looks ike we would need these value on the `InternalEdition` implementations too, so probably not worth worrying about at the moment 👍

aug24 (approved):

PR #1239, opened 2020-08-18
Created a simple "special edition" for testing

blishen:
## What's changed?
Added a special edition template
Identified it to the app in `app/model/editions/EditionsTemplates.scala`

## Implementation notes
All edited in Github by a product manager, Justin better give this a thumbs up or ideally test in CODE?

PR #1255, opened 2020-09-16
Special edition improvements

philmcmahon:
## What's changed?
 - Replaced special edition `image` field with `buttonImageUri` following update in the editions app
 - Introduced new `EditionBase` class for properties that are common between Internal and Regional Editions
 - Updated docs for creating a new edition to warn people that the class name needs to match the edition ID
 - Introduce new `isLive` field for editions - currently ignored by the client - to be used to filter out special editions or regional editions that we need to test in the app with live issues but that we don't want users to see
 - Merge in code changes for special edition @blishen created here https://github.com/guardian/facia-tool/pull/1239 but without actually adding it to the editions list, just so that we have some code in master that won't get into conflicts (once we have introduced `isLive` in the client we could safely merge the test edition)

## Implementation notes
Hopefully it's ok to merge in the test special edition - it's got a bit tedious refactoring it all the time. This way each time we deploy to code it'll be just a 2 line change to add the special edition to the list in `EditionTemplates`

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/editions/templates/TemplateHelpers.scala
+++ b/app/model/editions/templates/TemplateHelpers.scala
@@ -54,11 +54,6 @@ object Header {
   implicit val formatHeader: OFormat[Header] = Json.format[Header]
 }
 
-case class SpecialEditionImage(source: String, path: String)
-object SpecialEditionImage {
-  implicit val formatHeader: OFormat[SpecialEditionImage] = Json.format[SpecialEditionImage]
-}
-
 case class SpecialEditionHeaderStyles(backgroundColor: String, textColorPrimary: String, textColorSecondary: String)
 object SpecialEditionHeaderStyles {
   implicit val formatHeader: OFormat[SpecialEditionHeaderStyles] = Json.format[SpecialEditionHeaderStyles]
@@ -94,7 +89,7 @@ trait EditionDefinition {
   val editionType: EditionType
   val notificationUTCOffset: Int
   val topic: String
-  val image: Option[SpecialEditionImage]
+  val buttonImageUri: Option[String]
   val expiry: Option[String]
   val buttonStyle: Option[SpecialEditionButtonStyles]
   val headerStyle: Option[SpecialEditionHeaderStyles]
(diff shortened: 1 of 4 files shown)


philmcmahon:
Just to add- @aug24 already reviewed most of these changes in https://github.com/guardian/facia-tool/pull/1239 - the only new thing is the new `EditionBase` class and the `isLive` flag

app/model/editions/templates/TemplateHelpers.scala
```diff
@@ -94,30 +89,31 @@ trait EditionDefinition {
   val editionType: EditionType
   val notificationUTCOffset: Int
   val topic: String
-  val image: Option[SpecialEditionImage]
+  val buttonImageUri: Option[String]
   val expiry: Option[String]
   val buttonStyle: Option[SpecialEditionButtonStyles]
   val headerStyle: Option[SpecialEditionHeaderStyles]
+  val isLive: Boolean
```
aug24, line 96: Is this needed?  Surely all Special and Regional Editions are live?  That's why we introduced Training: so that we could have test issues right through to the app.

docs/AddingAndAlteringTemplates.md
```diff
@@ -26,6 +26,9 @@ object MyLovelyHorseEdition extends EditionDefinitionWithTemplate {
   ...
 ```
 
+NOTE: the `edition` property of your object must be a hypen case version of your object name otherwise the routing won't
```
aug24, line 29:
```suggestion
NOTE: the `edition` property of your object must be a hyphen case version of your object name otherwise the routing won't
```

aug24: And also... what?  Why must it match?

app/model/editions/templates/TemplateHelpers.scala
philmcmahon: Yes I think it is needed at the moment- otherwise we need to fully support training editions in the app UI which would be a load of extra work. I know yesterday we agreed that rather than doing something like this we would just use Preview/Proof for testing, but at least at the moment I think there's a real use case (as in, we need it on the team right now) for having editions which exist in PROD and multiple issues available, but are hidden to users.

docs/AddingAndAlteringTemplates.md
philmcmahon: It's your code! This caught me out when I initially tried to test this - I guess the dynamic routing thing defaults a route for ClassName to class-name

app/model/editions/templates/TemplateHelpers.scala
philmcmahon: Discussed in person - we decided `isLive` wasn't needed

aug24 (approved):

PR #1259, opened 2020-09-22
Edition Earth!

philmcmahon:
## What's changed?

This introduces the initial template for a new special edition - _Edition Earth_! The template is still in flux - some of the colours and fonts will need to change - but it would be great t- get this into PROD so that Katy V can start assembling an edition - my understanding is that so long as the slug and class name don't change we're safe to change the other properties of the edition.

Screenshot 2020-09-22 at 16 34 20

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


jonathonherbert:
🌎 🌍 🌏

aug24 (approved):
Exciting!

PR #1272, opened 2020-11-27
Housekeeping

twrichards:
## What's changed?
- clarify credentials part of README (i.e. which accounts needed, if not clicking the link which does it for ya).
- increase SBT stack space to prevent intermittent overflow errors during compilation


jonathonherbert (approved):
Thanks @twrichards, looks good 👍

PR #1277, opened 2020-12-02
fix stale writes by locking collections when auth/connection issue discovered

twrichards:
https://trello.com/c/B3dnmNti/2094-its-possible-to-publish-a-stale-collection-in-the-fronts-tool

## What's changed?
As per https://docs.google.com/document/d/14vbdereyeYU8MW6VDdUA_Pbbnv8_tf9ZJciKZMqpd0w/edit# it was possible for stale writes to occur in fronts, due to a combination of the 10second polling and people having auth/connection issues.

This PR seeks to address that by locking the collections when the latest fetch for collections is either errored or is stale by 20 seconds. By locking only the collections we allow users to potentially continue other productive work in the tool such as racking items up in the clipboard.

Before being able to do this there was a bit of re-work required, most notably...

- `FETCH_SUCCESS` events were only fired for collections when the data actually changed, and as such the `lastFetch` and `error` fields were not updated until the next material change to the collections (made by someone else), as such without changing, this the new locking feature would have displayed for much longer than necessary.

- Previously in `pandaFetch` any network failures would've thrown errors which were not caught in the [somewhat convoluted] async/await/Promise construct, and came out like this (Offline is forced on here, notice the `Uncaught` line)....
image
... they're now caught and calls the `reject` of the outer promise (this in turn populates the `error` field under `collections` in the state).

## GIF
fronts_stale_fix

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/actions/Collections.ts
+++ b/fronts-client/src/actions/Collections.ts
@@ -238,10 +238,20 @@ function getCollections(
         getCollectionActions(collectionResponse, getState)
       );
 
+      const actionsToBatch = flatten([
+        ...actions,
+        ...missingActions,
+        ...missingCollectionActions,
+      ]);
+
+      // this is necessary to ensure lastFetch, error etc. are always updated
+      // (as we rely on them elsewhere to lock editing if stale/errored)
+      const noChangesAction = collectionActions.fetchSuccess([]);
+
       dispatch(
-        batchActions(
-          flatten([...actions, ...missingActions, ...missingCollectionActions])
-        )
+        actionsToBatch.length > 0
+          ? batchActions(actionsToBatch)
+          : noChangesAction
       );
       return collectionResponses.map(({ id }) => id);
     } catch (error) {
--- a/fronts-client/src/components/FrontsEdit/FrontContent.tsx
+++ b/fronts-client/src/components/FrontsEdit/FrontContent.tsx
@@ -21,7 +21,7 @@ import { createSelectAlsoOnFronts } from 'selectors/frontsSelectors';
 import { AlsoOnDetail } from 'types/Collection';
 import { selectors as collectionSelectors } from 'bundles/collectionsBundle';
 
-const TWENTY_SECONDS_IN_MILLIS = 20_000;
+const STALENESS_THRESHOLD_IN_MILLIS = 30_000; // 30 seconds
 
 const CollectionContainer = styled.div`
   position: relative;
@@ -88,11 +88,13 @@ type FrontProps = FrontPropsBeforeState & {
 
 interface FrontState {
   currentlyScrolledCollectionId: string | undefined;
+  isCollectionsStale: boolean | undefined;
 }
 
 class FrontContent extends React.Component<FrontProps, FrontState> {
   public state = {
     currentlyScrolledCollectionId: undefined,
+    isCollectionsStale: undefined,
   };
   private collectionElements: {
     [collectionId: string]: HTMLDivElement | null;
@@ -143,6 +145,13 @@ class FrontContent extends React.Component<FrontProps, FrontState> {
         this.props.browsingStage
       );
     }
+    if (this.props.collectionsLastFetch !== newProps.collectionsLastFetch) {
+      this.updateCollectionsStalenessFlag();
+      setTimeout(
+        this.updateCollectionsStalenessFlag,
+        STALENESS_THRESHOLD_IN_MILLIS
+      );
+    }
   }
 
   public componentDidMount() {
@@ -166,11 +175,7 @@ class FrontContent extends React.Component<FrontProps, FrontState> {
   public render() {
     const { front, collectionsError } = this.props;
 
-    const isCollectionsStale =
-      this.props.collectionsLastFetch &&
-      Date.now() - this.props.collectionsLastFetch > TWENTY_SECONDS_IN_MILLIS;
-
-    const isEditingLocked = isCollectionsStale || !!collectionsError;
+    const isEditingLocked = this.state.isCollectionsStale || !!collectionsError;
 
     return (
       <FrontCollectionsContainer
@@ -227,6 +232,15 @@ class FrontContent extends React.Component<FrontProps, FrontState> {
       </FrontCollectionsContainer>
     );
   }
+
+  private updateCollectionsStalenessFlag = () => {
+    const isCollectionsStale =
+      !!this.props.collectionsLastFetch &&
+      Date.now() - this.props.collectionsLastFetch >
+        STALENESS_THRESHOLD_IN_MILLIS;
+
+    this.setState({ isCollectionsStale });
+  };
 }
 
 const mapStateToProps = () => {
--- a/fronts-client/src/components/FrontsEdit/FrontContent.tsx
+++ b/fronts-client/src/components/FrontsEdit/FrontContent.tsx
@@ -19,6 +19,9 @@ import { editorSelectCard } from 'bundles/frontsUI';
 import { initialiseCollectionsForFront } from 'actions/Collections';
 import { createSelectAlsoOnFronts } from 'selectors/frontsSelectors';
 import { AlsoOnDetail } from 'types/Collection';
+import { selectors as collectionSelectors } from 'bundles/collectionsBundle';
+
+const TWENTY_SECONDS_IN_MILLIS = 20_000;
 
 const CollectionContainer = styled.div`
   position: relative;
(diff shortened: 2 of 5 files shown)


fronts-client/src/lib/createAsyncResourceBundle/index.ts
```diff
@@ -389,9 +389,6 @@ function createAsyncResourceBundle<Resource>(
           ) {
             return state;
           }
-          if (!action.payload.error) {
-            return state;
-          }
```
twrichards, line 394: based on the `if` above this appeared redundant to me

fronts-client/src/services/pandaFetch.ts
```diff
@@ -16,26 +19,42 @@ const pandaFetch = (
...
-      return resolve(res);
+        return resolve(res);
+      }
+      catch (error) {
+        notifications.notify({
+          message: "Connection issue occurred. Your changes may not be saved. Please wait or reload the page.",
+          level: 'error'
+        });
+        return reject(`Connection Issue ${JSON.stringify(error || {})}`);
+      }
```
twrichards, line 57: without this new outer `try`/`catch` we would get badness in the console...

fronts-client/src/services/pandaFetch.ts
```diff
@@ -16,26 +18,43 @@ const pandaFetch = (
...
-      return resolve(res);
+        return resolve(res);
+      } catch (error) {
+        notifications.notify({
+          message:
+            'Connection issue occurred. Your changes may not be saved. Please wait or reload the page.',
+          level: 'error',
+        });
+        return reject(`Connection Issue ${JSON.stringify(error || {})}`);
+      }
```
twrichards, line 57: without this new outer `try`/`catch` we would get badness in the console...

fronts-client/src/components/FrontsEdit/FrontContent.tsx
```diff
@@ -143,6 +164,15 @@ class FrontContent extends React.Component<FrontProps, FrontState> {
 
   public render() {
     const { front } = this.props;
+
+    const isCollectionsStale =
+      this.props.collectionsLastFetch &&
+      Date.now() - this.props.collectionsLastFetch > TWENTY_SECONDS_IN_MILLIS;
```
twrichards, line 170: Whilst I'm not sure this is actually necessary in that we can check the error directly, I doubt it actually works as desired, as the only time its evaluated is when some thing causes it to re-render (which is likely fresh changes - at which point the lastFetch will be v. recent)... instead perhaps we use a `setTimeout` (with 20 second delay) to write some new field to state (something like `lastStaleCheck`) to force this to be re-evaluated (there might be a nice React way of doing that)... thoughts welcome....

twrichards: DONE in https://github.com/guardian/facia-tool/pull/1277/commits/da957b710fb4802b478d304481817b5acac588d4

fronts-client/src/components/FrontsEdit/FrontContent.tsx
```diff
@@ -20,6 +20,8 @@ import { initialiseCollectionsForFront } from 'actions/Collections';
 import { createSelectAlsoOnFronts } from 'selectors/frontsSelectors';
 import { AlsoOnDetail } from 'types/Collection';
 
+const TWENTY_SECONDS_IN_MILLIS = 20_000;
```
jonathonherbert, line 23: TIL, numeric separators! Handy!

fronts-client/src/actions/Collections.ts
```diff
@@ -238,10 +238,18 @@ function getCollections(
         getCollectionActions(collectionResponse, getState)
       );
 
+      const actionsToBatch = flatten([
+        ...actions,
+        ...missingActions,
+        ...missingCollectionActions,
+      ]);
+
+      const noChangesAction = collectionActions.fetchSuccess([]);
```
jonathonherbert, line 247: Might be worth leaving a comment here to indicate why we're dispatching this. Other devs may consider it and remove, introducing a bug – and there's no test to catch it AFAICS (not necessarily saying add a test, it may be time consuming!)

fronts-client/src/services/pandaFetch.ts
```diff
@@ -16,26 +19,45 @@ const pandaFetch = (
...
+          return reject(res);
         }
-      } else if (res.status < 200 || res.status >= 300) {
-        return reject(res);
-      }
 
-      return resolve(res);
+        return resolve(res);
+      } catch (error) {
+        notifications.notify({
```
jonathonherbert, line 54: This change is a Good Thing, but we've no way at present to assess its impact on users.

I'd recommend capturing this failure in Raven (it's got retry strats, this isn't as paradoxical as it sounds 😁 ) so we can see if users are experiencing high volumes of these messages, and diagnose as appropriate.

There's an example in `move.ts`, e.g. `Raven.captureMessage(`Moving item up in clipboard failed: ${e.message}`);` (but let's not chuck away the stack trace!)

In any case, great catch. Pun not intended.

fronts-client/src/components/FrontsEdit/FrontContent.tsx
```diff
@@ -197,6 +248,8 @@ const mapStateToProps = () => {
     return {
       front: selectFront(state, { frontId: id }),
       alsoOn: selectAlsoOnFronts(state, { frontId: id }),
+      collectionsError: state.collections.error,
+      collectionsLastFetch: state.collections.lastFetch,
```
jonathonherbert, line 252: We shouldn't access redux state directly. Instead, there are selectors in `asyncResourceBundle` for both of these pieces of state.

jonathonherbert:
This is great. Left a few recommendations.

I'm a little uncomfortable signposting in the overlay to a notification that the user can dismiss out of hand. Is there a reason we can't include the reason that the collection is locked in the overlay itself? E.g. `We couldn't refresh this collection. It may be out of date.` – but! don't use this messaging! Talk to CP 😁

fronts-client/src/services/pandaFetch.ts
twrichards: happy to add the call to Sentry, are you recommending that instead of a user banner?

fronts-client/src/services/pandaFetch.ts
```diff
@@ -16,26 +19,45 @@ const pandaFetch = (
...
+          message:
+            'Connection issue occurred. Your changes may not be saved. Please wait or reload the page.',
+          level: 'error',
+        });
```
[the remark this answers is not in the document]
jonathonherbert: I don't disagree – I just think it's important that we understand the impact.

fronts-client/src/actions/Collections.ts
twrichards: added a comment for now, will look into adding a test too

fronts-client/src/components/FrontsEdit/FrontContent.tsx
twrichards: Then DONE better in https://github.com/guardian/facia-tool/pull/1277/commits/3e0fa48b3f54b82e4f4a4d082cab76d799ccac1b

fronts-client/src/services/pandaFetch.ts
twrichards: added in addition to the banners (as discussed over chat) - see https://github.com/guardian/facia-tool/pull/1277/commits/5884f19903acdf639dff6849012a242567f8e32d

jonathonherbert (approved):
Tested locally, works as expected – any connection errors lock collection editing, latent connections have a tolerance window of 30s before we lock collection editing.

Minor: recommend a little margin and possible centre-justifying text to match header, as the text abuts the collection bounds on smaller screen.

Screenshot 2020-12-07 at 09 54 47

This PR goes a little further than I originally thought by locking collections immediately when there are errors, rather than waiting for collections to be stale regardless of network conditions. I'm in two minds about this, as we don't know how often connection problems occur, and whether for users on poor connections we're going to get e.g. intermittent flickers etc.

Let's pay attention to what Sentry picks up to understand how this affects users. I'd be tempted to display connection errors but avoid locking collections immediately, instead waiting for them to become stale before preventing edits – but I'm happy to approve as long as we've got a way of measuring the impact, if only by proxy.

twrichards:
> Minor: recommend a little margin and possible centre-justifying text to match header, as the text abuts the collection bounds on smaller screen.
> 
> Screenshot 2020-12-07 at 09 54 47

Good spot, thanks @jonathonherbert (wide monitor catching me out again). Done in https://github.com/guardian/facia-tool/pull/1277/commits/d108e9336cefdecbeac27b78f535b031911f4b34
image

PR #1278, opened 2020-12-02
Added first published time to clipboard items

jennygrahamjones:
Adds the time elapsed since a story was first published to clipboard items. This is already present on stories in the CAPI feed and stories that have been added into a container. A user suggested it would be helpful to have this on clipboard items too!

After some discussion in the comments below, we've come up with the following design:

Screenshot 2020-12-04 at 06 47 27

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/article/ArticleBody.tsx
+++ b/fronts-client/src/components/article/ArticleBody.tsx
@@ -77,6 +77,11 @@ const VideoIconContainer = styled(CircularIconContainer)`
   right: 2px;
 `;
 
+const ClipboardFirstPublished = styled.div`
+  font-size: 11px;
+  margin: 4px;
+`;
+
 interface ArticleBodyProps {
   newspaperPageNumber?: number;
 
@@ -263,6 +268,14 @@ const articleBodyDefault = React.memo(
             </CardHeading>
             {displayByline && <ArticleBodyByline>{byline}</ArticleBodyByline>}
           </CardHeadingContainer>
+          {!collectionId && firstPublicationDate && (
+            <ClipboardFirstPublished title="The time elapsed since this article was first published.">
+              {distanceInWordsStrict(
+                Date.now(),
+                new Date(firstPublicationDate)
+              )}
+            </ClipboardFirstPublished>
+          )}
         </CardContent>
         <ImageAndGraphWrapper size={size}>
           {featureFlagPageViewData && canShowPageViewData && collectionId && (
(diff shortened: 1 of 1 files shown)


paperboyo:
Great! I think the tooltip should be enough, but good point about it being sadly obscured. See this [internal discussion](https://chat.google.com/u/1/room/AAAAXBA-uZk/oK_CgmLdG1E) for some ideas and [this one](https://chat.google.com/u/1/room/AAAAXBA-uZk/H2EGs8_1sFE) (which suggests it might have gotten worse – I cannot force any tooltips to show?) cc @pradasa @zeek01 for a possible opinion for another PR?

jonathonherbert (approved):
Great work @jennygrahamjones! Only one thought, please ignore –  perhaps the time could be below the picture, to give us a little extra real estate (articles with long headlines take up a lot of space as it is)? Aware of the inconsistency w/ feed, mind (although there are many more!)

SHession (approved):
This looks great! I was wondering, is there a particular reason why it shows the relative time difference and not simply a date? Are clipboard items ordered by published date (I'm not familiar with fronts)?

jennygrahamjones:
@jonathonherbert I did think about that! I'm not sure how tied we are to our current design patterns. Here's how it might look if we went with displaying under the thumbnail:

Screenshot 2020-12-03 at 06 24 33

Since we're in the clipboard, we shouldn't ever have any other metadata - like the Ophan sparklines - under the thumbnail, so it feels fairly safe to put it there.

@SHession you can add items into the clipboard in any position you like, so they won't necessarily be in any order. In terms of the way we display the date, this change follows the pattern we use in the rest of the tool when displaying when an article was first published. I'm not sure if there was a specific reason we went with the current pattern. @jonathonherbert may have more insight into that.

jonathonherbert:
@SHession great question – presume it's because fronts editors, esp. on fast moving desks, routinely deal with very recent content, and it's easier to reason about it by staleness than by date. But! this is most likely in this format because it's the way the old tool displayed dates 😁 

@jennygrahamjones ah, nice, I think we do use less space there for long headlines, so perhaps that's a bit better – only other thought is that perhaps right justifying the text will move it a little further from the headline text.

jennygrahamjones:
What do you think @jonathonherbert ?

Screenshot 2020-12-04 at 06 10 29

paperboyo:
Designer in me would move it to the bottom (but, indeed, it's better under the pic). But that designer been dead for 20 years, so...

jennygrahamjones:
Screenshot 2020-12-04 at 06 47 27

Looks good, I think?

jonathonherbert:
@jennygrahamjones 👍 ✨

PR #1279, opened 2020-12-08
Revert "fix stale writes by locking collections when auth/connection issue discovered"

twrichards:
Reverting as this was locking users editing intermittently. I wonder if the page doesn't have focus the collections aren't fetched?! @jonathonherbert


jennygrahamjones (approved):

jonathonherbert:
@twrichards yes, this is likely, and not something I'd considered 🤦 v. happy to pair on a fix.

PR #1280, opened 2020-12-09
track stale writes etc. in Sentry

twrichards:
Reinstates most of #1277 (after reverting in #1279 - due to users complaining).

Can't actually replicate the 'false positives' myself, which suggests to me they are true positives and the user was actually seeing stale collections and should probably be prevented from make changes that could lead onto a stale write...

... but rather than bluntly reinstating as it was, the changes that actually locked editing (with the overlay) as well as the additional banners are commented out, so we can monitor for while with Sentry and perhaps ask to observe affected users when this happens.

### DIFF
--- a/fronts-client/src/services/pandaFetch.ts
+++ b/fronts-client/src/services/pandaFetch.ts
@@ -1,8 +1,12 @@
 import { reEstablishSession } from 'panda-session';
 import notifications from './notifications';
+import { isError } from 'tslint/lib/error';
+import Raven from 'raven-js';
 
 const reauthUrl = '/login/status';
-const reauthErrorMessage = `We couldn't log you back in. Your changes may not be saved. Please  <a href="${window.location.href}" target="_self">log in again</a> to reauthenticate.`;
+const reauthErrorMessage =
+  "We couldn't log you back in. Your changes may not be saved. " +
+  `Please <a href="${window.location.href}" target="_self">log in again</a> to reauthenticate.`;
 
 /**
  * Make a fetch request with Panda authentication.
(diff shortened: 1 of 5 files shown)


jonathonherbert (commented):
@twrichards mostly LGTM and great idea to just log, one oddity added to inline comments

fronts-client/src/services/pandaFetch.ts
```diff
@@ -16,26 +20,55 @@ const pandaFetch = (
...
-      return resolve(res);
+        return resolve(res);
+      } catch (error) {
+        // notifications.notify({
+        //   message:
+        //     'Connection issue occurred. Your changes may not be saved. Please wait or reload the page.',
+        //   level: 'error',
+        // });
+        Raven.captureException(error, {
+          extra: `'Connection issue' banner OUGHT TO BE presented to user`,
```
jonathonherbert, line 68: Adding text in `extra` seems to result in odd output. From the sentry request:

```
{project: "35467", logger: "javascript", platform: "javascript",…}
breadcrumbs: {,…}
event_id: "df43f97481fa476490c99363b51b983e"
exception: {values: [{type: "TypeError", value: "Failed to fetch",…}],…}
extra: {0: "'", 1: "C", 2: "o", 3: "n", 4: "n", 5: "e", 6: "c", 7: "t", 8: "i", 9: "o", 10: "n", 11: " ",…}
0: "'"
1: "C"
2: "o"
3: "n"
4: "n"
5: "e"
6: "c"
7: "t"
8: "i"
9: "o"
10: "n"
... etc
```

Nesting in the `extra` field seems to solve, suspect they're iterating on 'extra' and spreading into the request:

```
Raven.captureException(error, {
  extra: { message: `'Connection issue' banner OUGHT TO BE presented to user` },
});
```

jonathonherbert (approved):

fronts-client/src/services/pandaFetch.ts
twrichards: great spot, thanks, I've fixed

jonathonherbert (approved):

PR #1281, opened 2020-12-10
capture the calculated staleness value in Sentry

twrichards:
Minor tweak on https://github.com/guardian/facia-tool/pull/1280, to actually capture the calculated staleness value, in case...
- 30s isn't quite enough
- or, the retrieval of `collectionsLastFetch` is bound to when the function gets created (i.e. some sort of scoping mishap)
- or, some other dark magic 🧙


jonathonherbert (approved):
Can see the metric broadcast locally 👍 

... alongside a key with a colon in it (not your key!). Who _does_ that?!

Screenshot 2020-12-10 at 16 19 26

twrichards:
> ... alongside a key with a colon in it (not your key!). Who _does_ that?!
> 
> Screenshot 2020-12-10 at 16 19 26

gross! looks auto-generated...
image

PR #1283, opened 2020-12-15
Add new special edition

philmcmahon:
## What's changed?
It's the end of the year - time for and end of year special edition!!!! This is unfinished but it would be helpful to have it in master so we can test on CODE without blocking others from using that environment. It has been tested on CODE already - we just need to fix some of the colours.

Note that merging this won't result in any user facing changes - it just makes it possible for the production team to start building the edition. So long as we keep the slug: 'edition-bad-year' the same, we can continue to change this template.


jonathonherbert (approved):
The ... the theme 😅

twrichards (approved):
LGTM (haven't tested locally though)

PR #1284, opened 2020-12-16
fix performance issue of staleness detection

twrichards:
## What's changed?
Previously the logic which checked for staleness of collections was causing re-renders (of quite a large component) on every check (roughly every 10 seconds) which was having a big performance impact (UI lag) for a number of users, because it was calling `setState` regardless of whether the state was actually changing... NOW it only updates the state if `isCollectionsStale` has actually changed. Performance in chrome performance tab looks back to normal, so this means we can re-release the monitoring changes to better understand the staleness (without the staleness check ironically causing things to be more stale 🙈 ).

Also tweaked some of the messages to Sentry so the are more useful.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/services/pandaFetch.ts
+++ b/fronts-client/src/services/pandaFetch.ts
@@ -8,6 +8,8 @@ const reauthErrorMessage =
   "We couldn't log you back in. Your changes may not be saved. " +
   `Please <a href="${window.location.href}" target="_self">log in again</a> to reauthenticate.`;
 
+const AUTH_ERROR_STATUS_CODES = [401, 419];
+
 /**
  * Make a fetch request with Panda authentication.
  *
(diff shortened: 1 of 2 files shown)


jennygrahamjones (approved):

fronts-client/src/services/pandaFetch.ts
```diff
@@ -22,55 +24,61 @@ const pandaFetch = (
...
-          //     'Request failed. Your changes may not be saved. Please wait or reload the page.',
-          //   level: 'error',
-          // });
-          Raven.captureException(
-            `'Request failed' banner OUGHT TO BE presented to user, because... ${res.status} ${res.status}`,
-            { extra: res }
-          );
-          return reject(res);
+        } else if (response.status < 200 || response.status >= 300) {
+          if (count < 1 || !AUTH_ERROR_STATUS_CODES.includes(response.status)) {
```
aug24, line 54: This is a retry implementation, yes?
I ask because the retry appears to add one, rather than subtract one..?

PR #1285, opened 2020-12-16
set the environment property of the `sentryOptions` for easier filtering in the Sentry UI

twrichards:
Minor one but wanted to make use of the `environment` property of the sentry configuration, such that we get nice filtering in Sentry UI (and potentially can set up notifications differently for different environments)...


twrichards:
Closing as this doesn't actually work, probably because we're still using Raven (which is deprecated), we should upgrade to Sentry lib as per...
image

PR #1289, opened 2020-12-21
Update icon url

blishen:
This is an updated icon for the forthcoming special edition. The icon has been passed through the Fastly image resizer that can be found (on dc1 vpn only) here: http://image-url-signer.s3-website-eu-west-1.amazonaws.com/ - the width is identical to the last icon, but it is 3px taller which we're not expecting to be a problem.

## What's changed?
New icon


jonathonherbert (approved):
Image shows for me 👍

PR #1288, opened 2020-12-18
ensure lastSuccessfulFetchTimestamp is set on FETCH_SUCCESS_IGNORE

twrichards:
**_Best to review commit by commit._**

## What's changed?
First off, renamed ambiguous `lastFetch` to be `lastSuccessfulFetchTimestamp` as this what it actually represents.

This PR is mainly to ensure `lastSuccessfulFetchTimestamp` is set on `FETCH_SUCCESS_IGNORE`, as this is relied on for the collections staleness detection (see #1277). Plus added tests to check this and the same for `FETCH_SUCCESS`

## Implementation notes
Co-authored-by: @jonathonherbert 

### General
- [x] 🤖 Relevant tests added
- [x] ✅ CI checks / tests run locally
- [x] 🔍 Checked on CODE

### Client
- [ ] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ ] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #1290, opened 2020-12-21
Do not rerender child collections unnecessarily

jonathonherbert:
## What's changed?

When debugging slow fronts performance w/ @twrichards, we noticed collections were receiving an inline function as a prop, which is causing child collections to rerender every time their parent front rerenders.

This PR moves that function into a private method on the associated class, ensuring that it's stable between rerenders. As a result, containers no longer rerender every time their parent front renders.

This is unlikely to be the cause of the general performance issues facing Fronts in Chrome (that's a separate issue, likely due to a browser rendering performance regression), but is a nice get anyhow, and may improve the performance issues we've had when introducing the last fetch time as a prop to this component (via #1288) 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


tjsilver (approved):
This looks like a useful performance improvement 👍

twrichards (approved):

PR #1298, opened 2021-02-11
First stab at the next special edition

blishen:
We don't know the swatches yet, the colours for the headers and text, the icon etc. This is a start for 10, this will be further modified before launch, once we have some more details. NB the huge amount of fronts is to cover us for the fact that we won't know until late in the day, what the ordering of swatches applied to fronts will.

## What's changed?
New edition template.

## Implementation notes
Edited by a product manager in Github. Beware linting and usual nonsense.

## Checklist
Nothing checked.


aug24:
You're going to need to add stuff to another file before this does anything, yes?

aug24 (changes requested):
Need to add edition to list of editions per the instructions to make it have any effect at all.

philmcmahon:
Need to add it here:
https://github.com/guardian/facia-tool/blob/00155711665a9fe9c325bc7609f868bd8a0ebd11/app/model/editions/EditionsTemplates.scala#L21

and here:
https://github.com/guardian/facia-tool/blob/00155711665a9fe9c325bc7609f868bd8a0ebd11/app/model/editions/EditionsTemplates.scala#L70

I suggest just replacing EditionEarth from those lists

blishen:
Made changes that Pip requested, and updated the branch to main.

philmcmahon (approved):
LGTM

PR #1299, opened 2021-02-11
Coronavirus front for AU

blishen:
As requested by Merren:

## What's changed?
"could we designate Front special 3 as the coronavirus container and have it prefill with coronavirus-tagged content? The top container could be coronavirus+Australia news, then the second container coronavirus+world/UK/US news, then the third can be coronavirus+opinion/comment. I'm not sure if we want to then strip corona-tagged content out of world, national and opinion, but that could be the next step."

## Implementation notes
Edited in github


aug24 (approved):
Assuming the capi queries are good, this lgtm.

PR #1300, opened 2021-02-12
Remove v1 build from teamcity

AWare:
it looks like the teamcity build failures are coming from the v1 build script, which don't seem to be used- this comments out the line in `teamcity-ci.sh` to validate this


AWare:
absolutely not

PR #1308, opened 2021-02-25
Extended EditionDefinition and SpecialEditionButtonStyles

blishen:
This is a stab in the dark a bit - but the Editions team need some new values passed to support their new Edition picker design.
We've attempted to add them here, and if this was the right thing to do, we'd update the next special edition template to include the new values as well as the old.
There will be two versions of the app in the wild, those that support the existing values, and one that will expect the new ones. In time we'd remove: buttonImageUri as it is essentially replaced by pickerButtonImageUri

## What's changed?
David has possibly broken everything - will this change break existing editions? Will these changes flow happily through to the editions back end? These and more questions you can help with!

## Implementation notes
Edited in Github beware linting.

## Checklist
We checked nothing!

### DIFF
--- a/app/model/editions/templates/EditionBadYear.scala
+++ b/app/model/editions/templates/EditionBadYear.scala
@@ -22,7 +22,7 @@ object EditionBadYear extends SpecialEdition {
   override val buttonStyle: Option[SpecialEditionButtonStyles] = Some(
     SpecialEditionButtonStyles(
       backgroundColor = "#6B5840",
-      titleColor = ???,
+      titleColor = "#A1845C",
       title = EditionTextFormatting(color = "#FFFFFF", font="GHGuardianHeadline-Light", lineHeight = 34, size = 34),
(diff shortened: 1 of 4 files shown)


blishen:
@abeddow91 Do we still need these changes or shall we kill this PR? We *think* we can kill this?

blishen:
@abeddow91 Justin is asking if we should change the fields anyway, even if it is just to ensure they refer to the right thing, eg we use titleColor for a title. Let us know, meanwhile we'll prep the books template based on the existing values.

abeddow91:
@blishen - Thanks to design amends we can kill this if we want to. 

Thanks to Ben, we no longer need a new image url.

Title color is a bit more unclear. As you mentioned, this will lead to us using a backgroundColor prop for a title which isn't ideal for clarity but would only be for this one edition (moving forward, we would use the title.color property that exists).

I'm happy to go for a new titleColor value for the sake of clarity but am conscious of time to make sure this is properly tested  and released.

blishen:
OK - agree, no changes now and we can revisit when makes sense.

blishen:
Closing as we're not modifying for now.

PR #1311, opened 2021-03-05
Make Red, Orange.

blishen:
## What's changed?
We've swapped to Opinion for all fronts previously using News.


aug24 (approved):
This is the dullest PR I've ever seen.

blishen:
@aug24 I consider that a compliment.

PR #1317, opened 2021-04-12
Upversion scala 213

aug24:
## What's changed?

Does what it says on the tin: upversions Scala to 2.13.

It also upversions a fair few libraries which were not available for Scala 2.13 at the previous version.

Handle with care: this change should be rigorously tested.

## Implementation notes

Scala 2.13 requires a version of Scanamo which requires the 'new' Amazon AWS SDK.
This means we have two different SDKs for AWS in the build, which is unideal but should not lead to problems.

`// TODO Upversion the rest of the AWS SDK code!`

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


app/controllers/UserDataController.scala
```diff
@@ -83,16 +96,16 @@ class UserDataController(frontsApi: FrontsApi, dynamoClient: AmazonDynamoDB, val
   def putFeatureSwitch() = APIAuthAction { request =>
     val maybeFeatureSwitch: Option[FeatureSwitch] = request.body.asJson.flatMap(
       _.asOpt[FeatureSwitch])
-    val maybeUserData: Option[UserData] = Scanamo.exec(dynamoClient)(
-      userDataTable.get('email -> request.user.email)).flatMap(_.right.toOption)
+    val maybeUserData: Option[UserData] = Scanamo(dynamoClient).exec(
+      userDataTable.get("email" === request.user.email)).flatMap(_.toOption)
```
sihil, line 100: Is this safe? IIUC this can be `None` if there is no record or if scanamo blows up due to permissions or other failure. To be fair it was already like this, so perhaps not something to fix now....

sihil (approved):
This LGTM from a code perspective. I can't speak to the rigourous testing mentioned in the PR.

PR #1324, opened 2021-04-28
Add new special edition: "We were there"

blishen:
## What's changed?
To coincide with the G200 celebrations, we proudly present edition "We were there".

## Implementation notes
Edited by amateurs in Github, a copy of the last books special modified.
We will need to edit again to change icon, and possibly colours - but need to see it in preview.


jonathonherbert (approved):
This passes the high bar I set for a new edition, that is to say, it compiles.

PR #1339, opened 2021-05-18
Add global sports notification topic

frankie297:
## What's changed?
Add a global sports notification breaking news topic

### General
- [x] 🔍 Checked on CODE


jonathonherbert (approved):
Looks good, can see this is happy in CODE – thanks @frankie297!

PR #1341, opened 2021-05-20
Prevent form from being dragged when selecting text in sublinks

jonathonherbert:
## What's changed?

Prevents the form initiating a drag event. This was problematic when users were e.g. clicking and dragging to select text in inputs (card [here](https://trello.com/c/46B9iYGV/2362-fronts-ui-bug)):

no-drag-ko

Now, things behave as they should:

no-drag-ok

## How to test

Difficult to write an automated test for this – best to try the above locally, or in CODE, to see that it works as expected.


rhystmills (approved):
Tested on CODE and working splendidly for me.

SHession (approved):
Also tested locally, happy it's working! Going to merge today so we can get back to the user.

PR #1342, opened 2021-05-21
Revert "Prevent form from being dragged when selecting text in sublinks"

SHession:
Reverts guardian/facia-tool#1341 due to issues with selecting text in firefox.


rhystmills (approved):
Oops!

PR #1343, opened 2021-05-24
Move card sublinks out of drag container

jonathonherbert:
## What's changed?

Another attempt to solve the problem described in #1341 (see that card for gifs and detail). Instead of attempting to intercept drag events (which had problems with cross-browser behaviour, hence the revert #1342), this PR moves sublinks out of the draggable container entirely – AFAICS there's no reason to nest these elements.

## Implementation notes

In doing this, we needed a bit of a refactor to move the code around, hence the slightly larger diff – we've shuffled the drag intent detection up from the `Article` component to its parent, the `Card` component.

This also fixes a previously unnoticed bug where `Snaplink`s (a kind of `Card`, along with `Article`) did not show hint hover states when cards were hovering over them – as we've moved the hover detection out of `Article` to a place that handles both kinds of card.

Before:

snap-link-hover-ko

After:

snap-link-hover-ok

Tested in Chrome and Firefox.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/FrontsEdit/CardFormInline.tsx
+++ b/fronts-client/src/components/FrontsEdit/CardFormInline.tsx
@@ -362,15 +362,6 @@ class FormComponent extends React.Component<Props, FormComponentState> {
           (e: React.MouseEvent) =>
             e.stopPropagation() /* Prevent clicks passing through the form */
         }
-        // Making this component draggable, but preventing drag events from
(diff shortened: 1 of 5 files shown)


twrichards (approved):
Logic/refactors look good 👍  
Nice PR description.
(haven't tested myself as I've seen you've already tested in CODE)

PR #1351, opened 2021-06-21
Proposal for including Saturday magazine in the "Life" front

blishen:
Includes prefills for the new Saturday Magazine book tags

## What's changed?
Four new hidden containers at the end (almost) of the Life front on Saturday.
One for each new booksection tag. To be reviewed by the Editorial team.


blishen:
We're not going to merge this, while KV talks to Mel about how they want to represent the new magazine in the Editions app. We might, for example, create a new front instead of pushing it into the existing Life front.

KV also to think about what swatch should be applied to the edition.

jonathonherbert (approved):

PR #1353, opened 2021-06-29
Detect invalid URL when creating plain snap link, and launch error modal

jonathonherbert:
## What's changed?

This PR detects an invalid link when creating plain snap link, and launches an error modal.

An invalid link, for a snap link, is anything that isn't a valid URL, and doesn't start with a http(s) protocol (for example, a `mailto:` address would be pass the former test, but not the latter.)

Before (oh noo):
Screenshot 2021-06-29 at 09 38 49

After:
Screenshot 2021-06-29 at 09 31 16

## How to test

Drag and drop something egregious into the Fronts tool (e.g. this link, which you may need to save as a bookmarklet, and markdown refuses to turn into an `a` tag: `javascript:(function(){document.body.appendChild(document.createElement('script')).src='https://dashboard.ophan.co.uk/assets/js/heatmap-bookmarklet.js';})();`

You should get the above modal ... and definitely _not_ a snaplink.


SHession (approved):
Changes make sense, and the it's good to see some tests. I've tested on CODE and it looks like it's working well. 

My only concern is that there could be issues around unusual items in the fronts, but I'm not familiar enough with the tool to know for sure.

jonathonherbert:
Agreed – we should have specs for most of the things we've already encountered as unusual, but if we encounter them, let's roll back, add to our specs, and redeploy.

PR #1356, opened 2021-07-01
Update DailyEdition.scala

blishen:
## What's changed?
Added two additional containers for both World fronts, and one more for Food on a Saturday.

## Implementation notes
Duplicated existing hidden containers. Edited in Github.


paperboyo (approved):
Business logic only, so: 👍 from a code-blind.

PR #1359, opened 2021-07-13
Firefox banner

jonathonherbert:
_co-authored with: @rhystmills_ 

## What's changed?

Adds a banner to the Fronts tool, alerting users to use Firefox if the tool feels slow and they're using a browser that we know is affected by the bug detailed in [this card](https://trello.com/c/iII6siii/2154-fronts-tool-perf-issues-in-chrome):

Screenshot 2021-07-14 at 08 42 55

The banner will not reappear after the use dismisses it. The data is stored against the user profile in Dynamo, so it's tied to the Google account's e-mail, not e.g. local storage.

To reset the message, flip the relevant switch in `<tool>/v2/features`

## Implementation notes

In doing this, we've also fixed a bug in the feature switches code that would give the frontend more feature switches than existed in user data. We've added tests for that.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/FeatureSwitches.scala
+++ b/app/model/FeatureSwitches.scala
@@ -24,8 +24,14 @@ object PageViewDataVisualisation extends FeatureSwitch(
   enabled = true
 )
 
+object ShowFirefoxPrompt extends FeatureSwitch(
+  key = "show-firefox-prompt",
+  title = "Show the prompt to use Firefox if applicable",
+  enabled = true
+)
+
 object FeatureSwitches {
-  val all: List[FeatureSwitch] = List(ObscureFeed, PageViewDataVisualisation)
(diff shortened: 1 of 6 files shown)


twrichards (approved):
Great stuff, lovely tests. Pleased to see the reuse of feature switches rather than a random bit of local storage.

On question (non blocking) could we capture an expiry such that the user gets re-promoted if they continue to use Chrome having dismissed the banner because people habitually dismiss banners.

jonathonherbert:
We could do – but there are legitimate reasons for people continuing to use Chrome. Only those working on multiple, large fronts for long periods of time are likely to notice these problems – others should be fine. We'll be sending emails alongside this, which can be repeated.

I'd be tempted to say, let's keep CP aware that there are problems, and they can provide advice should users report them.

SHession (approved):
Tested locally, banner and messaging looks good.

PR #1360, opened 2021-07-15
Improves the V1 image validation process

SHession:
## What's changed?

This PR aims to fixes occurrences of internal Grid URL being distributed with a breaking news alert. Examples of which can be [seen in the logs](https://logs.gutools.co.uk/s/editorial-tools/goto/dc17b298da75d48f88241d3c9a9f5986). 

This issue appears to be related to how grid internal links are processed into links to S3. This process can be latent and result in the original URL not being replaced before the image is used in a front. The GIF below replicates a possible cause of the issue:

Before
Fronts_issue_2

After
Fronts_issue_1

This fix attempt proactively removes the entered src and assumes it's wrong until the images is retrieved via the image validate process. This should mean the src is only ever set correctly.

## Implementation notes

This functionally isn't 100% faithful to the current behaviour, setting a rejected src now doesn't clear the input for reason I can't discern due to the complexity of the codebase. This might also highlight other potential unknowns. Certain units tests are now also failing.

EDIT: The image source will now only be set once the chain of promises completes successfully, this will mean the thumbnail image accurately represents the image attached to the piece, even if the source url is invalid and left in place. This should provide a clear indication to the user what image will  be launched and highlight any issues. It may be the case we have reports of images not updating properly, this will highlight existing issues in a more obvious way and could justify further investigation.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/public/src/js/models/article/editor.js
+++ b/public/src/js/models/article/editor.js
@@ -154,6 +154,8 @@ export default class Editor extends BaseClass {
             image = imageSrc(),
             opts = params.options;
 
+        this.assignImageToSpreadElement(params, null);
+
         if (image) {
             let {src, origin} = extractImageElements(image);
             const frontId = this.article.front ? this.article.front.front() : 'clipboard';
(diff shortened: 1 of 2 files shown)


jonathonherbert (approved):
Great work, good to see the side-effects documented in this PR. As you say, better to surface odd behaviour to users – if it's problematic we can take another look, but at least users are aware of the true state of the application in the meantime.

PR #1367, opened 2021-08-12
Updates jspm and jspm-github versions

SHession:
## What's changed?

This PR updates the `JSPM` and `JSPM-GitHub` versions to latest patch versions, inline with [story-packages](https://github.com/guardian/story-packages/blob/main/package.json). This was necessary due to issue with the build failing on JSPM due to invalid GitHub credentials. 

Prior to this version of the dependencies the auth token was expected to be in the format [username:token](https://jspm.org/docs/0.16/registries.html), this now no longer seems to be the case, with only the access token being necessary. This PR ensures consistency between this project and `story-packages` and should allow us to use a shared parameter between the two. 

## Implementation notes

`story-packages` and this project stopped building for a currently unknown reason. The change in auth shape should only have affected story-packges, but both failed. It's possible the old credentials were deactivated by whomever owned them. 

It's also worth noting I haven't been able to find documentation for this change, from [the latest update PR](https://github.com/jspm/github/issues/136), it looks like the old `username:token` format should still be supported, but it can just accept a personal access token. NB Setting up JSPM creds using the interactive mode ultimately sets the auth credentials to the access token and strips the username and base64 encoding. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


SHession:
> Looks good, what a 🐰 🕳️! Presume that a passing build is evidence of success?

That's right it does now work: https://teamcity.gutools.co.uk/buildConfiguration/EditorialTools_FaciaTool?mode=branches

PR #1379, opened 2021-09-17
Fix setup and configuration process

dskamiotis:
## What's changed?

Currently the process outlined in the read me does not work.
By running `./scripts/setup.sh` it would not execute the fetch-config as part of the process, and would therefore need to be executed independently as a separate step prior to setup.

This PR aims to bring the setup process back in line to follow the steps outlined in the read me docs for setting up.
## Implementation notes

- all logic from the fetch-config module into the setup, so all functions are in one file
- the config is now invoked from the main function within the same module.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

paired with @akash1810


shtukas (approved):

build.sbt
```diff
@@ -60,8 +60,8 @@ TwirlKeys.templateImports ++= Seq(
 routesImport += "model.editions._"
 
 val awsVersion = "1.11.999"
-val capiModelsVersion = "15.6"
-val capiClientVersion = "17.1"
+val capiModelsVersion = "17.1.0"
+val capiClientVersion = "17.21"
```
akash1810, line 64: Are these changes intended to be in this PR? They're of a different concern to me and I'd suggest they belong in their own PR.

dskamiotis: No quite right these are unrelated, thanks will revert them 👍

akash1810 (approved):

PR #1381, opened 2021-09-20
update fapi-client library to 3.3.7

dskamiotis:
## What's changed?

This is a simple PR to update to `"com.gu" %% "fapi-client-play27" % "3.3.7",`.
This will allow the client to be able to add the Podcast item to the list of tags.
## Implementation notes

I significant jump from 3.2.0 to 3.3.7

This has been initially tested using SNAPSHOT locally and from facia-tool to ensure the logic change in the library works.

In addition:
-  CAPI libs bumped `capiModelsVersion = "17.1.0"`   `capiClientVersion = "17.21"`
-  Updates the models used in fronts to be in line with latest CAPI models. 
-  This results in all test passing

| Before  | After |
| --- | --- |
| image| image|

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

Tested by adding `Podcast` from tags dropdown in ` https://fronts.code.dev-gutools.co.uk/editorial/config`
### Client

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -75,7 +75,7 @@ download_config() {
 
 fetch_config(){
     if [[ -w ${CONFIG_DIR} ]]; then
-        downloadConfig
+        download_config
         echo -e "${GREEN}Done ${NOCOLOUR}"
     else
       echo "Cannot write to ${CONFIG_DIR}. It either doesn't exist or is not owned by $(whoami)."
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -75,7 +75,7 @@ download_config() {
 
 fetch_config(){
     if [[ -w ${CONFIG_DIR} ]]; then
-        download_config
+        downloadConfig
         echo -e "${GREEN}Done ${NOCOLOUR}"
     else
       echo "Cannot write to ${CONFIG_DIR}. It either doesn't exist or is not owned by $(whoami)."
--- a/build.sbt
+++ b/build.sbt
@@ -60,8 +60,8 @@ TwirlKeys.templateImports ++= Seq(
 routesImport += "model.editions._"
 
 val awsVersion = "1.11.999"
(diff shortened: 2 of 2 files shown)


Reettaphant:
Have you checked that you can still press the fronts after this change ie that frontend can still process any changes you make? To test this edit a front locally and add the new podcast tag on a container, save and launch changes and then check for changes here: http://m.code.dev-theguardian.com/{name-of-front} (more instructions in the Readme)

dskamiotis:
> Have you checked that you can still press the fronts after this change ie that frontend can still process any changes you make? To test this edit a front locally and add the new podcast tag on a container, save and launch changes and then check for changes here: [http://m.code.dev-theguardian.com/{name-of-front}](http://m.code.dev-theguardian.com/%7Bname-of-front%7D) (more instructions in the Readme)

dskamiotis:
Yes all working, I added the Podcast tag and checked pressed:
image

Reettaphant:
Looks great, thank you!

scripts/setup.sh
```diff
@@ -75,7 +75,7 @@ download_config() {
 
 fetch_config(){
     if [[ -w ${CONFIG_DIR} ]]; then
-        downloadConfig
+        download_config
```
jonathonherbert, line 78: Is this change necessary in this PR?

dskamiotis: I believe this is invoking the function which now lives in this file (since we changed the set up config in another PR).
I thought it may cause an error in CI/test builds if not amended. However seems to be working fine locally.

jonathonherbert: Ah, yeah – it may be working locally because there are other changes in your git tree that are untracked, or haven't been committed 👍

jonathonherbert (approved):
Change looks good, testing looks good – great work.

PR #1382, opened 2021-09-21
What's on and Environment

blishen:
## What's changed?
We've added What's On and Environment News booksections. We heard about this late in the day.
We're adding them into existing sections for now, and will tidy up when KV has had a chance to think about.

## Implementation notes
Edited by a PM in Github. Buyer beware.


tonytw1 (approved):
Looks like a straight addition which the comitter knows about.

PR #1386, opened 2021-09-27
Db post sat cleanup

blishen:
## What's changed?
Cleaned up after the Saturday mag. Removed old sections and rejigged.
## Implementation notes
Edited in Github


paperboyo (approved):
Business logic-only, so I feel I can approve. Plus, I trust you, Boss (and I have double-checked `theguardian/saturday/lifestyle` is a thing).

PR #1398, opened 2021-11-10
Update logging configuration

dskamiotis:
## What does this change?

Points facia-tool fronts to our new, account specific logging stream.
This would be the new Kinesis Role Arn as redefined on the stack.

## How to test

Deploy to CODE. Do we still see logs in ELK as usual?


jonathonherbert (approved):
Looking good in CODE!

PR #1399, opened 2021-11-16
Add fluentbit to facia-tool

dskamiotis:
co-authored with @jonathonherbert 
## What's changed?

Adding `fluentbit` to fronts tool and removing the kinesis logging appender.

This PR aims to ship logs via [Fluent Bit](https://docs.fluentbit.io/manual/pipeline/outputs/kinesis) instead of using the standard [Logback Kinesis appender](https://github.com/guardian/kinesis-logback-appender). This was related to the general tools outage, more information can be found in the [incident document](https://docs.google.com/document/d/17jGwlvc5F-SOCSZ90lNl5ZRpJMvvaJ_UXtuUyY4ZM1s/edit#heading=h.98xvxa3bjhw).

This work was completed as part of a spike, the main objective of which is improve our logging by being more robust in the face of issues with the Kinesis stream. This change achieves this by removing writing to the Kinesis stream as an application concern, and instead relies on Fluent Bit running on the instances. Fluent Bit becomes responsible for handling our logs and our application is not blocked by these concerns.

** For more details on our approach please see [this PR](https://github.com/guardian/atom-workshop/pull/304)

Existing issue:
Currently this initial Play logs are output to journald without being well formatted, this means we see logs in ELK which are split across line. It would be better if these could be grouped together and formatted properly:
image
 

## How to test

Deploying to CODE in conjunction with the necessary [CFN changes](https://github.com/guardian/editorial-tools-platform/pull/564). Do the [logs show up in ELK](https://logs.gutools.co.uk/goto/ec7a6d8c65121a9fbd2cac29f2430d1a)?

## How can we measure success?

Our logs no longer block the running application, they are therefore more reliable. More logs are also included as a result, which may be useful in debugging.

** For more details on our approach please see [this PR](https://github.com/guardian/atom-workshop/pull/304)


rhystmills (approved):
All looking good to me 👍 . I've checked the CODE logs and everything seems to be there as expected, with no duplication of logs.

App is also successfully building.

PR #1400, opened 2021-11-23
Add new email permission

jonathonherbert:
## What's changed?

Following on from [this PR in the permissions tool](https://github.com/guardian/permissions/pull/141), this PR makes use of a new email permission to restrict users with just that permission to accessing email fronts only.

We've preserved the logic that allows users with editorial permissions to edit email fronts.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### DIFF
--- a/app/controllers/BaseFaciaController.scala
+++ b/app/controllers/BaseFaciaController.scala
@@ -61,7 +61,6 @@ abstract class BaseFaciaController(deps: BaseFaciaControllerComponents) extends 
 
   def getCollectionPermissionFilterByPriority(priority: String, acl: Acl)(implicit ec: ExecutionContext): ActionBuilder[UserRequest, AnyContent] = {
     val permissionsPriority = PermissionsPriority.stringToPermissionPriority(priority)
-    println(permissionsPriority)
     permissionsPriority match {
       case Some(EditorialPermission) => AccessAuthAction andThen new EditEditorialFrontsPermissionCheck(acl)
       case Some(CommercialPermission) => AccessAuthAction andThen new LaunchCommercialFrontsPermissionCheck(acl)
--- a/app/controllers/BaseFaciaController.scala
+++ b/app/controllers/BaseFaciaController.scala
@@ -61,10 +61,11 @@ abstract class BaseFaciaController(deps: BaseFaciaControllerComponents) extends 
 
   def getCollectionPermissionFilterByPriority(priority: String, acl: Acl)(implicit ec: ExecutionContext): ActionBuilder[UserRequest, AnyContent] = {
     val permissionsPriority = PermissionsPriority.stringToPermissionPriority(priority)
+    println(permissionsPriority)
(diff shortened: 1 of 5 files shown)


app/controllers/BaseFaciaController.scala
```diff
@@ -61,10 +61,11 @@ abstract class BaseFaciaController(deps: BaseFaciaControllerComponents) extends
 
   def getCollectionPermissionFilterByPriority(priority: String, acl: Acl)(implicit ec: ExecutionContext): ActionBuilder[UserRequest, AnyContent] = {
     val permissionsPriority = PermissionsPriority.stringToPermissionPriority(priority)
+    println(permissionsPriority)
```
andrew-nowak, line 64: is this a debugging println?

jonathonherbert: Yep – thanks for spotting!

app/permissions/PermissionsActionCheck.scala
```diff
@@ -85,6 +88,18 @@ class EditEditorialFrontsPermissionCheck(val acl: Acl)(implicit ec: ExecutionCon
...
+  override implicit val executionContext: ExecutionContext = ec
+  override val testAccess: String => Authorization = (email: String) => {
+    val emailAccess = acl.testUser(Permissions.LaunchAndEditEmailFronts, "facia-tool-allow-edit-email-fronts-for-all")(email)
+    emailAccess match {
+      case AccessGranted => emailAccess
+      case AccessDenied => new EditEditorialFrontsPermissionCheck(acl)(executionContext).testAccess(email)
+    }
+  }
+  override val restrictedAction: String = "Edit email fronts."
+}
```
andrew-nowak, line 101: nit: I think if a user has neither permission, this will show a message saying unauthorized to edit editorial fronts, rather than unauthorized to edit email fronts. It might be nice to warn/suggest the narrower permission first?

andrew-nowak (approved):

app/permissions/PermissionsActionCheck.scala
jonathonherbert: We'd definitely want to avoid that – is there a side effect in this method when we delegate to `EditEditorialFrontsPermissionCheck`? The return value should be just `AccessDenied` or `AccessGranted`, so I'm unsure how the caller would distinguish between the two paths – unless that function has a side-effect I've missed 🤔 

In general, I don't think users see a particular permission when access is denied in Fronts, although it may appear in logs:
Screenshot 2021-11-26 at 11 59 29

andrew-nowak: You're completely right! I've misunderstood how the restrictedAction was being used, ignore this comment :)

jonathonherbert: Ah, gotcha, no problem!

PR #1406, opened 2021-12-20
Updated colours and menu icon for End-of-Year Special Edition

groakland:
Updated colours and the Special Edition menu icon for edition-end-of-year

image

### DIFF
--- a/app/model/editions/templates/EditionEndOfYear.scala
+++ b/app/model/editions/templates/EditionEndOfYear.scala
@@ -21,7 +21,7 @@ object EditionEndOfYear extends SpecialEdition {
   )
   override val buttonStyle: Option[SpecialEditionButtonStyles] = Some(
     SpecialEditionButtonStyles(
-      backgroundColor = "#BB3B80",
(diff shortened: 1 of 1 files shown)


app/model/editions/templates/EditionEndOfYear.scala
```diff
@@ -10,18 +10,18 @@ import org.joda.time.{DateTime, DateTimeZone}
 //noinspection TypeAnnotation
 object EditionEndOfYear extends SpecialEdition {
   override val title = "Best of 2021"
-  override val subTitle = "A special edition of the Guardian's best journalism from 2021"
+  override val subTitle = "A lookback at our readers' and editors' favourite pieces of the year"
```
paperboyo, line 13:
```suggestion
  override val subTitle = "A lookback at our readers’ and editors’ favourite pieces of the year"
```

Just curlies…

paperboyo (approved):
Approved as just business logic, so my inability to read code doesn’t stand in the way…

PR #1410, opened 2021-12-22
Update EditionEndOfYear.scala

blishen:
## What's changed?
Swapped the 'grid' image for a signed url

## Implementation notes
Used this
http://image-url-signer.s3-website-eu-west-1.amazonaws.com/


tjsilver (approved):
I assume that `s` is the signature for the image? I also noticed that the quality is set in the signed image but not in the original - if that's intentional then 👍

groakland (approved):
LGTM

PR #1412, opened 2022-01-17
Set timestamp in output jars from SBT

andrew-nowak:
Since SBT 1.4.0, SBT has edited the last modified dates in packaged
JARs, which include web assets (css, js, etc.); Grid's current version
sets the date to 1 Jan 2010. Play framework uses these dates to
calculate the ETag and Last-Modified headers. Since the headers are now
based upon a static date, caching (especially via Cloudfront) is broken,
returning outdated assets.

This commit restores the previous behaviour, so cloudfront should now
notice and pick up new assets upon deployment.

See also <playframework/playframework#10572>
See also <sbt/sbt#6237>

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -75,6 +75,14 @@ resolvers ++= Seq(
 
 PlayKeys.devSettings := Seq("play.akka.dev-mode.akka.http.parsing.max-uri-length" -> "20480")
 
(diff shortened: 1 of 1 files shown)


andrew-nowak:
not required as facia-tool sends assets with no-cache directives

PR #1438, opened 2022-04-12
Restrict slideshows with less than two images

Fweddi:
## What's changed?

We have always allowed slideshows containing one image. Until now, we think very few, if any, have been made. 

However, now we have released captions, there is a (bizarre) incentive to publish a slideshow with only one image. Slideshows are the only fronts media to have captions, so if you want a single image with a caption, you can create a slideshow with that one image.

This looks bad, and isn't the intended behaviour. This PR adds a simple restriction in the tooling to prevent this. Now you can only publish a slideshow with more than one image.

https://user-images.githubusercontent.com/40991816/163005580-3b4c81d8-5f27-445a-9d82-52bad6262a44.mov

The form is disabled if (a) slideshow is enabled and (b) the slideshow has fewer than two images present.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/FrontsEdit/CardFormInline.tsx
+++ b/fronts-client/src/components/FrontsEdit/CardFormInline.tsx
@@ -85,8 +85,7 @@ type Props = ComponentProps &
 type RenderSlideshowProps = WrappedFieldArrayProps<ImageData> & {
   frontId: string;
   change: (field: string, value: any) => void;
-  hasAtLeastTwoImages: boolean;
-  setHasAtLeastTwoImages: (state: boolean) => void;
+  slideshowHasAtLeastTwoImages: boolean;
 };
 
 const FormContainer = styled(ContentContainer.withComponent('form'))`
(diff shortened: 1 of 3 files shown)


fronts-client/src/components/FrontsEdit/CardFormInline.tsx
```diff
@@ -446,6 +481,9 @@ class FormComponent extends React.Component<Props, FormComponentState> {
 
     const imageDefined = (img: ImageData | undefined) => img && img.src;
 
+    const setSlideshowHasAtLeastTwoImages = (state: boolean) =>
```
jonathonherbert, line 484: This approach will work, but we're mantaining this property as state, when it could be derived from the current form state. I can see why it's necessary for this approach – we need to get the slideshow state out of our FieldArray component somehow.

If it's possible (unsure if you've tried this!), I suspect it'd be simpler to pass through necessary slideshow values to our form directly as props from `mapStateToProps`. This can be done with the `valueSelector` function, which we already use (e.g. [here](https://github.com/guardian/facia-tool/blob/2d8462c6f43cef56401f8cbe24811386ce7c33f1/fronts-client/src/components/FrontsEdit/CardFormInline.tsx#L1049)) to pass through form values so our form can react to them.

V. happy to pair.

Fweddi: I knew there would be a way with the form state, but I couldn't find it - will give this a go!

jonathonherbert (approved):
Tested with @Fweddi locally, LGTM!

PR #1430, opened 2022-03-29
Correct preview for code

jonathonherbert:
_co-authored with @twrichards__ 

## What's changed?

A second attempt to correct the preview link for non-PROD environments, from `https://preview.gutools.co.uk` to `m.code.dev-theguardian.com`, following an initial attempt, #1424. 

The first attempt missed the `https://` in the PROD path, which causes some very interesting problems in preview:

❌  `https://preview.gutools.co.uk/responsive-viewer/preview.gutools.co.uk/uk` (this might trouble your browser, as preview will recursively preview ... itself)
✅  `https://preview.gutools.co.uk/responsive-viewer/https://preview.gutools.co.uk/uk`

It's difficult to verify the behaviour without deploying, so I've added a unit test to verify the link is good when the client is in PROD.

## Implementation notes

I've also tidied up the types in our config, which had two mistakes I spotted – a mispelling, and a property that ... didn't exist 😅 

## How to test

Run locally, or deploy to CODE. Does the preview now take you to the correct URL?

### General
- [ ] 🤖 Relevant tests added
- [ ] ✅ CI checks / tests run locally
- [x] 🔍 Checked on CODE

### Client
- [x] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [x] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #1440, opened 2022-05-03
Restore 'preview' target

jonathonherbert:
## What's changed?

From a user:

> I wonder if anyone can help me with this. On the fronts tool it has suddenly started opening a new tab every time I preview, rather than overlaying on the previous preview. Consequently i spend my life closing tabs, which is obviously quite annoying. A couple of my colleagues say they have been living with this same glitch for a while now so I wonder if you could fix it, or point me in the right direction.

Likely introduced by the `target="_blank"` attribute in #1430.

This restores the previous behaviour. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


Fweddi (approved):
Assuming the blank target was to speed up dev work on https://github.com/guardian/facia-tool/pull/1430?

PR #1447, opened 2022-06-30
Guess what kids! Covid is over, at least down under.

blishen:
## What's changed?
Removes the Covid section from the edition.

## Implementation notes
Edited by a product manager in Github.

## Checklist

Katy Vans and David can create and delete an edition in PROD before production starts for real.


Fweddi (approved):
For historians: I'm approving the PR, not the message.

PR #1468, opened 2022-11-15
Financial is now Business (again)

blishen:
Rejig of the Financial sections. Fortunately book tags are unchanged, so not a big change.

## What's changed?
Renamed two fronts and associated containers.
Made the containers Mon - Saturday more logical.
Split Cash into its own container on a Sunday.

## Implementation notes
Edited by a toothless, jetlagged Product Manager in Github.


paperboyo (approved):
Based on (blind) trust! (but well founded)

PR #1483, opened 2022-12-07
update karma

alinaboghiu:
## What's changed?

Updating Karma libs to fix critical vulnerabilities. 

Also enabling karma tests to run as part of the build and fixing some failing tests.

### DIFF
--- a/grunt-configs/karma.js
+++ b/grunt-configs/karma.js
@@ -6,7 +6,7 @@ module.exports = function(grunt, options) {
             singleRun: options.singleRun,
             browserDisconnectTimeout: 10000,
             browserDisconnectTolerance: 3,
-            browserNoActivityTimeout: 60000,
+            browserNoActivityTimeout: 400000,
             reportSlowerThan: 2000,
             colors: options.color,
             port: 9876,
--- a/public/src/js/models/config/collection.js
+++ b/public/src/js/models/config/collection.js
@@ -11,7 +11,6 @@ import fullTrim from 'utils/full-trim';
 import populateObservables from 'utils/populate-observables';
 import urlAbsPath from 'utils/url-abs-path';
 import isPlatformSpecificCollection from 'utils/platform';
-import getAvailableTerritories from 'utils/get-available-territories';
 import CONST from 'constants/defaults';
 
 export default class ConfigCollection extends DropTarget {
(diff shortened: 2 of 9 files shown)


public/src/js/models/config/collection.js
```diff
@@ -20,11 +19,9 @@ export default class ConfigCollection extends DropTarget {
 
         this.id = opts.id;
 
-        const defaults = vars.model.state().defaults;
-
         this.parents = ko.observableArray(findParents(opts.id));
         this.userVisibilities = CONST.userVisibilities;
-        this.availableTerritories = getAvailableTerritories(defaults);
+        this.availableTerritories = [];
```
jonathonherbert, line 24:
```suggestion
        this.availableTerritories = defaults ?? [];
```

Totally agree that we should remove the silly function, but this removes the defaults altogether, which means we can no longer target territories:

Screenshot 2022-12-12 at 09 21 01

Unsure if this was a response to breaking tests – wonder if they're need in a fixture somewhere.

jonathonherbert (commented):
Can see these running, and passing – great to have them back.

public/src/js/utils/validate-image-src.js
```diff
@@ -142,13 +142,13 @@ function validateActualImage (image, frontId) {
         } else if (criteriaRatio && criteriaRatio - ratio > 0.01) {
             reject(new Error('Images must have a ' + widthAspectRatio + ':' + heightAspectRatio + ' aspect ratio'));
         } else {
-            const resolveWithImage = () => { resolve({ path, origin, thumb, width, height });};
+            const resolveWithImage = () => { return resolve({ path, origin, thumb, width, height });};
```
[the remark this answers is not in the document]
alinaboghiu: thank you!

grunt-configs/karma.js
```diff
@@ -6,7 +6,7 @@ module.exports = function(grunt, options) {
             singleRun: options.singleRun,
             browserDisconnectTimeout: 10000,
             browserDisconnectTolerance: 3,
-            browserNoActivityTimeout: 60000,
+            browserNoActivityTimeout: 400000,
```
alinaboghiu, line 9: This was because locally PhantomJS takes longer to start up (I wasn't able to understand why this is the case, doesn't seem to be an issue when running in TeamCity)

public/src/js/models/config/collection.js
alinaboghiu: Having looked into this further you are completely right! 

`defaults` come from the server on calling /defaults 

Tests were failing because defaults will be undefined when the tests run so adding an extra check to fix that.

All should be well

jonathonherbert: Ace – just tested, territories are now there.

jonathonherbert (approved):

PR #1502, opened 2023-03-31
Fix link to gloassary

davidfurey:
## What's changed?
Tiny change fixing the glossary link


ParisaTork (approved):

ParisaTork:
N.B. TC build is failing - definitely not due to this change, but worth investigating the cause of the execution timeout

PR #1503, opened 2023-04-05
add stale PR checker

aracho1:
## What does this change?
Automatically mark any pull requests that have been inactive for 30 days as "Stale" then close them 3 days later if there is still no activity.

Uses standard template: https://github.com/actions/stale#all-options

## How to test
I think this PR would need to be merged and we can see if it marks and closes stale PRs accordingly

## How can we measure success?
Stale PRs are marked and closed according to set timelines

## Have we considered potential risks?
Low risk as we can always re-open stale PRs if needed


rebecca-thompson (approved):

rebecca-thompson (commented):
It looks like the build for this timed out, so I would rerun in TC to see it passing on the pr before merging

PR #1510, opened 2023-05-17
Remove s3 assets

davidfurey:
## What's changed?

Serves JS bundles from app instead of S3. Serving from S3 adds unnecessary complexity and increases our count of public S3 buckets.

Once this is released, we should update our cloudformation since it no longer needs to drop `/etc/gu/facia-tool.assets-map.json` on the instance.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


Gruntfile.js
```diff
@@ -40,10 +40,7 @@ module.exports = function (grunt) {
     /**
      * Compile tasks
      */
-    grunt.registerTask('compile', function () {
-        grunt.task.run(['clean', 'shell:collections', 'shell:config', 'replace', 'cacheBust']);
-    });
     grunt.registerTask('bundle', function () {
```
alinaboghiu, line 43: think we're meant to keep the `compile` task rather than `bundle` - tasks look correct, just the name

alinaboghiu (approved):

PR #1511, opened 2023-05-24
GitHub actions build

davidfurey:
## What's changed?

Builds project on Github Actions instead of Team City

## Release

Once this PR is reviewed, we need to:
- [x] disable the TeamCity build
- [x] set `LAST_TEAMCITY_BUILD` in `ci.yml` to match the actual last teamcity build
- [x] Revert "Skip riffRaffUpload during testing"
- [x] Add configure-aws-credentials similar to https://github.com/guardian/typerighter/blob/main/.github/workflows/ci.yaml#L14


rebecca-thompson (approved):
To confirm for my understanding: this breaks our current CI by stopping Teamcity from publishing to RiffRaff, so we can safely raise a second pr with the last build number in TC.

Out of curiously, why aren't we using the [RiffRaff GH actions](https://github.com/guardian/actions-riff-raff)?

davidfurey:
I've broken our current CI build by pausing it in the console. This PR doesn't stop it.

Not using RiffRaff GH actions to keep this PR as small as possible (and because I haven't used it before).

I've added two more commits which might make the PR make more sense

PR #1512, opened 2023-05-24
Fix integration tests on M1 Macbook

davidfurey:
## What's changed?

Use docker compose instead of docker-testkit. Docker-testkit doesn't seem to work on M1 Macs. The library has changed enough that upgrading is difficult. Also, this pattern of using docker-compose for integration tests is used in other projects.

Port 4724 is a random choice to avoid issue of multiple docker postgres databases requiring the same host port.


Fweddi (approved):
This now requires the docker-compose script to be run, but this is run by the dev-start script (and other apps have the same requirement).

All works locally.

PR #1514, opened 2023-05-24
Remove analytics

davidfurey:
## What's changed?

Rips outs analytics from v1 and v2 clients.  We aren't using this data, and if we don't remove it we need update it to the newest GA SDK (GA4) before first of July

## Implementation notes

This change touches a lot of files, so we should extensively test in CODE and coordinate the release with CP.

Once this is released and we are happy we won't need to roll back, we should remove the (now used) analytics secret from the CODE and PROD configs.


davidfurey:
I'm exploring less drastic approaches

rhystmills:
Removing it seems like a good idea to me! We removed it from `media-atom-maker` for similar reasons: https://github.com/guardian/media-atom-maker/pull/1094

It's not being used to analyse user behaviour at the moment (or in recent years) and I've not had any luck getting access.

PR #1515, opened 2023-05-25
Disable analytics

davidfurey:
## What's changed?

Disables analytics from v1 and v2 clients.  We aren't using this data, and if we don't remove it we need update it to the newest GA SDK (GA4) before first of July

## Implementation notes

This leaves in place most of the tracking points but with noop tracking functions, so it should be easy for us to plug in another tracking solution in future.

Once this is released and we are happy we won't need to roll back, we should remove the (now used) analytics secret from the CODE and PROD configs.


Fweddi (approved):

davidfurey:
Deployed to CODE. Limiting testing of normal fronts and config editor both seem fine.

PR #1519, opened 2023-06-01
Simplify snyk action

NovemberTang:
## What's changed?

The snyk workflow has been significantly simplified by migrating it to the Guardian's [centralised snyk action](https://github.com/guardian/.github/blob/main/.github/workflows/sbt-node-snyk.yml).

The old project on snyk needs to be deleted after this PR is merged.

## Implementation notes

The cron job that ran at 6am every morning has been removed. This is not needed as snyk only needs to be notified when dependencies change, and will run regular checks on existing projects independently (usually daily)

## Testing

run the action using the github UI, verify it shows up in snyk

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -1,45 +1,19 @@
-# This action runs every day at 6 AM and on every push
-# If the branch it's running on is main then it will run snyk monitor (reports vulnerabilities to snyk.io)
-# Otherwise it will run snyk test
-name: Snyk
+# This action submits a report to snyk on every push
 
+name: Snyk
 on:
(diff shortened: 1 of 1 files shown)


NovemberTang:
> Seems that the names in Snyk aren't quite as descriptive, is there a way we could keep the V1/V2/Scala tags from before?
> 

@alinaboghiu Unfortunately not. This action picks up all the manifests in one big scoop, so theres no way to rename them individually

twrichards (approved):
although it's a shame we lose the custom sub-project names, the standardisation is definitely an improvement!! Thank you @NovemberTang - I'll merge this now

PR #1520, opened 2023-06-02
[V1] replace `JSPM` & `Grunt` with `esbuild`

twrichards:
https://trello.com/c/iYxnuAy1/1409-jspm

JSPM is a dated technology and difficult for devs to work with. It's used for V1 of facia-tool (so just config editor and breaking news notifications AFAIK) which is therefore not worked on frequently and ultimately the functionality needs to migrated to V2 (or moved into a different application). 

Importantly, we have no visibility on 

Ideally we'd use a simpler more modern toolchain for V1 - this PR attempts to do that.

## EDIT - giving up...
I've sunk too many hours into this without being able to get things loading correctly in the browser... so cutting my losses.
Instead https://github.com/guardian/facia-tool/pull/1521 is a tactical measure which moves all the build tooling (grunt, JSPM etc.) into devDependencies where they belong and adds a workaround to surface the JSPM deps. in Snyk.


twrichards:
Giving up on this, so closing PR - see EDIT in PR description.

PR #1521, opened 2023-06-13
[tactical] track JSPM deps in Snyk

twrichards:
https://trello.com/c/iYxnuAy1/1409-jspm

Following giving up on removing JSPM all-together (see https://github.com/guardian/facia-tool/pull/1520). In this PR we move all the build tooling (grunt, JSPM etc.) into `devDependencies` where they belong.

Then since Snyk can't detect JSPM dependencies, this PR introduces a workaround script which transforms the JSPM dependencies list and adds it to the top-level dependencies in a new dedicated package.json file, which the `Snyk` GitHub Action is now configured to use (the idea to do this is inspired by https://github.com/guardian/discussion-modtools/pull/812 - thanks @Mark-McCracken - although the implementation in this PR is a little simpler IMO). This looks to have had the desired effect...
image
(rather than showing zero dependencies recognised in Snyk dashboard)

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -9,10 +9,22 @@ on:
 
 
 jobs:
-  security:
+  snyk-V2-and-scala:
     uses: guardian/.github/.github/workflows/sbt-node-snyk.yml@main
     with:
       ORG: guardian
       JAVA_VERSION: 8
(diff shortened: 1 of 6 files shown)


jonathonherbert (approved):
Can see the extra deps in Snyk 👍

PR #1524, opened 2023-08-07
Two more collections for Observer/National front

jonathonherbert:
## What's changed?

Katy Vans'd like two more collections for the Observer National front. Greedy.

cc. @blishen


blishen (approved):
LGTM

PR #1525, opened 2023-09-05
upgrade v1 libs

twrichards:
because!

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
       "fastselect": "github:dbrekalo/fastselect@0.2.0",
       "font-awesome": "npm:font-awesome@4.5.0",
       "grid-util-js": "npm:grid-util-js@1.0.5",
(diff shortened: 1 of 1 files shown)


Georges-GNM (approved):
Had a poke around in CODE but couldn't find anything looking like a chart. From a glance at the code it sounds like it's used for [sparklines](https://github.com/guardian/facia-tool/blob/c327c10ecd7b22d3e61e9636c32105631334f084/public/src/js/utils/sparklines.js#L7), but I'm not 100% sure what that is either! 
There are some tests for sparklines which reference [highcharts](https://github.com/guardian/facia-tool/blob/c327c10ecd7b22d3e61e9636c32105631334f084/public/test/spec/sparklines.spec.js#L396) and which I can see are passing in the CI build logs. Between that and the fact that CODE seems to be behaving as normal, I'm inclined to say this update is ok.

PR #1528, opened 2023-10-13
Get facia tool to publish a message onto the SNS topic

silvacb:
### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


app/controllers/FaciaToolController.scala
```diff
@@ -246,11 +247,13 @@ class FaciaToolController(
 
   def pressLivePath(path: String) = AccessAPIAuthAction { request =>
     faciaPressQueue.enqueue(PressJob(FrontPath(path), Live, forceConfigUpdate = Option(true)))
+    FaciaPressTopic.publish(PressJob(FrontPath(path), Live, forceConfigUpdate = Option(true)))
```
silvacb, line 250: Using the logs, we couldn't see that this function was being invoked in the workflow that we were testing, but because, at some point, the plan is to remove the SQS code with SNS, we decided to add the call to the SNS topic publish function here as well.

app/services/FaciaPress.scala
```diff
@@ -82,6 +116,14 @@ class FaciaPress(val faciaPressQueue: FaciaPressQueue, val configAgent: ConfigAg
...
               EnqueuePressSuccess.increment()
           }
+          val fut2 = Future.traverse(paths)(path => faciaPressTopic.publish(PressJob(FrontPath(path), Draft, forceConfigUpdate = pressCommand.forceConfigUpdate)))
+          fut2.onComplete {
+            case Failure(error) =>
+              EnqueuePressFailure.increment()
+              logger.error("Error manually pressing draft collection through update from tool", error)
+            case Success(_) =>
+              EnqueuePressSuccess.increment()
+          }
```
silvacb, line 126: The `fut` and `fut2` variables only need to co-exist while this is not implemented on the frontend side. Once the frontend queue is subscribed to the topic, only the second variable will be required.

app/controllers/FaciaToolController.scala
Georges-GNM: Briefly tried following the trail, I'm not 100% sure either, but it looks like there's a view for fronts which are considered "stale"; this includes a button that enables repressing the front, which would then trigger this logic, _I think_. Not sure if this actually gets used, but seems worth adding here too!

app/services/FaciaPress.scala
Georges-GNM: Makes sense. Not blocking, but I wonder if there's a better name that can be given than `fut` (and later `fut2`) which doesn't feel very descriptive to me. If I'm understanding correctly it provides the outcome of a "send message" or "publish" action?

conf/application.conf
```diff
@@ -122,6 +122,8 @@ PROD {
   publish_events.queue_url="https://sqs.eu-west-1.amazonaws.com/163592447864/publish-events-PROD"
 }
 
+faciatool.sns.tool_topic_arn="arn:aws:sns:eu-west-1:163592447864:facia-CODE-FrontsUpdateSNSTopic-RepwK3g95V3w"
```
Georges-GNM, line 125: For some reason I'm a bit surprised this (along with the other values) is hardcoded here but I guess it's ok!

app/services/FaciaPress.scala
silvacb: Yes, I was thinking once subsequent changes are made and one of the variables is removed, a better name should be given to it.

app/services/FaciaPress.scala
```diff
@@ -82,6 +116,14 @@ class FaciaPress(val faciaPressQueue: FaciaPressQueue, val configAgent: ConfigAg
             case Success(_) =>
               EnqueuePressSuccess.increment()
           }
+          val fut2 = Future.traverse(paths)(path => faciaPressTopic.publish(PressJob(FrontPath(path), Draft, forceConfigUpdate = pressCommand.forceConfigUpdate)))
+          fut2.onComplete {
+            case Failure(error) =>
+              EnqueuePressFailure.increment()
+              logger.error("Error manually pressing draft collection through update from tool", error)
+            case Success(_) =>
+              EnqueuePressSuccess.increment()
```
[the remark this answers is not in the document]
silvacb: That sounds sensible, I will comment them out.

conf/application.conf
silvacb: Yeah, let me know if you think it should live somewhere else, but this seemed like the right place for it.

app/services/SQSQueues.scala
```diff
@@ -35,6 +39,31 @@ object SQSQueues {
   }
 }
 
+object SNSTopics {
```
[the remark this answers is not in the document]
silvacb: Yes, I think it does make sense, I'll make those changes. Thanks!

conf/application.conf
Georges-GNM: Looks like it to me, as it seems we're happy for other urls to be there!

app/services/FaciaPress.scala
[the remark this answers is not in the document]
silvacb: I've ended up removing the lines instead of commenting them out, it's less messy.

app/services/FaciaPress.scala
```diff
@@ -67,6 +91,14 @@ class FaciaPress(val faciaPressQueue: FaciaPressQueue, val configAgent: ConfigAg
             case Success(_) =>
               EnqueuePressSuccess.increment()
           }
+          val fut2 = Future.traverse(paths)(path => faciaPressTopic.publish(PressJob(FrontPath(path), Live, forceConfigUpdate = pressCommand.forceConfigUpdate)))
+          fut2.onComplete {
+            case Failure(error) =>
```
Georges-GNM, line 96: Think these lines should be removed too then!

silvacb: True, I missed this. Done now.

Georges-GNM (approved):
Confirmed from AWS and logging that editing a front in code published to the desired topic, so this looks good to me!

PR #1529, opened 2023-10-18
fronts client v2 deps update

dblatcher:
## What's changed?

https://trello.com/c/yMRC0Qmx/1786-fix-vulnerabilities-tool-v2

 - Miscellaneous dependency upgrades to reduce vulnerabilities 
 - removed the [guration](https://github.com/guardian/guration) dependency - was mentioned in the README but had already been replaced
 - The upgrades to the Babel packages triggered stricter compiling of `styled-components` - required some extra specifications of types in the source code - EG the type of Events in onChange functions 

## Implementation notes

This PR makes significant reduction in vulnerabilities, but there is more to do. The main targets area seem likely to require significant changes to source code and tests, which seemed best handled on dedicated PRs:
 - `testcafe` (integration test suite) is to source of 27/45 remaining vulnerabilities, but there are several breaking changes to address and version conflicts for upgrading jest - might be worth looking into switching to Cypress since it is more commonly used in Guardian projects
 - Using node v14, typescript 3 and webpack 4 -   should upgrade to unblock future upgrades, but didn't want too many major changes on one PR
 - `styled-components` may need an upgrade to v6+ at some point, but the way we handle theming (by overriding and re-exporting base types) no longer seems to be supported: see this issue https://github.com/styled-components/styled-components/issues/4087

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


rhystmills (approved):
Looks good to me, and running as expected locally.

Had some odd behaviour when I first tried to run - permissions errors to do with yarn and babel that were preventing the startup script from running. Deleted `node_modules` and it started working. Don't exactly understand what happened but thought it worth mentioning in case it happens again down the line.

PR #1539, opened 2024-01-03
Display front priority in the troubleshooting tool

davidfurey:
## What's changed?

Adds the front priority in parenthesis after the front name in the troubleshooting tool so that it is more apparent if you are diagnosing a lower priority issue (i.e. training front)

Screenshot 2024-01-03 at 12 54 48

Screenshot 2024-01-03 at 12 53 30

### DIFF
--- a/public/src/js/troubleshoot/views/stale.js
+++ b/public/src/js/troubleshoot/views/stale.js
@@ -174,7 +174,7 @@ function diagnoseStaleFront (container, front, priority, config, when, status) {
     const troubleshootResults = clone('staleFront');
     troubleshootResults.querySelector('.lastModifyDate').textContent = when || 'never';
     troubleshootResults.querySelector('.frontName').textContent = front;
-    troubleshootResults.querySelector('.frontPrioity').textContent = priority;
+    troubleshootResults.querySelector('.frontPriority').textContent = priority;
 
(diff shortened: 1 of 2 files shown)


Georges-GNM:
Looking at code, this seems ok with training, editorial, email and showcase fronts, however for the main editorial ones, they don't appear to have a priority set from what I can see in the config, so appears as: 
image

Considering editorial fronts would be the higher priority ones, I guess ideally the message in brackets would reflect that (or at least say editorial)?

public/src/js/troubleshoot/views/stale.js
```diff
@@ -134,17 +134,18 @@ function fetchLastPressed (front) {
 }
 
 function checkPressedState (front, config, container, lastPress) {
+    const priority = config.fronts[front].priority;
```
Georges-GNM, line 137: Maybe something like this?
```suggestion
    const priority = config.fronts[front].priority ? config.fronts[front].priority : "editorial";
```

davidfurey: Interesting, it does seem that editorial is the default priority.

Georges-GNM (approved):
Tested on code again after the latest change, looks good to me

image

PR #1534, opened 2023-11-17
simplify `press` function in `FaciaPress` and add logging to SNS publish operation

twrichards:
Co-authored-by: @Georges-GNM @silvacb 

Just some follow-up refactoring from #1528 (and #1533) to simplify/abstract the behaviour with some clearer variable names and more logging.

### TESTED on `CODE` ✅ 
image

PR #1533, opened 2023-11-17
Add prod SNSTopic

Georges-GNM:
Co-authored-by: @twrichards 
Co-authored-by: @silvacb 

## What's changed?
Now that this [platform PR](https://github.com/guardian/editorial-tools-platform/pull/707) is in prod, creating an SNS Topic, as a follow-up to this [previous PR](https://github.com/guardian/facia-tool/pull/1528) which enabled the fronts tool to publish to an sns topic, this PR specifies the prod SNS topic for the tool to point at. 

### General
- [ ] 🤖 Relevant tests added
- [ ] ✅ CI checks / tests run locally
- [ ] 🔍 Checked on CODE

### Client
- [ ] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ ] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #1540, opened 2024-01-04
remove facia press queue (in favour of SNS topic)

twrichards:
This builds on the simplification started in https://github.com/guardian/facia-tool/pull/1534, but now that we've decided to use the SNS topic (see #1528, https://github.com/guardian/editorial-tools-platform/pull/707, #1533 plus diagram in https://github.com/guardian/mobile-fastly-cache-purger) as the only way out of fronts tool and frontend is subscribing their queue to the SNS topic instead (see https://github.com/guardian/frontend/issues/26790, https://github.com/guardian/platform/pull/1569 and  https://github.com/guardian/editorial-tools-platform/pull/750) we can now remove all traces of the 'facia press queue' and further simplify the logic.

_Will need to update some diagrams (incl. main fronts diagram and https://github.com/guardian/mobile-fastly-cache-purger)._


app/services/FaciaPress.scala
```diff
@@ -5,7 +5,7 @@ import com.amazonaws.services.sns.AmazonSNSAsyncClientBuilder
 import com.amazonaws.services.sns.model.PublishResult
 import com.amazonaws.services.sqs.AmazonSQSAsyncClientBuilder
```
Georges-GNM, line 6: Nitpick, I guess these two sqs imports can be removed

Georges-GNM (approved):
Looks good to me - can see this seems to be working in code at the moment (with appropriate logs coming through) so should be good to merge 👍

PR #1545, opened 2024-02-27
Upgrade Typescript to 5.x and related dependencies

jonathonherbert:
## What's changed?

Bumps Typescript and Jest, in preparation for Vite.

As a result of changes in types, this means bumping a few other packages and/or their types, including eslint, react, redux and their bindings, and styled-components.

Because the linting style of eslint and its typescript plugin has changed, we must also include a large `lint --fix` as part of this PR. I've isolated that to a single commit to make it easier to tell apart the linting no-op from other changes.

## Implementation notes

This is a large PR! I've tried to make sure the commits tell the story. Happy to pair to review.

There are very few runtime changes outside of those versions – I've kept them scoped to [662fd39](https://github.com/guardian/facia-tool/pull/1545/commits/662fd39f0872b983cfde11ff84c0f50f281545cf) to make them easier to identify in case of problems.

Worth noting that we have good integration tests on this project for card rendering, drag and drop, etc. that should give us some confidence things are working as expected.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


fronts-client/config/webpack.config.dev.js
```diff
@@ -1,7 +1,7 @@
-import path from 'path';
-import common from './webpack.config.common.js';
+const path = require('path');
+const common = require('./webpack.config.common.js');
```
jonathonherbert, line 2: We should not be mixing and matching import styles here, so I've moved all webpack config to CommonJS.

Fweddi (approved):

PR #1546, opened 2024-02-27
Use vite

jonathonherbert:
## What's changed?

Use vite, not webpack, to eliminate what is now a _30 second_ wait for client-side dependencies to build each time a change is made.

|Before|After|
|-|-|
|download|snappy-changes|

The production build is also quicker:

Screenshot 2024-03-04 at 18 30 30

Tested locally and in CODE – both are important as local development has a slightly different configuration to enable hot reloading.

There are a fair few changes here – the commit history should give them in detail, and I've added a few comments to the PR, too.

## How to test

- Run locally with `script/dev-start.sh`, making sure you run `yarn` in the `fronts-client` directory to get the new dependency tree. Make a few visual changes. Updates should be quick after the initial build!
- Our integration tests are passing in CI, which should give us some confidence that the client is working as expected. (They cover a reasonable surface area – see `./fronts-client/integration` for specs.)
- Deploy to CODE. THe application should work as expected.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


fronts-client/integration/server/index.html
```diff
@@ -116,6 +116,6 @@
         capiPreviewUrl: '/api/preview/'
       });
     </script>
-    <script src="/dist/app.bundle.js"></script>
+    <script type="module" src="/fronts-client-v2/dist/index.js"></script>
```
jonathonherbert, line 119: Necessary for Vite's ES6 output.

scripts/ci.sh
```diff
@@ -27,7 +26,7 @@ javascriptV1() {
 
 javascriptV2() {
     pushd fronts-client
-
+    setupNvm
```
jonathonherbert, line 29: Vite expects our new version of node.

fronts-client/package.json
```diff
@@ -53,28 +50,23 @@
...
     "jsdom": "^21.1.1",
     "markdown-toc": "^1.2.0",
     "prettier": "^2.0.2",
-    "puppeteer": "^19.11.1",
+    "puppeteer": "^22.3.0",
     "react-test-renderer": "^16.13.0",
     "redux-mock-store": "^1.5.3",
     "style-loader": "^0.23.1",
-    "testcafe": "^1.1.4",
+    "testcafe": "^3.5.0",
```
jonathonherbert, line 62: Necessary to fix some odd inconsistencies between GHA and local runs. Happily, two major version bumps did not require any changes (!)

fronts-client/package.json
```diff
@@ -128,7 +120,7 @@
     "react-router-redux": "^4.0.8",
     "react-testing-library": "6.1.2",
     "react-transition-group": "^2.6.0",
-    "recharts": "^1.6.2",
+    "recharts": "^2.12.2",
```
jonathonherbert, line 123: Recharts 1.x threw up problems with Vite as it mixed CJS and ES6 imports.

fronts-client/package.json
```diff
@@ -140,12 +132,14 @@
     "styled-components": "5.3.11",
     "ts-optchain": "^0.1.2",
     "tti-polyfill": "^0.2.2",
-    "uuid": "^3.2.1"
+    "url": "^0.11.3",
```
jonathonherbert, line 135: Somehow this was missing from our dependencies, but the webpack configuration provided it (it's imported by application code) – this was confusing, but we should certainly have an explicit dependency here.

app/views/V2App/app.scala.html
```diff
@@ -32,7 +32,25 @@
...
     <div class="wrap" id="react-mount"></div>
     <div id="config" data-value="@clientConfigJson"></div>
-    <script src="@jsLocation" type="application/javascript"></script>
+    <script type="module">window.global = window</script>
+
+    @if(isDev) {
+        <!-- Connect directly to Vite dev server in development -->
+        <!-- See https://vitejs.dev/guide/backend-integration.html -->
+        <script type="module">
+            import RefreshRuntime from 'http://localhost:5173/assets/fronts-client/@@react-refresh'
```
JustinPinner, line 41: I'm not familiar with this yet, and I still have many changes to look at so this question may resolve itself but in the meantime: `fronts-client` isn't `fronts-client-v2` here?

.nvmrc
```diff
@@ -1 +1 @@
-v16.20.2
+14.17.0
```
fredex42, line 1: What's the reason for requiring a deprecated version of Node here?

fredex42 (commented):
Just a question about the Node version

JustinPinner (approved):
Not sure if my approval really matters too much here. Only one naive question from me, otherwise I think it's making sense. Will 👍 on the basis of it having been tested and shown to be working as expected but please do seek other opinions 😄

.nvmrc
jonathonherbert: This was a change that was made in #1545 and lost in a rebase – thanks for spotting! Reinstated in c243b73726ca76835fd1bc4837d6b51b34062154.

app/views/V2App/app.scala.html
jonathonherbert: 👍  this just happens to be the URL we serve V2 assets from – but since moving the `dist` assets directory from `fronts-client` to `fronts-client-v2`, this is now confusing! e9ef7f3f704c57f30fd157797dde39b462fb54a4 ensures we're using fronts-client-v2 everywhere assets are needed.

rhystmills:
The `./scripts/setup.sh` script is failing to run for me currently - in the top dir the node version specified is 16 and it looks like vite is unhappy with that:

rhystmills (approved):
Tested locally, running as expected and building speedily 🥇

PR #1562, opened 2024-04-22
Show users editions list or message based on permissions

Divs-B:
## What's changed?

At present when user tries to access [Fronts](https://fronts.code.dev-gutools.co.uk/v2) they gets list of Editions on Home page, which they are able to select but depends on permissions Editions might not get edited if not allowed.
We want Home page to not to show editions-list at all if user is not permitted instead they should get relevant message displayed on the screen.

ref: https://trello.com/c/QCmlviYl/2867-fronts-tool-does-not-check-permissions-before-permitting-access-to-editions

If this gets approval our next task will be to implement for Fronts-list.

**UI Testing:**
Tested on my email id which don't have permission to access Editions at the moment hence I will see on screen:

image

 
### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/conf/Configuration.scala
+++ b/app/conf/Configuration.scala
@@ -227,6 +227,7 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
 
   object faciatool {
     lazy val breakingNewsFront = "breaking-news"
+    lazy val editEditions = "edit-editions"
     lazy val frontPressToolTopic = getString("faciatool.sns.tool_topic_arn")
     lazy val publishEventsQueue = getMandatoryString("publish_events.queue_url")
     lazy val showTestContainers = getBoolean("faciatool.show_test_containers").getOrElse(false)
--- a/app/conf/Configuration.scala
+++ b/app/conf/Configuration.scala
@@ -227,7 +227,7 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
 
   object faciatool {
     lazy val breakingNewsFront = "breaking-news"
-    lazy val editEditions = "edit-editions"
+    lazy val canEditEditions = "edit-editions"
     lazy val frontPressToolTopic = getString("faciatool.sns.tool_topic_arn")
     lazy val publishEventsQueue = getMandatoryString("publish_events.queue_url")
     lazy val showTestContainers = getBoolean("faciatool.show_test_containers").getOrElse(false)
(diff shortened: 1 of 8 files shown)


app/controllers/DefaultsController.scala
```diff
@@ -41,9 +41,11 @@ class DefaultsController(val acl: Acl, val isDev: Boolean, val deps: BaseFaciaCo
   def configuration = AccessAPIAuthAction { implicit request =>
     val hasBreakingNews = acl.testUser(Permissions.BreakingNewsAlert, "facia-tool-allow-breaking-news-for-all")(request.user.email)
     val hasConfigureFronts = acl.testUser(Permissions.ConfigureFronts, "facia-tool-allow-config-for-all")(request.user.email)
+    val hasEditionsPermissions = acl.testUser(Permissions.EditEditions, "facia-tool-allow-edit-editorial-fronts-for-all")(request.user.email)
```
jonathonherbert, line 44: Non-blocking: is this the right name for a potential editions switch? Perhaps better as 
```suggestion
    val hasEditionsPermissions = acl.testUser(Permissions.EditEditions, "facia-tool-allow-edit-editions-for-all")(request.user.email)
```

There's no scenario I can think of where we'd need to turn this off for every user, but as it stands this referencing editorial fronts, not editions, might be confusing to developers.

app/conf/Configuration.scala
```diff
@@ -227,6 +227,7 @@ class ApplicationConfiguration(val playConfiguration: PlayConfiguration, val isP
 
   object faciatool {
     lazy val breakingNewsFront = "breaking-news"
+    lazy val editEditions = "edit-editions"
```
jonathonherbert, line 230: Non-blocking: can I suggest the following?
```suggestion
    lazy val canEditEditions = "edit-editions"
```
`breakingNewsFront` actually refers to the name of a front, as well as being a permission, but `editEditions` is not a front, and it's perhaps not clear this is being used in a permissions context.

fronts-client/src/components/Home.tsx
```diff
@@ -52,6 +55,16 @@ const Home = ({ availableEditions }: IProps) => (
...
 const mapStateToProps = (state: State) => ({
   availableEditions: selectAvailableEditions(state),
+  editEditionsIsPermitted: selectEditionsPermission(state)?.['edit-editions'],
 });
 
+const displayNoPermissionMessage = (onContent: string) => {
+  return (
+    <p>
+      You do not have permission to edit {onContent}. Please contact
+      central.production@guardian.co.uk to request access.
```
jonathonherbert, line 65: Non-blocking:

```suggestion
      <a href="mailto:central.production@guardian.co.uk">Central Production</a> to request access.
```

(or similar)

jonathonherbert (approved):
Tested locally with permissions on and off, works as expected. Lovely! A few non-blocking comments re: messaging and naming.

app/controllers/DefaultsController.scala
Divs-B: Makes sense I will change to suggested one 👍

app/conf/Configuration.scala
Divs-B: Yeah sure @jonathonherbert , 
when I started typing this variable I indeed gave "canEditEditions" name but later changed it 😄

fredex42 (approved):
Nice and succinct :)

Divs-B:
Confirmed with @jonathonherbert to proceed ahead for this PR and to inform Katy (Content Production Team) about the change.

PR #1558, opened 2024-04-09
Migrate Editions article to card, and add 'card_type' property to card model

jonathonherbert:
~_NB: Depends on #1563 – dear reader, please review that first!~ ✅ 

## What's changed?

Migrates the Editions `article` entity to `card`, and add 'card_type' property to card model. This opens the door to supporting more card types for our recipe work – `recipes`, `chefs`, and `collections`. We represent `card_type` as an enum in the Editions model.

We also rename `pageCode`, which is an article-specific identifier, to `id`, for the same reason.

Reviewing commit-by-commit might be sensible, as the renaming changes touch many files.

## Implementation notes

Things I've altered:
- the database table, and references to it
- every entity upstream of the Editions publication process

Things I've avoided:
- Collection `items`. This is really `cards`, but I'm reluctant to alter anything that might introduce changes to the publication model, yet.
- Article-related entities that are part of the Editions publication model. These really are articles 😀 (and again, I don't want to change the publication model.)

We may find that we can be bolder with the publication models after working on Recipes publication, but in the mean time, this PR is large, so this perhaps is plenty for now. 

### General
- [x] 🤖 Relevant tests ~added~ modified
- [x] ✅ CI checks / tests run locally
- [x] 🔍 Checked on CODE (I've tested creating a new edition, editing, and proofing/publishing)

### Client
- [x] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [x] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #1563, opened 2024-04-23
Do not throw errors when collection.type is not present

jonathonherbert:
## What's changed?

As part of work on #1558, I noticed that [we see a lot of errors in Sentry](https://the-guardian.sentry.io/issues/5042764688/?project=35467&query=&referrer=project-issue-stream) related to the `collection.type` property not being visible. This is always true in an Editions context, where the collection model differs slightly from Fronts, and does not include presentation information.

This PR amends the function that fetches this information, `getVisibleArticles`, to avoid fetching this data if the `type` property is not present.

## How to test

- The automated tests should pass
- The desktop/mobile markers should be present in Fronts as per, and not visible in editions:

Screenshot 2024-04-23 at 14 44 24

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/actions/Collections.ts
+++ b/fronts-client/src/actions/Collections.ts
@@ -291,9 +291,15 @@ function updateCollection(
         getState(),
         frontStages.draft
       );
-      dispatch(
-        recordVisibleArticles(collection.id, visibleArticles, frontStages.draft)
-      );
+      if (visibleArticles) {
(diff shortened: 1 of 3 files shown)


fredex42 (approved):
👍 Always better when frontend expects what backend is sending 😁

PR #1564, opened 2024-04-23
Add query logging to local postgres

jonathonherbert:
## What's changed?

Adds query logging to the local postgres container, so we can see what the DB is doing in more detail:

Screenshot 2024-04-23 at 18 16 04

This only applies to local development, so it should be safe.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fredex42 (approved):
🌳

PR #1565, opened 2024-04-24
Publication chain for Feast fronts

fredex42:
**Note** Requires https://github.com/guardian/editorial-tools-platform/pull/770 as well

## What's changed?

Previously, the Editions workflow only targeted one product - the Editions app.  This meant outputting "standard" issue fronts into S3 from where they are picked up by the Editions app backend.

Now, we are extending the UI to use the same process for the Feast app.  This has a different requirement; the front must have its format converted into the form that the Feast app expects and it needs to be sent over to the Feast app backend via SNS.

Another difference is in "proofing".  This is an integral part of the Editions publication process, but is not (currently) supported by the Feast app.  In-app support is necessary for it to work.  An extra flag has therefore been added to the Edition object, `requiresProofing`, which tells both frontend and backend if the Proofing stage is required.  

- When this is `true` (always the case for Editions app; the logic here sets the flag to `true` if the template is defined in the list of Editions app templates) then the UI shows the `Proof` button, proofing is required to publish and the Publish action publishes the last proofed version.
- When this is `false` (currently true for non-Editions app templates) then the UI does not show the `Proof` button and the proofing process is short-circuited to directly output the current state of the Edition.
- There is therefore no change in the Editions app workflow, and the Proofing stage is effectively hidden for Feast

This PR makes the relevant changes, by extracting the generic parts of the publication chain into Traits, which are then extended by concrete implementations for both Editions and Feast.

### How will this data be used?

See https://github.com/guardian/recipes-backend/pull/36

## How to test in CODE

- Create a new Feast edition
- Add some recipes to the `All Recipes` and `Meat-Free` fronts
- Hit `Check` to ensure that it's all ready to go
- Hit `Publish`. Make a mental note of the time. You should see the status is `Started` (currently this won't change as the feedback loop is not present)
- Log into the CMS Fronts account via AWS console
- Open up the SQS console
- Go to the topic `feast-facia-test`. This is subscribed to the SNS topic that the app writes to
- Click `Send and Receive Messages`
- In the lower pane, click `Poll for messages`
- You should see a message appear with a timestamp about when you hit 'Publish'
- Open it up to inspect. Drag the text box so you can see the whole thing
- You should see your front as Feast-style data, double-encoded as a JSON object within the `Message` field. Something like this:
```
  "Message" : "{\"all-recipes\":[{\"id\":\"f36ee25d-54f5-4b63-9998-df8cb424f13f\",\"title\":\"Dish of the day\",\"body\":\"\",\"items\":[{\"recipe\":{\"id\":\"14049060\"}}]},{\"id\":\"fa9d34d5-ffa4-4caf-9647-2e785f0f7a89\",\"title\":\"Collection 2\",\"body\":\"\",\"items\":[{\"recipe\":{\"id\":\"14002695\"}},{\"recipe\":{\"id\":\"13891262\"}}]},{\"id\":\"6ab6b4a9-f02a-435c-999c-96145eb383e7\",\"title\":\"Collection 3\",\"body\":\"\",\"items\":[{\"recipe\":{\"id\":\"14077819\"}}]}],\"meat-free\":[{\"id\":\"45be31c5-931c-4c85-a77d-86ebff3eaf0a\",\"title\":\"Dish of the day\",\"body\":\"\",\"items\":[{\"recipe\":{\"id\":\"13898146\"}},{\"recipe\":{\"id\":\"13891262\"}}]},{\"id\":\"1e9e56ea-cc91-4ac6-8847-11fcf2b6c091\",\"title\":\"Collection 2\",\"body\":\"\",\"items\":[{\"recipe\":{\"id\":\"13726752\"}},{\"recipe\":{\"id\":\"14048612\"}}]}]}",
```

Note the Feast-style formatting, article internal IDs in place of recipe IDs

## Implementation notes

- The `EditionsPublishing` class (containing generic publication code) has been renamed to simply `Publishing`
- `EditionsPublishing` used to directly reference `EditionsBucket` instances for pushing to Preview and Live locations.  `EditionsBucket` now inherits from a `PublicationTarget` trait, in common with the new `FeastPublicationTarget` implementation
- A new `PublicationTargetWithTransform` layer is also put in, that represents any potential publication target which needs to have the data converted from the Fronts format into 'something' else
- The transform is implemented for Feast (though this is a dummy implementation at present, because the relevant data is not yet in the upstream Fronts tool)

### Held over for future PRs

- It seems strange to have strongly-typed output methods for PublicationTargets alongside one which just takes a `String` (which has been JSON-encoded by the caller).  Changing this is out of scope for this PR, however.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client
  - Note, there is a warning on initial load of editorial fronts `404 accessing https://fronts.code.dev-gutools.co.uk/undefined`.  This is also present on main branch so leaving it out of scope here. 


app/services/editions/publishing/EditionsBucket.scala
```diff
@@ -20,20 +20,24 @@ object EditionsBucket {
...
+    putIssueJson(issue, key)
+  }
+
+  override def putIssueJson[T: Writes](content: T, key:String): Unit = {
+    val request = EditionsBucket.createPutObjectRequest(bucketName, key, content)
     s3Client.putObject(request)
   }
 
-  def putEditionsList(rawJson: String) = {
+  def putEditionsList(rawJson: String): Unit = {
```
fredex42, line 40: Hmmmmm thinking that this could be redundant, can we use `putIssueJson` instead?

app/services/editions/publishing/PublicationTarget.scala
```diff
@@ -0,0 +1,27 @@
+package services.editions.publishing
+
+import model.editions.PublishableIssue
+import play.api.libs.json.Writes
+import PublishedIssueFormatters._
+
+trait PublicationTargetBase[T] {
+  def putIssue(issue: PublishableIssue, key:String = "")(implicit evidence: Writes[T]): Unit = {
+    putIssueJson(issue, "")
```
fredex42, line 9: FIXME!! Should be `key` not hard-coded to empty string

app/controllers/EditionsController.scala
```diff
@@ -86,7 +86,8 @@ class EditionsController(db: EditionsDB,
 
   def republishEditionsAppEditionsList = EditEditionsAuthAction { _ => {
     try {
-      val raw = Json.toJson(Map("action" -> "editionList")).as[JsObject] + ("content", Json.toJson(getAvailableEditionsAppEditions))
+      //TODO: Make this a case class and serialise it properly
```
jonathonherbert, line 89: Still pending, or a future PR?

app/controllers/EditionsController.scala
```diff
@@ -190,7 +191,7 @@ class EditionsController(db: EditionsDB,
   def publishIssue(id: String, version: EditionIssueVersionId) = EditEditionsAuthAction { req =>
     val lastProofedIssueVersion = db.getLastProofedIssueVersion(id)
     // Protect against stale requests.
-    if (lastProofedIssueVersion.filter(_.equals(version)).isEmpty) {
+    if (version != Publishing.ProofingNotRequiredMagicVersion && !lastProofedIssueVersion.exists(_.equals(version))) {
```
jonathonherbert, line 194: The issue knows whether it supports proofing, or not (`issue.supportsProofing`?) – could we fetch the issue and use that to make this decision, rather than relying on a magic version?

If so, we could do the same thing instead of the magic version check in Publishing.scala.

app/services/editions/publishing/FeastPublicationTarget.scala
```diff
@@ -0,0 +1,42 @@
...
+      ).asJava)
+  }
+
+  override def putIssueJson[T:Writes](issue: T, key:String): Unit = {
+    val content = Json.stringify(Json.toJson(issue))
+    snsClient.publish(createPublishRequest(content, FeastPublicationTarget.MessageType.Issue))
+  }
+
+  override def putEditionsList(rawJson: String): Unit = {
+    snsClient.publish(createPublishRequest(rawJson, FeastPublicationTarget.MessageType.EditionsList))
```
jonathonherbert, line 40: Ooh, are we planning to do anything with this?

app/services/editions/publishing/PublicationTarget.scala
```diff
@@ -0,0 +1,30 @@
...
+  def putIssue(issue: PublishableIssue, key:String = "")(implicit evidence: Writes[T]): Unit = {
+    putIssueJson(issue, key)
+  }
+
+  protected def putIssueJson[C: Writes](content: C, key:String): Unit
+
+  //FIXME: It seems strange and awkward to have one method which puts structured data and another that does the same
+  //but with an arbitrary string, making it unchecked.  This can probably be pulled into `putIssueJson` but I think that
+  // is out of scope for this PR
+  def putEditionsList(rawJson: String): Unit
```
jonathonherbert, line 17: 👍 👍

app/services/editions/publishing/PublicationTransform.scala
```diff
@@ -0,0 +1,12 @@
+package services.editions.publishing
+
+import model.editions.PublishableIssue
+import play.api.libs.json.Writes
+
+/**
+ * PublicationTransform is a generic interface that takes a PublishableIssue and turns it into
+ * some other shape, which must be JSON-serializable
+ */
+trait PublicationTransform[O] {
```
jonathonherbert, line 10: Should this be a polymorphic trait – or, put another way, is this something the publication target hierarchy needs to know about?

AFAICS it only happens in one place on `putIssue` – given that the `PublicationTargetBase` trait only has two methods, only one of which implicates an issue, could we not just do the transform locally in the recipe publication target, and save ourselves some polymorphism complexity and this trait?

IIUC (may be missing something) I'd be tempted to wait for another use case before adding a level of indirection here.

app/services/editions/publishing/Publishing.scala
```diff
@@ -0,0 +1,100 @@
...
+import model.editions.Edition.{FeastNorthernHemisphere, FeastSouthernHemisphere}
+import model.editions.{EditionsIssue, PublishAction}
+import net.logstash.logback.marker.Markers
+import services.editions.db.EditionsDB
+import play.api.libs.json.Writes
+
+import scala.jdk.CollectionConverters._
+
+class Publishing[+A:Writes, +B:Writes](editionsAppPublicationBucket: PublicationTargetBase[A],
+                 editionsAppPreviewBucket: PublicationTargetBase[A],
```
jonathonherbert, line 15: Could these be, concretely, `EditionsBucket`, `EditionsBucket`, and `FeastPublicationTarget`? Struggling to see how these type params will be exercised in a useful way, but may be missing something!

app/services/editions/publishing/Publishing.scala
```diff
@@ -0,0 +1,100 @@
...
+                 feastAppPublicationTarget: PublicationTargetBase[B],
+                 db: EditionsDB
+                ) extends Logging {
+
+  def updatePreview(issue: EditionsIssue) = {
+    val previewIssue = issue.toPreviewIssue
+    // Archive a copy
+    issue.edition match {
+      case FeastNorthernHemisphere | FeastSouthernHemisphere =>
+        //Feast does not currently support previewing, but we don't need to be reminded of that every time someone drag/drops something!
```
jonathonherbert, line 25: 👍 👍

app/services/editions/publishing/transform/FeastAppModel.scala
```diff
@@ -0,0 +1,35 @@
...
+  case class FeastAppContainer(id:String, title:String, body:Option[String], items:Seq[ContainerItem])
+  type FeastAppCuration = Map[String, IndexedSeq[FeastAppContainer]]
+
+  implicit val recipeIdentifierFormat:Format[RecipeIdentifier] = Json.format[RecipeIdentifier]
+  implicit val recipeFormat:Format[Recipe] = Json.format[Recipe]
+  implicit val chefFormat:Format[Chef] = Json.format[Chef]
+  implicit val paletteFormat:Format[Palette] = Json.format[Palette]
+  implicit val subCollectionFormat:Format[SubCollection] = Json.format[SubCollection]
+
+  implicit val containerItemFomat:Format[ContainerItem] = Format.apply(
```
jonathonherbert, line 22:
```suggestion
  implicit val containerItemFormat:Format[ContainerItem] = Format.apply(
```

conf/logback.xml
```diff
@@ -16,6 +16,9 @@
     </appender>
 
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <!--<encoder>
```
jonathonherbert, line 19: No longer needed?

jonathonherbert (commented):
Testing this afternoon – a few comments on the code as I do that!

jonathonherbert:
Tested in CODE, I can see publish messages on the queue 🎉 

Before we merge: there's an odd error when editing collections in Editions App issues: S3 is spitting out errors at preview, giving ['request too big'.](https://logs.gutools.co.uk/s/editorial-tools/app/discover#/doc/67604290-baa6-11e9-bea2-633437abb232/logstash-ed-tools-cms-fronts-2024.05.01?id=cI3YNY8Bxbj-ikXcPAUX) I've deployed `main` and don't see these errors. Happy to help investigate tomorrow.

app/services/editions/publishing/FeastPublicationTarget.scala
fredex42: Yes my next stop will be the recipes-backend to build the receiver for it 😁

app/services/editions/publishing/Publishing.scala
fredex42: Yes, they could TBH

fredex42:
S3 errors now appear to be fixed

app/services/editions/publishing/PublicationTransform.scala
fredex42: Yeah maybe I was over-engineering it 🤷

app/controllers/EditionsController.scala
fredex42: Done 😁

jonathonherbert (approved):
Tested proofing and publishing again against Editions in CODE, LGTM 👍

PR #1566, opened 2024-04-25
Improved build process

fredex42:
## What's changed?

- Removes the old ci.sh build script
- Replaces it with paralleling Github actions
- Removes the deprecated SBT riffraff plugin in favour of most recent GHA version
- Build on Java11, compile Scala optimised for java 11
- Update the target AMI to Java11 as well

## Implementation notes

None

## Checklist

Things that still need checking:

- [x] ~~There are errors being reported of files not found during the frontend v2 integration tests. These do not fail the tests and I have a suspicion that they relate to the Webpack->Vite migration but not sure.~~ [FIXED in https://github.com/guardian/facia-tool/pull/1568]
- [x] I'm not familiar enough with the frontend to be able to check everything, but what I do know about looks OK

### General
- [x] 🤖 Relevant tests added               <-- n/a
- [x] 🔍 Checked on CODE                   

### Client
- [ ] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work) <-- partially checked
- [x] 📷 Screenshots / GIFs of relevant UI changes included [None]


jonathonherbert:
This looks great! A big improvement.

A bit of work to do on the v1 client, as currently it's throwing an error:
Screenshot 2024-04-26 at 09 52 20
Unsure off top of head of a fix, but v. happy to pair.

fredex42:
@jonathonherbert thanks for that. Looks like it's not bundling the JSPM deps...

fredex42:
> @jonathonherbert thanks for that. Looks like it's not bundling the JSPM deps...

Fixed this morning :)

build.sbt
```diff
@@ -10,40 +10,26 @@ packageDescription := "Guardian front pages editor"
 
 ThisBuild / scalaVersion := "2.13.5"
 
-import com.gu.riffraff.artifact.BuildInfo
 import sbt.Resolver
 
-debianPackageDependencies := Seq("openjdk-8-jre-headless")
+debianPackageDependencies := Seq("java11-runtime-headless")
```
jonathonherbert, line 15: Just FMI as I'm sure it is – why is this necessary? Doesn't the AMI provide a Java runtime?

.github/workflows/improvedci.yml
```diff
@@ -0,0 +1,269 @@
...
+        permissions:
+            id-token: write
+            contents: read
+        needs:
+            - build_client_v1
+            - build_client_v2
+            - generate_build_number
+        services:
+            # See https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers
+            postgres:
```
jonathonherbert, line 194: Is this necessary for the build? (No tests exercising a local DB?)

jonathonherbert (approved):
Tested:
- [x] V2 client (fronts edits, proof an edition)
- [x] V1 client (config, treats)
- [x] Breaking news (sent notification)

Bloomin' lovely. A few non-blocking comments.

build.sbt
fredex42: _Technically_ it's not necessary - but it's good practise when building a java-based DEB to have a generic JRE dependency

.github/workflows/improvedci.yml
fredex42: Ah, probably not 😆

build.sbt
davidfurey: Andy is correct that technically it isn't necessary. But if this line is present, it needs to be correct. AFAIK if you left this as openjdk-8-jre-headless the package would fail to install.

PR #1570, opened 2024-04-29
Add a recipe card type to the Feast platform

jonathonherbert:
_co-authored-by: @Divs-B_ 

## What's changed?

Adds a recipe card type to the Feast platform. It's useable when you're editing Feast editions.

We've added some fixture data that's visible in the Feed on the left hand side to get us started. This will eventually be powered by recipe-backend search.

We've added the minimum possible to test this from client to server and back again:
- ✅ add a new recipe `CardType` to the server-side enum
- ✅ recipe feed cards in client
- ✅ support for recipe drag-and-drop behaviour
- ✅ correctly persisting card type information on collection update
- ❌ recipe card representation in collections (in the gif, note the blank cards in the Front) – this will come in a future PR
- ❌ the hover links to the external resource, and Ophan, don't work. The hover button components need a refactor for this to work, and we thought this'd be better as a separate PR.

Here's a GIF of it in action. The UI is incomplete, but proves that we can see recipes in the feed, drag them into collections, and have them persist (note the page refresh – the data loads and is correct.) We don't have any users yet, so we are able to ship work in this state 🎨 🔥 

add-recipes

I've checked that cards in Fronts and Editions for articles, snaplinks etc. work as normal.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE (proofed Edition to prove publication is still ok 👌)

### Client


emdash-ie (approved):
Looks good to me! I haven’t run it, but I’m convinced by the gif and tests.

app/services/editions/db/IssueQueries.scala
```diff
@@ -107,21 +107,26 @@ trait IssueQueries extends Logging {
...
+        created_by,
+        created_email,
+        launched_on,
+        launched_by,
+        launched_email
+      FROM edition_issues
+      WHERE issue_date BETWEEN $dateFrom AND $dateTo AND name = ${edition.entryName}
+      """.map(EditionsIssue.fromRow(_)).list().apply()
+
+    maybeIssues.partitionMap(identity) match {
```
emdash-ie, line 126: This strikes me as a bit strange: I’m used to haskell, where `partition :: [Either a b] -> ([a], [b])`, so I had assumed this expression could be replaced by partition in scala as well. However, scala’s partition is instead roughly `partition: (List[A], A => Boolean) => (List[A], List[B])`, so I can’t see a nicer way to do this. I guess what we’re really looking for here is [Validated](https://typelevel.org/cats/datatypes/validated.html), maybe this would be nicer if using that. (Potentially a big change though!)

emdash-ie: I think leave it as-is here! I mostly mentioned it just to bring it to your attention :)

PR #1571, opened 2024-04-30
Add recipe cards to collections and clipboard

jonathonherbert:
## What's changed?

Following on from #1570, this PR adds a visual representation of the recipe card to the clipboard and issue collections.

We can now drag recipes into collections and clipboard, and they should appear as in the screenshot.

Screenshot 2024-05-01 at 21 54 52

The cards are fairly plain at the moment, just some text to indicate that they're recipes, their difficulty level, their title, an image. We can add other information as and when – it's a simple change, and is probably best discussed with Feast.

## Implementation notes

At the moment, we're storing all our recipes data in a single reducer, `recipes`. This is used by the feed, collection and clipboard, and it's a different strategy to the way CAPI data is stored, where the feed has a separate store for CAPI preview/live (`feed`), and collections has another store (`externalArticles`).

I don't think there's a good reason to split these stores, so I've kept us to just the one for now. We can revisit if there proves to be a good reason to do this.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


emdash-ie (approved):
This looks good to me, but I can’t currently run it locally to test (lack of permissions). The PR checklist is currently blank: is that deliberate?

PR #1584, opened 2024-05-15
Store EditionsClientCard model for editions clipboard data

jonathonherbert:
## What's changed?

Store editions clipboard data as `EditionsClientCard`, not `Trail`, which means the `cardType` property will be persisted. At the moment, it's not appearing, and as a result recipe cards stored in the clipboard are blank.

|Before|After|
|--|--|
|Screenshot 2024-05-21 at 17 25 59|Screenshot 2024-05-21 at 17 25 19|

## Implementation notes

The clipboard is the only place where the Fronts' `Trail`, and the Editions' `EditionsClientCard` (both of which represent cards in those domains) coexist. So it was necessary to add a new model, `ClipboardCard`, which can contain either of these models.

Because `Trail` does not belong to the `facia-tool` project, it is not possible to create a sealed trait of both of these case classes, and because we are not running Scala 3, a union type is not possible. I've used `Either`, taking its 'this or that' semantic at face value, rather than resorting to adding a library like Shapeless and using something like [`CoProduct`](https://www.scala-exercises.org/shapeless/coproducts) – I hope the intent is clear. We can circle back if there are strong opinions about this.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/UserDataController.scala
+++ b/app/controllers/UserDataController.scala
@@ -31,7 +31,6 @@ class UserDataController(frontsApi: FrontsApi, dynamoClient: DynamoDbClient, val
   private lazy val userDataTable = Table[UserData](config.faciatool.userDataTable)
 
   private def updateClipboardContentByFieldName[T](maybeJson: Option[JsValue], userEmail: String, fieldName: String)(implicit dynamoFormat: DynamoFormat[T], jsonFormat: Reads[T]) = {
-    val a = maybeJson.map(_.validate[T])
     maybeJson.map(_.validate[T]) match {
       case Some(JsSuccess(model, _)) =>
         Scanamo(dynamoClient).exec(userDataTable.update(UniqueKey("email" === userEmail), set(fieldName, model)))
--- a/app/controllers/UserDataController.scala
+++ b/app/controllers/UserDataController.scala
@@ -1,20 +1,17 @@
 package controllers
 
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
-import com.gu.facia.client.models.{Trail, TrailMetaData}
+import com.gu.facia.client.models.Trail
+import model.editions.{CardType, EditionsClientCard}
 import org.scanamo._
 import org.scanamo.syntax._
 import model.{FeatureSwitch, FeatureSwitches, UserData}
 import org.scanamo.generic.auto.genericDerivedFormat
 import org.scanamo.query.UniqueKey
-import play.api.libs.json.{JsArray, JsValue, Json}
-import model.{UserData}
-
-import play.api.libs.json.JsValue
+import play.api.libs.json.{JsError, JsResult, JsSuccess, JsValue, Json, JsonValidationError, Reads}
 import services.FrontsApi
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient
 
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.ExecutionContext
 import org.scanamo.generic.semiauto._
 
 import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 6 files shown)


fredex42:
Hmmmmmmmm.  I see the logic of using `Either` here, for its simplicity and being well-understood, but it is a bit of a short-term fudge (just because there are two _now_, doesn't mean we might not want more in the future).  `CoProduct` seems a much more logical choice IMO, but I take the point that it may represent more complexity/unfamiliarity than is warranted right now (and I think it's a safe assumption that upgrading to Scala3 is out of the question for a while, given the size of the codebase!).

I'm not scala expert though and would welcome an eye from someone who has more direct experience of shapeless etc. - @emdash-ie @JustinPinner what do you think?

fredex42:
It might be worth adding some notes to the doc about the logic for the differences between `com.gu.facia.client.models.Trail` (from the path I assume it's a Thrift model) and `model.editions.EditionsClientCard` (from the path I assume it's _not_ a Thrift model) e.g. why is one Thrift and one not, what are the field differences between them, why can one not inherit from the other, etc.

fredex42 (approved):
Does what it says on the tin 😀 

I've left my 2p worth on the `Either`/`CoProduct`/something question in the comments... I don't think it's a blocker but I think it's worth a conversation with the rest of the team

emdash-ie:
I’m totally happy with this use of Either, and I don’t think using shapeless’ coproduct type would be a good idea: it’s far too awkward to work with for direct use. There’s nothing (beyond a dislike of boilerplate) stopping us from wrapping these in an extra ADT, which is how we can evolve this when we add a third type of thing:

```scala
object ClipboardCard {
  def apply(trail: Trail): ClipboardCard = TrailCard(Left(trail))
  def apply(editionsCard: EditionsClientCard): ClipboardCard = EditionsCard(Right(editionsCard))

  val reads: Reads[ClipboardCard] =
    JsPath.read[EditionsClientCard].map(ClipboardCard.apply) or
      JsPath.read[Trail].map(ClipboardCard.apply) // presumably this could use TrailCard and EditionsCard directly, but I don’t know how offhand

  // can we even get this and the Reads automatically once we’re using an ADT? Feels like it should be possible
  val writes =  new Writes[ClipboardCard] {
    override def writes(o: ClipboardCard): JsValue =
      o match {
        case TrailCard(trail) => Json.toJson(trail),
        case EditionsCard(editionsCard) => Json.toJson(editionsCard)
      }
  }

  implicit val format = Format(reads, writes)
}

sealed trait ClipboardCard
case class TrailCard(card: Trail) extends ClipboardCard
case class EditionsCard(card: EditionsClientCard) extends ClipboardCard
```

app/controllers/UserDataController.scala
```diff
@@ -28,35 +25,30 @@ class UserDataController(frontsApi: FrontsApi, dynamoClient: DynamoDbClient, val
...
-  private def updateClipboardContentByFieldName(articles: Option[JsValue], userEmail: String, fieldName: String) = {
-    val clipboardArticles: Option[List[Trail]] = articles.flatMap(jsValue =>
-      jsValue.asOpt[List[Trail]])
-
-    clipboardArticles match {
-      case Some(articles) => {
-
-        Scanamo(dynamoClient).exec(userDataTable.update(UniqueKey("email" === userEmail), set(fieldName, articles)))
+  private def updateClipboardContentByFieldName[T](maybeJson: Option[JsValue], userEmail: String, fieldName: String)(implicit dynamoFormat: DynamoFormat[T], jsonFormat: Reads[T]) = {
+    val a = maybeJson.map(_.validate[T])
```
JustinPinner, line 34: This `val a` looks suspiciously like a leftover debugging aid. Because I do it all the time :)

jonathonherbert: Yep, getting a line location so I can have the debugger produce a result (macros mean evaluating in-band does not work) – unsure if there's another way to do this – thanks for spotting!
```suggestion
```

jonathonherbert: Removed in [6eb392c](https://github.com/guardian/facia-tool/pull/1584/commits/6eb392c490ab62aaffc0c3609e0792162212dbe0).

jonathonherbert:
Thanks for the comments, all. @fredex42 yes, the third thing breaks this! @emdash-ie, that's a nice way of solving for that, and thanks for going to the trouble of writing out an example. I think in the back of my mind is the idea that I'd expect the number of entities to _shrink_, not grow, as we converge on a single representation of a trail/card across products, but this is a) a hunch and b) a hostage to fortune 😀 

Sounds like we're happy enough to merge as-is, but if there's any additional feedback, let me know.

PR #1585, opened 2024-05-21
Add an (incomplete) form to the Chef card

jonathonherbert:
## What's changed?

Adds an (incomplete) form to the Chef card.

chef form

I've only included one field for now – bio – as a reference implementation of the general card for will be useful in parallelising work.

## How to test

Add a chef card to a container or clipboard, and edit the meta:
 - It should not be possible to save a form that has not been edited.
 - It should be possible to save a form that has been edited. The override should appear in the chef card.
 - Refreshing the page should show that the data has been persisted.
 - Hitting the 'rewind' button above the 'bio' field, once it's been edited, should return it to its original state.

## Implementation notes

It'd be nice to introduce a sum type that reflected the different meta available on cards of different types.

However, this touches many things on server and client, as everything that reasons about meta now needs to know it's dealing with the right type. So to keep the PR small, I've just added fields to the card meta types. There's a question as to whether the above is worth the engineering effort, but at an initial glance it seems sensible.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


app/model/editions/EditionsCard.scala
```diff
@@ -43,13 +42,14 @@ case class CardMetadata(
   replaceImage: Option[Image],
   overrideArticleMainMedia: Option[Boolean],
   coverCardImages: Option[CoverCardImages],
-  promotionMetric: Option[Double]
+  promotionMetric: Option[Double],
+  bio: Option[String] = None
```
Divs-B, line 46: I think good to add under the direct list of `CardMetaData` as we have only "bio" for Chef card for now. Once we have more parameters to come under Chef and Recipe then may be we can have chefMetaData and RecipeMetaData?

Divs-B (approved):
Looks great to me.
Tested locally and working as expected.
Code change looks like code to be more structured when welcoming chef meta.

PR #1588, opened 2024-05-24
Add ui to switch between recipes and chefs feeds

Divs-B:
## What's changed?

This will add radio buttons to switch between recipes and chefs feeds. 
Based on selection the result will display accordingly.
At present data for both recipes and chefs are based on fixture data and will be replaced by live items soon.

ref: https://trello.com/c/GlwlhweJ/2935-add-ui-for-switching-between-different-searches
Note: Search functionality is not done in this PR, will be covered in next task.

https://github.com/guardian/facia-tool/assets/17780995/a38de0e9-8a25-4a65-9261-a038e7e97b7d

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/feed/RecipeSearchContainer.tsx
+++ b/fronts-client/src/components/feed/RecipeSearchContainer.tsx
@@ -2,7 +2,7 @@ import ClipboardHeader from 'components/ClipboardHeader';
 import TextInput from 'components/inputs/TextInput';
 import ShortVerticalPinline from 'components/layout/ShortVerticalPinline';
 import { styled } from 'constants/theme';
-import React from 'react';
+import React, { useState } from 'react';
 import { connect } from 'react-redux';
 import { selectors as recipeSelectors } from 'bundles/recipesBundle';
 import { selectors as chefSelectors } from 'bundles/chefsBundle';
@@ -13,6 +13,7 @@ import { SearchResultsHeadingContainer } from './SearchResultsHeadingContainer';
 import { SearchTitle } from './SearchTitle';
 import { RecipeFeedItem } from './RecipeFeedItem';
 import { ChefFeedItem } from './ChefFeedItem';
+import { RadioButton, RadioGroup } from '../inputs/RadioButtons';
 
 const InputContainer = styled.div`
   margin-bottom: 10px;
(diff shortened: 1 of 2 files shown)


fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -38,34 +39,53 @@ const FeastSearchContainerComponent = ({
...
-          key={chef.id}
-          chef={chef}
-        /> /*TODO: as of now, added rendering of chefs just after the recipes. Need
-      to change when "search-chefs" action gets finalised*/
-      ))}
-    </FixedContentContainer>
-  </React.Fragment>
-);
+}: Props) => {
+  const [selectedOption, setSelectedOption] = useState<string>('recipeFeed');
```
fredex42, line 43: Tiny point - I've found that if you specify a default value like this, Typescript can infer the type and you don't have to specify it directly with `<string>`

fredex42: My preference for this is to use numbers, defined as enums (https://www.typescriptlang.org/docs/handbook/enums.html); reason is that it's quicker for the machine to compare and process numbers than strings.  However these days I guess that the performance difference is negligible though I would encourage the use of enums or constants over "magic value" strings because if you put a typo in an enum value the compiler complains wheras if you put a typo in a "magic value" string the app breaks

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -38,34 +39,53 @@ const FeastSearchContainerComponent = ({
...
+    <React.Fragment>
+      <InputContainer>
+        <TextInputContainer>
+          <TextInput placeholder="Search recipes" displaySearchIcon />
+        </TextInputContainer>
+        <ClipboardHeader />
+      </InputContainer>
+      <RadioGroup>
+        <RadioButton
+          checked={selectedOption === 'recipeFeed'}
```
fredex42, line 57: See my point above about defining `'recipeFeed'` as an enum value

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -38,34 +39,53 @@ const FeastSearchContainerComponent = ({
...
+          </SearchTitle>
+        </SearchResultsHeadingContainer>
+        {selectedOption === 'recipeFeed'
+          ? Object.values(recipes).map((recipe) => (
+              <RecipeFeedItem key={recipe.id} recipe={recipe} />
+            ))
+          : Object.values(chefs).map((chef) => (
+              <ChefFeedItem key={chef.id} chef={chef} />
+            ))}
+      </FixedContentContainer>
```
fredex42, line 85: I think that this is fine for the time being, but the ternary syntax doesn't extend very nicely if we want to put a third option in.  To be honest my preference here would be to use a second functional component in the file which contains a `switch` block on the `selectedOption` value (passed over as a param) but I wouldn't bother making that change now unless we know that we _are_ going to need a third case :grin:

fronts-client/src/components/feed/RecipeSearchContainer.tsx
Divs-B: Makes more sense. updated with enum. thanks @fredex42

fronts-client/src/components/feed/RecipeSearchContainer.tsx
Divs-B: Yeah changed to switch case, thanks. This will help in scalability.

jonathonherbert (approved):
Looks great! 👏

PR #1589, opened 2024-06-03
Add overrides for the Chef card

jonathonherbert:
## What's changed?

Adds the missing overrides for the Chef card – image and palette overrides. The palette data is copied from mis-en-place, the tool that Fronts will eventually replace, [here.](https://github.com/guardian/ios-feast/blob/main/MiseEnPlace/FrontBuilder/Palettes/Collections/CollectionPalette.swift)

chef-overrides

## Implementation notes

- I've added the collection data in this PR, as it was a similar shape and it seemed to make sense to do the lot at once.
- It'd be much nicer to have ADTs that reflected the fact that cards with different `cardType` values have different types of metadata. I've left that for now, as this PR was already large, but it'd be good to do this in another PR on both client and server.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fronts-client/src/components/form/PaletteForm.tsx
```diff
@@ -0,0 +1,197 @@
+import React, { useCallback } from 'react';
+import { noop } from 'lodash';
+
+import {
+  ChefPalette,
+  ChefPaletteId,
+  CustomPalette,
+  chefPalettes,
+} from 'constants/feastPalettes';
+import { styled } from 'constants/theme';
```
JustinPinner, line 10: Noob Q: is `PaletteForm.tsx` unique for Chef-based work? Do we imagine it being more widely adopted later? Just wondering about the naming convention really.

fronts-client/src/constants/feastPalettes.ts
```diff
@@ -0,0 +1,278 @@
...
+    darkForegroundHex: '#ACD8E5',
+    darkBackgroundHex: '#363632',
+    imageURL: 'https://uploads.guim.co.uk/2024/03/21/Picnic.png',
+  },
+  Potatos: {
+    lightForegroundHex: '#99614B',
+    lightBackgroundHex: '#F9F9F5',
+    darkForegroundHex: '#F0C5B4',
+    darkBackgroundHex: '#363632',
+    imageURL: 'https://uploads.guim.co.uk/2024/03/21/Potatos.png',
```
JustinPinner, line 192: Nitpick #1: Potato**e**s :)

fronts-client/src/constants/feastPalettes.ts
```diff
@@ -0,0 +1,278 @@
...
+    darkForegroundHex: '#FCF1F0',
+    darkBackgroundHex: '#363632',
+    imageURL: 'https://uploads.guim.co.uk/2024/03/21/Traybake.png',
+  },
+  'TV snacks': {
+    lightForegroundHex: '#427585',
+    lightBackgroundHex: '#F9F9F5',
+    darkForegroundHex: '#ACD8E5',
+    darkBackgroundHex: '#363632',
+    imageURL: 'https://uploads.guim.co.uk/2024/03/21/TVsbacks.png',
```
JustinPinner, line 269: Nitpick #2: TVs**n**acks?

fronts-client/src/constants/feastPalettes.ts
jonathonherbert: ikr? not my images 😅

fronts-client/src/constants/feastPalettes.ts
jonathonherbert: as above, but we could always rename 'em, it's somewhat irritating

fronts-client/src/components/form/PaletteForm.tsx
jonathonherbert: I think parts of this file will be used by the Collection palette work, but Collection palettes have a slightly different format – they have pictures, and a light and dark palette.

So, suspect that we'll have two forms here, and some of the components contained in this file will be split between them, if that makes sense.

Happy to change something if we reckon there might be a better name here.

fronts-client/src/types/Collection.ts
```diff
@@ -98,9 +98,17 @@ type CardMeta = CardRootMeta & {
   supporting?: string[];
 };
 
-type ChefCardMeta = {
-  bio: string;
-};
+interface ChefCardMeta {
+  bio?: string;
+  foregroundHex: string;
```
Divs-B, line 103: Reading others in this schema, do we want foregroundHex, backgroundHex and paletteId to be optional as well?

Divs-B: Also, do we want to add them in CardRootMetaData as well? 
https://github.com/guardian/facia-tool/blob/5cd8690d07d8c61b0985a68ac0de8a6e433dfc56/fronts-client/src/types/Collection.ts#L44

fronts-client/src/components/form/ChefMetaForm.tsx
```diff
@@ -84,12 +142,46 @@ const Form = ({
...
+    bio,
+    ...valuesToCopy
+  } = meta;
+
+  return {
+    ...valuesToCopy,
+    bio: bio ?? '',
+    image: {
+      src: imageSrc,
+      height: strToInt(imageSrcHeight),
```
fredex42, line 161: I believe that if the `imageSrcHeight` string is not a valid base-10 number then `parseInt()` (called by `strToInt`) will return `NaN` - is this case handled OK when the ChefCardFormData is serialized?

fronts-client/src/util/form.ts
```diff
@@ -63,8 +67,10 @@ export interface CapiFields {
   thumbnail?: string | void;
 }
 
-const strToInt = (str: string | void) => (str ? parseInt(str, 10) : undefined);
-const intToStr = (int: number | void) => (int ? int.toString() : undefined);
+export const strToInt = (str: string | void) =>
```
fredex42, line 53: No problems with this, but see my earlier comment that it may return `NaN` depending on the contents of the string and you might want to make sure that this is handled properly

fronts-client/src/actions/OptionsModal.ts
```diff
@@ -1,12 +1,13 @@
 import type { StartOptionsModal } from 'types/Action';
 import type { OptionsModalChoices } from 'types/Modals';
 import { ReactNode } from 'react';
+import { noop } from 'lodash';
```
fredex42, line 4: Do we need to bring in `'lodash'` just for this?

fredex42 (commented):
Like it! I made some comments which I hope are helpful, though feel free to disregard as I don't know the frontend code on this project well

fronts-client/src/actions/OptionsModal.ts
jonathonherbert: We use lodash in many places, but I can see elsewhere in the project we (mostly) tend to use `lodash/${util}`, which probably does avoid some bundling weight – so I've replaced the few references I can find to destructuring imports in f093e3f767f0c413419068b94363f438ae7b84b1.

fronts-client/src/types/Collection.ts
jonathonherbert: Agreed re: palette type – 719c9ded2b9c2548620d290e70fc64af5a5b7413.

Thanks for spotting that @Divs-B – I've adjusted the types slightly and made sure the relevant fields are optional. I've also used a nested image type to avoid the transformation between `image: { src: string }` and `imageSrc: string ...` in the form, which has simplified things a bit.

The `chefImageOverride` property will be able to have a more sensible name once we move to a sum type (as mentioned in the PR description.)

jonathonherbert: I've followed up by nesting the palette field in [53b15b5](https://github.com/guardian/facia-tool/pull/1589/commits/53b15b53bb8d5445706c9a8b35924b9fa1208c21). I think generally I've been operating with the mindset that we need to maintain as close a shape as possible to the original tooling for compatibility reasons, but your comments have made me realise that this is not the case, and ultimately the original model is problematic in part for the reasons you mention above – so thanks for the feedback!

emdash-ie (approved):
I haven’t yet run this locally, but it looks good to me! I have some questions below, but I’m happy to merge as-is as well :)

fronts-client/src/components/form/ChefMetaForm.tsx
```diff
@@ -100,23 +149,43 @@ interface ChefMetaFormProps {
...
+  const valueSelector = formValueSelector(form);
   const card = useSelector((state: State) => selectCard(state, cardId));
   const chefWithoutOverrides = useSelector((state: State) =>
     selectors.selectById(state, card.id)
   );
   const chef = useSelector((state: State) =>
     selectors.selectChefFromCard(state, cardId)
   );
+  const palette = useSelector(
+    (state: State) => valueSelector(state, 'palette') as Palette
```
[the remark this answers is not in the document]
jonathonherbert: > IIRC this is us telling the typescript compiler “trust me, this is the type of this value”, which makes me a little uneasy

Correct, and agreed! Sadly, the state of typings in `redux-form` is ... not very good. Here's `formValueSelector`:

```ts
export function formValueSelector<State = {}>(
    form: string,
    getFormState?: (state: State) => FormStateMap,
): (state: State, ...field: string[]) => any;
```

So we're stuck with an `any` type unless we cast. The compiler permits this (the tsconfig in this project is not strict, which is why I think this is permitted elsewhere), so a developer assertion at least allows us to reason about the type of this value as it enters the form. But I'm not sure we can do much more here without writing our own types for `redux-form` 😢

fronts-client/src/components/form/ChefMetaForm.tsx
```diff
@@ -100,23 +149,43 @@ interface ChefMetaFormProps {
...
+    []
+  );
 
   return (
     <ConnectedChefForm
       chef={chef}
       chefWithoutOverrides={chefWithoutOverrides}
       card={card}
-      initialValues={initialValues}
+      initialValues={card.meta as ChefCardMeta}
```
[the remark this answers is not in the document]
jonathonherbert: To be fair, this should go away with a larger refactor: comment added in 8cb223e77a51f05f2e50ac747c68f58cca50b6c4.

fronts-client/src/components/form/PaletteForm.tsx
```diff
@@ -0,0 +1,208 @@
...
+    const dispatch = useDispatch();
+    const formPalette = useSelector((state: State) => {
+      const formValues = getFormValues(formName)(state) as {
+        [T: string]: Palette;
+      };
+      if (!formValues) {
+        return undefined;
+      }
+
+      return formValues[fieldName] as Palette;
```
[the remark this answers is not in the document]
jonathonherbert: Frustratingly, I think `Object.entries` will not give a union type for the object keys, even if they're known type – we just get `string`, requiring a cast further downstream.

In reminding myself why Typescript behaves this way, I stumbled across the first answer in [this post on stackoverflow,](https://stackoverflow.com/questions/60141960/typescript-key-value-relation-preserving-object-entries-type) which was very informative, and gave a few examples of a better-typed implementation! I've added that in 10213da5b1476cf0ff13474472ba9947c3891895. Thanks for the prompt, this is nicer.

fronts-client/src/constants/feastPalettes.ts
```diff
@@ -0,0 +1,277 @@
...
+    imageURL: 'https://uploads.guim.co.uk/2024/03/21/Cakes.png',
+  },
+  Celebration: {
+    lightForegroundHex: '#801919',
+    lightBackgroundHex: '#F9F9F5',
+    darkForegroundHex: '#FCF1F0',
+    darkBackgroundHex: '#363632',
+    imageURL: 'https://uploads.guim.co.uk/2024/03/21/Celebration.png',
+  },
+  'Celebration\nCakes': {
```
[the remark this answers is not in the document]
jonathonherbert: We're stuck for now, but I will point this out to Feast. It's weird.

fronts-client/src/components/form/ChefMetaForm.tsx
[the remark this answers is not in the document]
emdash-ie: Ah yes, in that case a cast is better than nothing! Thanks for the thorough explanation :)

fronts-client/src/components/form/PaletteForm.tsx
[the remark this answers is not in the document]
jonathonherbert: I think I read your comment as relating to `as PaletteId`! The improvements above are fine, but spot on, this is not needed. Possibly artefact from previous model where keys were a union. 564fab795c3a3e111556769e79c8f5b159ec6358

PR #1591, opened 2024-06-10
set criteria by collection type, style portrait thumbnails, validate on drop

dblatcher:
## What's changed?
Adds features (behind a feature switch) to the fronts tool to support some collection types using Portrait Crops for replacement trail images and/or slideshows

Utility code etc:
 - adds new feature switch "support-portrait-crops", hides it from the features menu of PROD
 - defines an empty string array `COLLECTIONS_USING_PORTRAIT_TRAILS` -  in the "constants" folder. This can be populated with the `collectionType` of the new collection when implemented.
 - refactors image validation so the tests for whether a set of dimensions(as opposed to a whole media object) matches image `Criteria` can be exported separately (as below, this is used to validate if a crop dragged from one card match the criteria of the card it was dropped on
 - Criteria validation error strings (shown to the user in alerts) specific which type of crop is missing (eg "does not have a valid 5:4 crop" rather than "does not have any valid crops")

Component behaviour:
`Card` 
 - adds `collectionType` prop (mapped from app State) and a `determineCardCriteria` using that prop 
 - when a crop is dragged onto a Card from another Card, if will validate that other Card's Crop matches its criteria

`ArticleBody` & `ThumbnailSmall`
 - If there is a replacement image defined and the dimensions are portrait, `ArticleBody`  passes a prop to `ThumbnailSmall` to vary the styling to suit a 4:5 image
 
`ArticleFormMeta` & `InputImage` 
  - adds `collectionType` prop (mapped from app State) and a `determineCardCriteria` using that prop  (like `Card` does) and proves the criteria to `InputImage`
  - refactors the styling code for the ImageContainer to allow for four variants (based on the values of the 'portrait' and 'small' boolean props.

`CollectionDisplay`
 - adds a 'crop' icon next to the title if the collection uses portrait crop (**we should get design input before enabling the feature in earnest- not sure of the best way to communicate the difference in crop type**)

## testing locally

- add a collection type (eg 'fixed/small/slow-IV') to `COLLECTIONS_USING_PORTRAIT_TRAILS` 
- open https://fronts.local.dev-gutools.co.uk/v2/features and set "Use portrait crops for the experimental big card containers" to "on"
- open a front with a collection of of the type set to use portrait trail - you should see the "Crop" icon to the left of the title
- Setting the feature switch to 'off' and refreshing the front page should remove the icon and stop you selecting portrait crops - however, if you have saved portrait crops on any of the cards, they will still be in place - if you try to replace them, the choice should be restricted to landscape crops.

## Implementation notes
As long as `COLLECTIONS_USING_PORTRAIT_TRAILS` is empty, there should be no noticeable change in behaviour, other than the changes in the error alert including the crop dimensions of the criteria.

Using a hard coded string array on the frontend isn't the ideal way for configuring which collections use which crop type. It should be okay whilst this feature is experimental/in development.

The extra validation on drag and drop will be happening, but should:
 - never fail as all cards can only have 5:3 crops as replacement images
 - not impact performance. the image dimensions are looked up in the global state, not by querying the grid

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

Collection ("Reader's Tips") with Portrait crops:
Screenshot 2024-06-12 at 10 51 27

Cards with portrait thumbnails:
Screenshot 2024-06-12 at 10 58 01

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -13,7 +13,7 @@ jobs:
     uses: guardian/.github/.github/workflows/sbt-node-snyk.yml@main
     with:
       ORG: guardian
-      JAVA_VERSION: 8
+      JAVA_VERSION: 11
       EXCLUDE: package-lock.json # exclude V1, since it has its own special job below (because of JSPM)
     secrets:
       SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
@@ -22,7 +22,6 @@ jobs:
     uses: guardian/.github/.github/workflows/sbt-node-snyk.yml@main
     with:
       ORG: guardian
-      JAVA_VERSION: 8
       EXCLUDE: fronts-client # exclude V2, since it's captured by the main job above
       SKIP_SBT: true # exclude scala, since it's captured by the main job above
       NODE_PACKAGE_JSON_FILES_MISSING_LOCK: v1_jspm_snyk_workaround/package.json v1_jspm_snyk_workaround/result/package.json
--- a/app/controllers/UserDataController.scala
+++ b/app/controllers/UserDataController.scala
@@ -1,20 +1,17 @@
 package controllers
 
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
-import com.gu.facia.client.models.{Trail, TrailMetaData}
+import com.gu.facia.client.models.Trail
+import model.editions.{CardType, EditionsClientCard}
 import org.scanamo._
 import org.scanamo.syntax._
 import model.{FeatureSwitch, FeatureSwitches, UserData}
 import org.scanamo.generic.auto.genericDerivedFormat
 import org.scanamo.query.UniqueKey
-import play.api.libs.json.{JsArray, JsValue, Json}
-import model.{UserData}
-
-import play.api.libs.json.JsValue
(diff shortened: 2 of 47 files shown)


rebecca-thompson (approved):
Have played with this locally and in CODE and it all works as expected. A couple of visual spacing issues, maybe?

Screenshot 2024-06-17 at 10 26 32

Can we adjust the spacing, so the thumbnails are not overlapping?

Screenshot 2024-06-17 at 10 31 59

It looks like this icon could use some right hand spacing. Also, just a question... is this the final icon or a placeholder? Seems an inconsistent choice given the crop icon generally represents re-cropping an image in our tools.

dblatcher:
> Have played with this locally and in CODE and it all works as expected. A couple of visual spacing issues, maybe?
> 
> Can we adjust the spacing, so the thumbnails are not overlapping?

Good point, I must have missed that. I'll take a look.

> It looks like this icon could use some right hand spacing. Also, just a question... is this the final icon or a placeholder? Seems an inconsistent choice given the crop icon generally represents re-cropping an image in our tools.

It was intended as a placeholder,  but you're right that its not ideal. I was thinking of it as a symbol for 'crops' in general, but that's not how it's used elsewhere. I think I'll swap the icon for a small badge component of some kind  that can include the "5:4" text.

PR #1593, opened 2024-06-11
don't style the ImageContainer as portrait on editions

dblatcher:
## What's changed?
Fixes an unintended change from https://github.com/guardian/facia-tool/pull/1590/

When selected Covers images in the editions tool, any shape of crop can be used - including "portrait" crops . If a crop was selected on editions that is taller than it is wide, this caused the `portrait` prop to be passed the the `ImageContainer` component (display wrapper that controls the size and shape of the ImageInput UI).  This prop was only intended to be used for non-editions cards for which a "Portrait" (IE 5:4) crop was selected (as per https://github.com/guardian/facia-tool/pull/1590/ this option is not yet supported, but is planned for a new collection type)

This PR adds an extra conditional to avoid using the portrait styling in Editions Cards with replacement images.

## Implementation notes

Believe the behaviour prior  https://github.com/guardian/facia-tool/pull/1590/ would be for thumbnails of  non-landscape crops in the `InputImage` to bet 'cut off' as they are in the "after" (IE after this change to put back how it was)screenshot. 

This doesn't seem  ideal - we could do another change to make these thumbnails more dynamic.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### Screenshots

before (mobile cover on left  is portrait, tablet on right is landscape):
Screenshot 2024-06-11 at 12 56 37

after: before (mobile cover on left  is portrait, tablet on right is landscape):
Screenshot 2024-06-11 at 12 56 07

### DIFF
--- a/fronts-client/src/components/inputs/InputImage.tsx
+++ b/fronts-client/src/components/inputs/InputImage.tsx
@@ -272,8 +272,6 @@ class InputImage extends React.Component<ComponentProps, ComponentState> {
 
     const { src } = valueRecord ?? {};
     const imageSrc = typeof src === 'string' ? src : '';
-
-
     this.state = {
       isDragging: false,
       modalOpen: false,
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
Tested locally, LGTM 👍

jonathonherbert:
@dblatcher sounds like Katy might not mind seeing more of the image, so perhaps this is not needed.

PR #1594, opened 2024-06-12
Add search chef

Divs-B:
## What's changed?

Add search for chef. 
For that we first need chef data rendered into the left container that is search, 
which we are trying to get based on tags that are only contributors at the moment.
Later we will change this to fetch tags that are contributor of type chefs only. 

ref: https://trello.com/c/YsDDwuQr/2761-enable-search-for-chefs

Co-authored: @jonathonherbert (Thanks for sorting issues out and to make PR work altogether 💫).

https://github.com/guardian/facia-tool/assets/17780995/bad6d226-1613-495d-b549-abc2108a1340

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/bundles/chefsBundle.ts
+++ b/fronts-client/src/bundles/chefsBundle.ts
@@ -96,7 +96,7 @@ export const createFetch =
 
              */
         dispatch(
-          actions.fetchSuccess(resultData, {
+          actions.fetchSuccess(resultData.results, {
             pagination: resultData.pagination || undefined,
             order: resultData.results.map((_) => _.id),
           })
--- a/fronts-client/src/bundles/chefsBundle.ts
+++ b/fronts-client/src/bundles/chefsBundle.ts
@@ -1,16 +1,11 @@
 import createAsyncResourceBundle from '../lib/createAsyncResourceBundle';
 import { Chef } from '../types/Chef';
-import chefOttolenghi from './fixtures/chef-ottolenghi.json';
-import chefStein from './fixtures/chef-stein.json';
-import chefCloake from './fixtures/chef-cloake.json';
 import { selectCard } from 'selectors/shared';
 import { State } from 'types/State';
 import { createSelector } from 'reselect';
 import { stripHtml } from 'util/sanitizeHTML';
-// import { createSelectIsArticleStale } from 'util/externalArticle';
 import { ThunkResult } from 'types/Store';
 import { previewCapi, liveCapi } from 'services/capiQuery';
-import { createSelectIsArticleStale } from '../util/externalArticle';
 import { Tag } from '../types/Capi';
 
 const sanitizeTag = (tag: Tag) => ({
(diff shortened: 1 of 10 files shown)


fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -31,24 +32,56 @@ const FixedContentContainer = styled.div`
...
+
+  const getParams = (query: string) => ({
+    'web-title': query,
+    'page-size': '20',
+    'show-elements': 'image',
+    'show-fields': 'all',
+  });
+
+  const runSearch = useCallback(
+    (page = 1) => {
```
emdash-ie, line 73: Am I right to think this default value for page is always used? Presumably this means you can only see the first page of results for now.

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -67,21 +100,28 @@ const FeastSearchContainerComponent = ({
     <React.Fragment>
       <InputContainer>
         <TextInputContainer>
-          <TextInput placeholder="Search recipes" displaySearchIcon />
+          <TextInput
+            placeholder="Search recipes"
```
emdash-ie, line 104: It might be good for this placeholder to vary according to whether the chefs or recipes toggle is selected. Not sure whether that should be part of this PR or could come later, though!

emdash-ie (approved):
Looks good, merge away :) I have a couple of questions below, but there’s no need to address them before merging.

Another thing I wondered about when using the chef search: would it be worthwhile to include the chef’s bio in the feeditems? Searching for `ottol` gave me two results, `ottolenghi` and `Yotam Ottolenghi` and it wasn’t clear to me that the former was a separate person until I dragged into place on the right and saw the description of the contributor. (If this is something we’ll get to later, then ignore me!)

fronts-client/src/bundles/chefsBundle.ts
```diff
@@ -1,28 +1,71 @@
...
+    results: response.results,
+    pagination: {
+      totalPages: response.pages,
+      currentPage: response.currentPage,
+      pageSize: response.pageSize,
+    },
+  };
+};
+
+export const createFetch =
```
jonathonherbert, line 40: I'd recommend refactoring this to remove the additional wrapping function – as I think we'll always want tags from live CAPI, and we'll always want to use bundle.actions.

So this function might be renamed `fetchChefs`, and `fetchLive` below would no longer be needed.

jonathonherbert (commented):
One non-blocking comment to follow up @emdash-ie's review – this looks great! Agreed re: the placeholder changing, I didn't spot that when we were pairing yesterday.

I've created a card for pagination in a follow-up PR 📄 📄📄

Divs-B:
Thanks @emdash-ie and @jonathonherbert for review, 
As discussed we will merge this PR and will address the changes in the next PR with @jonathonherbert

PR #1595, opened 2024-06-13
Fetch chefs as collections are loaded

jonathonherbert:
_co-authored-by: @Divs-B @JustinPinner_ 

## What's changed?

Fetch chefs as collections are loaded.

It should be easy to extend the logic to recipes once we've a recipes search.

|Before|After|
|-|-|
|fronts-chefs-fetch-bad|fronts-chefs-fetch-good|

## Implementation notes

We've slimmed down the fixtures with a utility assertion function. It's still a bit verbose, but hopefully a bit less repetitious than before.

The size of the PR is increased by a refactor to ensure that selectors that referenced 'articles' are now referencing 'cards', which is more correct. Inspecting individual commits should give a clearer picture of the relevant changes.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/actions/Collections.ts
+++ b/fronts-client/src/actions/Collections.ts
@@ -50,7 +50,7 @@ import type { State } from 'types/State';
 import { cardSets, noOfOpenCollectionsOnFirstLoad } from 'constants/fronts';
 import { Stages, Collection, CardSets } from 'types/Collection';
 import difference from 'lodash/difference';
-import { selectArticlesInCollections } from 'selectors/collection';
+import { selectCardsInCollections } from 'selectors/collection';
 import {
   editorOpenCollections,
   editorCloseCollections,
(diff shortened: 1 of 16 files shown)


fronts-client/src/actions/Collections.ts
```diff
@@ -313,15 +314,16 @@ function updateCollection(
 const fetchArticles =
   (articleIds: string[]): ThunkResult<Promise<void>> =>
   async (dispatch, getState) => {
-    const articleIdsWithoutSnaps = uniq(
-      articleIds.filter((id) => !id.match(/^snap/))
+    const uniqueArticleIdsWithoutSnaps = articleIds.filter(
```
jonathonherbert, line 317: Made sense to change this for readability as we went through the code.

fronts-client/src/actions/Collections.ts
```diff
@@ -68,8 +67,10 @@ import { fetchCollectionsStrategy } from 'strategies/fetch-collection';
 import { updateCollectionStrategy } from 'strategies/update-collection';
 import { getPageViewDataForCollection } from 'actions/PageViewData';
 import { isMode } from 'selectors/pathSelectors';
+import { groupBy, uniqBy } from 'lodash';
+import { fetchChefsById } from 'bundles/chefsBundle';
 
-const articlesInCollection = createSelectAllArticlesInCollection();
+const articlesInCollection = createSelectAllCardsInCollection();
```
JustinPinner, line 73: Presumably `articlesInCollection` could also be renamed `cardsInCollection` (but not essential, obvs)

JustinPinner (approved):
Happy to approve this (my "co-authoring" input was more at the observational level anyway) 👍 Please seek other opinions too though

fronts-client/src/actions/Collections.ts
jonathonherbert: Yep, missed this, thanks for spotting! e91b656f981d50d9730aaf169ba88eec5314769e

PR #1597, opened 2024-06-17
Add pagination on chef search 

Divs-B:
## What's changed?
This adds pagination on chef search, a follow up PR on #1594 that missed pagination [(as per review too)](https://github.com/guardian/facia-tool/pull/1594#discussion_r1638395015) ,
This will use `IPagination` to make it easy to maintain paging (similar to CapiSearchContainer).

This will be draft PR and will go under changes once @jonathonherbert PR #1595 (which is currently going parallely) gets deployed because this PR also contains refactored code [as per review on PR 1594](https://github.com/guardian/facia-tool/pull/1594#discussion_r1638514036) around fetch chef.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -56,11 +67,21 @@ export const RecipeSearchContainer = ({ rightHandContainer }: Props) => {
     chefSelectors.selectLastFetchOrder(state)
   );
 
+  const [page, setPage] = useState(1);
+
   const debouncedRunSearch = debounce(() => runSearch(), 750);
```
jonathonherbert, line 72: Sorry, I missed this on the initial review of #1594 – I think this will need to be wrapped in `useCallback` to preserve the identity of the `debouncedRunSearch` function – at the moment, I think we will be getting a new function every render, which will make the debouncing behaviour a bit unpredictable. We could add this in another PR (which I'm happy to write!)

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -109,22 +134,33 @@ export const RecipeSearchContainer = ({ rightHandContainer }: Props) => {
...
+            onChange={() => setSelectedOption(FeedType.chefs)}
+            label="Chefs"
+            inline
+            name="chefFeed"
+          />
+        </RadioGroup>
+        {selectedOption === FeedType.chefs && chefsPagination && hasPages && (
+          <PaginationContainer>
+            <Pagination
+              pageChange={(page) => setPage(page)}
```
jonathonherbert, line 157: At the moment, setPage triggers the effect that runs our search – but the result of that is that selecting pages feels quite slow, as we have to wait 750ms for the fetch to begin! Perhaps we could trigger a fetch immediately here, after updating the page number in the state?

jonathonherbert (commented):
This is looking very promising! A few comments on behaviour inline.

A few things I've noticed:
- The results do not seem to change if I have a search term in the box and then change pages:
tag-search-page
I am clicking the page buttons after searching for 'Alex', but nothing is happening in the UI – although new searches are being triggering in the network tab.
- we can't scroll through results that are longer than the page. (I didn't notice this in the previous PR, my fault!) Fixing this could be another PR.

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -56,11 +67,21 @@ export const RecipeSearchContainer = ({ rightHandContainer }: Props) => {
...
   const debouncedRunSearch = debounce(() => runSearch(), 750);
 
   useEffect(() => {
-    debouncedRunSearch();
-  }, [selectedOption, searchText]);
+    if (page > 1 && searchText.length === 0) runSearch(page);
+    else debouncedRunSearch();
+  }, [selectedOption, searchText, page]);
+
+  //const chefsPagination = IPagination | null;
```
emdash-ie, line 79: Could we remove this comment? I don’t think it adds anything, unless I misunderstand

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -56,11 +67,21 @@ export const RecipeSearchContainer = ({ rightHandContainer }: Props) => {
...
+  const chefsPagination: IPagination | null = useSelector((state: State) =>
+    chefSelectors.selectPagination(state)
+  );
+
+  const hasPages = !!(chefsPagination && chefsPagination.totalPages > 1);
```
emdash-ie, line 84: I find this condition a bit hard to understand (possibly because I’m less used to javascript/typescript conventions!). Would the following be an equivalent condition, and if so could we use that instead?

```suggestion
  const hasPages = (chefsPagination != null) && (chefsPagination.totalPages > 1);
```

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -109,22 +134,33 @@ export const RecipeSearchContainer = ({ rightHandContainer }: Props) => {
...
+          />
+          <RadioButton
+            checked={selectedOption === FeedType.chefs}
+            onChange={() => setSelectedOption(FeedType.chefs)}
+            label="Chefs"
+            inline
+            name="chefFeed"
+          />
+        </RadioGroup>
+        {selectedOption === FeedType.chefs && chefsPagination && hasPages && (
```
emdash-ie, line 154: Do we not also need pagination for recipes? (Is there a search for recipes yet? I can’t remember)

emdash-ie (commented):
I have a couple of minor questions below, but nothing that would block a merge. I’ll hold off on approving though, as I think @jonathonherbert’s point about the delay selecting pages is worth addressing pre-merge

fronts-client/src/components/feed/RecipeSearchContainer.tsx
emdash-ie: Sorry, meant to check this before I finished the review! I see that we don’t have recipe search yet, so no pagination is needed: ignore me.

fronts-client/src/components/feed/RecipeSearchContainer.tsx
Divs-B: Thanks @emdash-ie for suggesting another way of writing above conditional code, I think I might leave it as-is to maintain consistency of using `!!` negation that are there on other places in code too. Also it return strict boolean which is common way of doing in JS it seems . 
I would like to have view from @jonathonherbert to check on this.

Divs-B:
@jonathonherbert , it seems changes have fixed - 'moving to different pages when chefs are listed on the basis of search text'. Also scrolling is added too. thanks for spotting 👍

Divs-B:
I can see a one issue (different to this ) on searching chefs -> Any text in search box is fine and giving results except text "Alex" that you have used in example 😄 .  i think data is duplicated on one of contributor i think "Alex Clapham"  where code is throwing an error. I am looking into this and will report back.

fronts-client/src/components/feed/RecipeSearchContainer.tsx
emdash-ie: Sure, happy to stick with the approach taken elsewhere! I think the condition I’ve proposed would also strictly return a boolean as well though, since each of the comparisons should return a boolean, feeding into the `&&` to return a boolean. (JavaScript can be odd about types, but I don’t think it’ll turn a boolean into something else here, and I don’t think the `> 1` or `!= null` comparisons can return a non-Boolean value.)

Divs-B: O yes absolutely, your code will be returning boolean too (thats why I mentioned first that thanks for suggesting another way of conditional code), i think my main emphasising was on "common" way of using `!!` in JS for returning boolean which might ask to keep the condition as-is. 
But will check with @jonathonherbert too.

jonathonherbert (commented):
Tested locally, both comments addressed! Happy to pair on the outstanding question re: debouncing.

Divs-B:
> I can see a one issue (different to this ) on searching chefs .....

So, when we search "Alex" in Chef search and click on page 2,
We get error:
**The two children with the same key, `profile/alex-clapham`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior**


image

________

We have two conflicting keys in response:


image

_________

I think if I remember it correctly we talked about de-duplicating the data in collections so dont know why still this issue is there.

My question is - to fix this, do we need to correct in Source level so that we should not be having this duplicacy means to correct why two profile exists at the first place?
Or, we should try accessing `r2ContributorId` as key which seems unique in this case : `r2ContributorId: "81919"` and `r2ContributorId: "81920"`?

@jonathonherbert

Divs-B:
While waiting for discussion on the issue of duplicate keys, a change is pushed in that fixes it but need to verify as aforementioned.

fronts-client/src/components/feed/RecipeSearchContainer.tsx
emdash-ie: Yes, that seems a bit clearer to me than the double negation.

Divs-B: Thanks @emdash-ie for revising, i will commit above suggestion then 👍  Thanks @jonathonherbert

Divs-B:
Thanks @jonathonherbert for the discussion on the issue for duplicate contributor key, 
We agreed on fixing this by constructing unique id rather then removing any duplicate because we can have contributors with same profile id, the work will be in next PR. 
For now, this PR is ready to go.

jonathonherbert (approved):

davidfurey:
We have manually deleted 81919 from the database. Hopefully that solves the issue for you. It will still be hanging around in CAPI until the next reindex.

Divs-B:
> We have manually deleted 81919 from the database. Hopefully that solves the issue for you. It will still be hanging around in CAPI until the next reindex.

That's really great! Thanks for helping in clearing out redundant profile from TagMngr.

PR #1600, opened 2024-06-24
Fix duplicate chef ids exception on search

Divs-B:
## What's changed?

This is follow up PR to fix an issue, a scenario where duplicate chef id's present in the chef-feeds when doing search.
Fronts throws exception that key should be unique to render on the feed.
Details: https://github.com/guardian/facia-tool/pull/1597#issuecomment-2180419689

We are using `lastFetchOrder` to get search list page by page, which are collection of `id`'s  but prone to duplicates.
This can be fix by making `lastFetchOrder` to be unique for chefs ,means combining `id` and unique id's that are available as `r2ContributorId` under contributor data structure.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/bundles/chefsBundle.ts
+++ b/fronts-client/src/bundles/chefsBundle.ts
@@ -50,7 +50,9 @@ export const fetchChefs =
         dispatch(
           actions.fetchSuccess(resultData.results.map(sanitizeTag), {
             pagination: resultData.pagination || undefined,
-            order: resultData.results.map((_) => _.id),
+            order: resultData.results.map(
+              (_) => _.id + '#' + _.r2ContributorId //to fix duplicate chef id where Fronts throws exception (example "profile/alex-clapham" was twice in the data), hence suffix with unique r2ContributorId.
+            ),
           })
         );
       } else {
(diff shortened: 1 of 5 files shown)


jonathonherbert (commented):
This looks like it will work well for this use case! Well done for finding a novel solution.

However, I'd recommend not storing non-id information in our order table, as every action that writes this data must know the format, and every consumer of the order information must know the format as well!

(We think we know that there is only one action/consumer at the moment, but this may change in the future.)

Can I recommend writing a selector to derive this information when it's needed, instead? It would

- select the ordered ids
- map across them to return full cards

We would then have access to the full card within our search UI, and constructing a unique composite key with the id and the r2 id would be simple.

Very happy to pair!

Divs-B:
> This looks like it will work well for this use case! Well done for finding a novel solution.
> 
> However, I'd recommend not storing non-id information in our order table, as every action that writes this data must know the format, and every consumer of the order information must know the format as well!
........

Yes @jonathonherbert absolutely! creating a selector is utmost appropriate choice, thanks for confiorming.
I had that in mind too so lets discuss in pairing i think that will fill where I am was missing to get the `selector` done. 👍

Divs-B:
@jonathonherbert and I tried investigating more on this and we agreed to check this at TagManager's end where data is originated for contributors so that duplicate id's should not exist for same contributor. Posted query to CoPro team.

PR #1602, opened 2024-07-01
Create feast clipboard

Divs-B:
## What's changed?

We observed that if a user adds feast based (recipes/chefs) cards to the clipboard while using the Feast Edition and tries to switch to other Editions, those feast clipboard cards appears in other editions too. 
Having a separate feast-clipboard will fix this.

Guided and Co-authored by @jonathonherbert

## Implementation notes
reference: PR #840
A separate clipboard fields for feast cards is created in the database, to get used under Feast Edition only.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fronts-client/src/actions/Collections.ts
```diff
@@ -368,7 +369,7 @@ const fetchCardReferencedEntities =
     const itemSets = Array.isArray(itemSetCandidate)
       ? itemSetCandidate
       : [itemSetCandidate];
-
+    //1.
```
jonathonherbert, line 372:
```suggestion
```

fronts-client/src/lib/createAsyncResourceBundle/index.ts
```diff
@@ -246,7 +246,7 @@ function createAsyncResourceBundle<Resource>(
   const selectLastFetchOrder = (state: RootState): string[] =>
     selectLocalState(state).lastFetchOrder || defaultArray;
 
-  const selectAll = (state: RootState) => selectLocalState(state).data;
+  const selectAll = (state: RootState) => selectLocalState(state)?.data || {};
```
jonathonherbert, line 249: Is this change still necessary? I can't remember whether we still need it, or whether it's an artefact of the journey we took to get here 😀

app/controllers/V2App.scala
```diff
@@ -47,14 +51,17 @@ class V2App(isDev: Boolean, val acl: Acl, dynamoClient: DynamoDbClient, val deps
...
-    val clipboardArticles = if (editingEdition)
-      maybeUserData.map(_.editionsClipboardArticles.getOrElse(List()).map(ClipboardCard.apply))
-    else
+    val clipboardCards = if (editingEdition) {
+      if(isFeast)
+        maybeUserData.map(_.feastEditionsClipboardCards.getOrElse(List()).map(ClipboardCard.apply))
+      else
+        maybeUserData.map(_.editionsClipboardArticles.getOrElse(List()).map(ClipboardCard.apply))
+    } else
       maybeUserData.map(_.clipboardArticles.getOrElse(List()).map(ClipboardCard.apply))
```
jonathonherbert, line 60: Non-blocking – when there's some logic spread over a few variables, one pattern is to use a tuple and pattern-match on the possibilities. This can look a bit more readable:

```suggestion
    val clipboardCards = (editingEdition, isFeast) match {
      case (true, true) =>  maybeUserData.map(_.feastEditionsClipboardCards.getOrElse(List()).map(ClipboardCard.apply))
      case (true, false) => maybeUserData.map(_.editionsClipboardArticles.getOrElse(List()).map(ClipboardCard.apply))
      case _ => maybeUserData.map(_.clipboardArticles.getOrElse(List()).map(ClipboardCard.apply))
    }
```

Anyhow, take a look and see what you think. (I've tested this locally, but do go ahead and make sure it is working yourself before merging!)

jonathonherbert (approved):
Tested in CODE and this looks great. A few non-blocking comments/suggestions. Great work!

PR #1605, opened 2024-07-17
Sealed trait for Editions Card

jonathonherbert:
_co-authored-by: @Divs-B_

## What's changed?

Makes the Editions Card a sealed trait, which will improve our modelling as we introduce different metadata for different card types, and make it easier to guard against incorrect behaviour – for example, articles in Feast issues, and recipes in Editions issues.

This is a no-op – nothing in the app behaviour should change. We have good tests as guardrails here, but a end-to-end check of Editions preview and publication 

## Implementation notes

There are two important changes:
- a sealed trait for the `EditionsCard` type, which is now inherited by Article, Recipe, Chef and FeastCollection cards. ([ddc7dec](https://github.com/guardian/facia-tool/pull/1605/commits/ddc7decdf95e932bf88d875624ebf7337af47234), [e7b2851](https://github.com/guardian/facia-tool/pull/1605/commits/e7b28519463505ffb750a99b56f22d071baa9bbf))
- moving the Editions App- and Feast-specific modelling and logic for publication into their respective `PublicationTarget` implementations, so `FeastPublicationTarget` does not depend on a `PublishedIssue` (which contains Editions-specific logic.) [f043cd4](https://github.com/guardian/facia-tool/pull/1605/commits/f043cd41d9d51ceaa777fea621474743a114d5f2)

As an incidental change, we've renamed `EditionsBucket` to `EditionsAppPublicationTarget`, to better reflect its relationship with the Editions App product and the trait it implements. ([6a3c606](https://github.com/guardian/facia-tool/pull/1605/commits/6a3c6060d4d79a1b5231555cd9aa8ed46919bef7))

We _haven't_ touched the `EditionsClientCard` model (which could also be a sealed trait.) The reasoning for this is that this model is simply there to adapt a few article-specific properties (`id` formatting,  `frontPublicationDate` -> `addedOn`), and replicating the hierarchy in another place felt like overkill. Sadly, this means we still keep the expansive `ClientCardMetadata` case class, which contains all possible meta properties. We could follow this up with an alternative approach if there was appetite.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### Additional checks
- [x] Preview editions behaves as expected
- [x] Proof editions behaves as expected
- [x] Proof feast behaves as expected

### DIFF
--- a/app/Components.scala
+++ b/app/Components.scala
@@ -21,7 +21,7 @@ import services._
 import services.editions.EditionsTemplating
 import services.editions.db.EditionsDB
 import services.editions.publishing.events.PublishEventsListener
-import services.editions.publishing.{EditionsBucket, FeastPublicationTarget, Publishing}
+import services.editions.publishing.{EditionsAppPublicationTarget, FeastPublicationTarget, Publishing}
 import slices.{Containers, FixedContainers}
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient
 import thumbnails.ContainerThumbnails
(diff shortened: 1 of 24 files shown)


fredex42 (commented):
All makes sense to me 😁 . It would be nice to overhaul the client-facing specific class as the modelling is still a bit cumbersome there but that can wait till we have something out and working initially

fredex42 (approved):
🆒

jonathonherbert:
Tested in preview over VPN, updates appearing as expected 👍

PR #1607, opened 2024-07-22
Add collection overrides

jonathonherbert:
## What's changed?

Adds a collection overrides form:
- overrides for feast collection title and theme
- show current palette selection in feast collection and chef cards

palettes

Most of the diff relates to changes to the palette fixtures, but nevertheless, this PR is quite substantial: the palette data and model needed a bit of an overhaul to let us model the palette data in a way which let us refer to it stably (by id and palette key), and I perhaps went a little overboard in producing a palette form that could be reused between chefs and collections 😅 

## Implementation notes

One shortcoming of the current approach to client side meta, despite the better modelling introduced in #1605, is that metadata properties that are specific to cards, e.g. `feastCollectionTheme`, `chefTheme`, must have discrete names, rather than a single `theme` property, in order to work with the autogenerated JSON serialisers. We could introduce a sealed trait of `Theme`, which would require us to add a discriminator field to the client model, or introduce custom serde code for that field. Let me know what you think.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


app/controllers/EditionsController.scala
```diff
@@ -24,6 +24,7 @@ import util.{SearchResponseUtil, UserUtil}
 import scala.jdk.CollectionConverters._
 import scala.concurrent.ExecutionContext
 import scala.util.Try
+import model.editions.client.EditionsFrontendCollectionWrapper
```
jonathonherbert, line 27: Necessary as code in `editions/client` is now in the `editions.client` package, on the premise that the package structure should reflect the file structure.

app/model/editions/EditionsCard.scala
```diff
@@ -28,13 +28,24 @@ object CoverCardImages {
   implicit val format: OFormat[CoverCardImages] = Json.format[CoverCardImages]
 }
 
-
-case class Palette(paletteId: String, foregroundHex: String, backgroundHex: String)
+case class Palette(foregroundHex: String, backgroundHex: String)
```
jonathonherbert, line 31: ID is now a property of the meta theme, not the palette, because a theme can contain multiple palettes.

fronts-client/src/types/Collection.ts
```diff
@@ -42,53 +41,54 @@ type NestedCard = NestedCardRootFields & {
...
-    width?: string;
-    height?: string;
-    origin?: string;
-  }>;
-  overrideArticleMainMedia?: boolean;
-  coverCardImageReplace?: boolean;
-  coverCardMobileImage?: ImageData;
-  coverCardTabletImage?: ImageData;
-}
+type CardRootMeta = ChefCardMeta &
```
jonathonherbert, line 44: Per #1605, the ambition is for the Card model on the client to be a union of different card times, but because this touches a great many places where cards are used, I've opted to avoid this for now to keep the scope of this PR reasonable.

fronts-client/src/components/form/PaletteForm.tsx
fredex42, line 1: Big readability improvements 😁

fredex42 (approved):
This looks really great to me, demo is impressive and while there is a lot of layout the end result is much more readable and the object modelling is better. 💯

PR #1609, opened 2024-07-29
Store recipes nested in feast collections, and ensure other card types are not permitted within them

jonathonherbert:
## What's changed?

Stores recipes nested in feast collections, and ensures that other card types are not permitted within them.

Note the refresh, indicating that the data is stored correctly. Prior to this PR, refreshing the page would remove the nested items.

collections

## Implementation notes

Although it's no longer possible to put chefs and feast collections _within_ feast collections, the drop icons do still appear when the user hovers over them. I timeboxed fixing this but it turned out to be a bit more complicated that expected! Perhaps we can circle back if user feedback suggests it's a priority.

It wasn't far off the beaten track for this PR, so I've added some messaging changes, too – the drop placeholders say 'Place recipe here' (they used to say 'sublink'), and the sublink expander at the bottom of the Feast Collection card says 'recipes' rather than 'sublinks'.

## How to test

- Attempt to add various cards to a Feast Collection. Do recipes go in? Do all others stay out?
- Refresh the page. The recipes should remain in the collection.
- The behaviour of article drag and drop in usual fronts should not have changed – the 'sublinks' messaging should remain: 
Screenshot 2024-07-29 at 16 24 04

All tested locally.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/FrontsEdit/CollectionComponents/Sublinks.tsx
+++ b/fronts-client/src/components/FrontsEdit/CollectionComponents/Sublinks.tsx
@@ -71,7 +71,7 @@ class Sublinks extends React.Component<SublinkProps> {
               this.setState({ dragHoverActive: false });
             }}
             delay={100}
-            filterRegisterEvent={this.dragEventNotDenylisted}
+            filterRegisterEvent={e => !denyDragEvent()(e)}
             onIntentConfirm={() => {
               toggleShowArticleSublinks();
             }}
@@ -99,8 +99,6 @@ class Sublinks extends React.Component<SublinkProps> {
       </>
     );
   }
-
-  private dragEventNotDenylisted = (e: React.DragEvent) => !denyDragEvent()(e);
 }
 
 export default Sublinks;
--- a/fronts-client/src/components/FrontsEdit/CollectionComponents/Sublinks.tsx
+++ b/fronts-client/src/components/FrontsEdit/CollectionComponents/Sublinks.tsx
@@ -40,7 +40,7 @@ interface SublinkProps {
   toggleShowArticleSublinks: (e?: React.MouseEvent) => void;
   showArticleSublinks: boolean;
   parentId: string;
-  // The label to show the user, e.g. '2 sublinks'. Should be singular.
+  // The singular label to show the user, e.g. 'sublink'.
   sublinkLabel?: string;
 }
 
@@ -100,8 +100,7 @@ class Sublinks extends React.Component<SublinkProps> {
     );
   }
 
-  private dragEventNotDenylisted = (e: React.DragEvent) =>
-    !denyDragEvent()(e);
+  private dragEventNotDenylisted = (e: React.DragEvent) => !denyDragEvent()(e);
 }
 
 export default Sublinks;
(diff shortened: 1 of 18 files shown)


fronts-client/src/lib/dnd/Level.tsx
```diff
@@ -72,9 +66,8 @@ interface OuterProps<T> {
   onDrop: (e: React.DragEvent, to: PosSpec) => void;
   renderDrag?: (data: T) => React.ReactNode;
   renderDrop?: (props: DropProps) => React.ReactNode | null;
-  // Any occurence of these in the data transfer will cause all dragging
-  // behaviour to be bypassed.
-  denylistedDataTransferTypes?: string[];
+  // If this function returns false, all dragging behaviour is bypassed.
+  denyDragEvent?: (e: React.DragEvent) => boolean;
```
jonathonherbert, line 70: This is now a function because we no longer what consumers will have to match on until render time.

fronts-client/src/util/collectionUtils.ts
```diff
@@ -37,27 +37,29 @@ export type MappableDropType =
...
-    [CardTypesMap.CHEF]: (data: string): ChefDrop => ({
-      type: 'CHEF',
-      data: JSON.parse(data),
-    }),
-    [CardTypesMap.FEAST_COLLECTION]: (data: string): FeastCollectionDrop => ({
-      type: 'FEAST_COLLECTION',
-    }),
-    text: (url: string): RefDrop => ({ type: 'REF', data: url }),
-  };
+const dropToCardMap = {
```
jonathonherbert, line 40: Moved to its own data structure so we could statically derive its keys, and get some type safety upstream.

fronts-client/src/util/dragAndDrop.ts
```diff
@@ -0,0 +1,17 @@
+import {
+  dragOffsetX,
+  dragOffsetY,
+} from 'components/FrontsEdit/CollectionComponents/ArticleDrag';
+import { CardTypesMap } from 'constants/cardTypes';
+import { CARD_TYPE } from 'lib/dnd/constants';
+import { InsertDropType } from './collectionUtils';
+
+export const handleDragStartForCard =
```
jonathonherbert, line 9: Helper, as this code was repeated in four different places.

fronts-client/src/components/FrontsEdit/CollectionComponents/Sublinks.tsx
```diff
@@ -98,8 +100,7 @@ class Sublinks extends React.Component<SublinkProps> {
     );
   }
 
-  private dragEventNotDenylisted = (e: React.DragEvent) =>
-    !dragEventIsDenylisted(e, collectionDropTypeDenylist);
+  private dragEventNotDenylisted = (e: React.DragEvent) => !denyDragEvent()(e);
```
fredex42, line 103: Not sure I understand what this is for - is in why the double-negative, and whether it would be clearer to remove the indirection through the method given it's just a single call?

fredex42 (approved):
One small query, otherwise LGTM!

fronts-client/src/components/FrontsEdit/CollectionComponents/Sublinks.tsx
jonathonherbert: Yes, I see no reason not to move to an inline method! The reflex to use a method may have been related to a common pattern when class components were used and arrow fns were useful for lexical scoping, but that doesn't apply here. 88ec333aa9483a59f59cb445b23d5e0e050e70d0

PR #1610, opened 2024-07-30
Integrating new backend recipe search!

fredex42:
## What's changed?

Initial integration of the new recipe search backend, currently does a "generic" recipe search

### To Do

- [ ] Fix up `fetchCardReferencedEntitiesForCards` (Collections.ts line 390) to load in state for recipes. This will need an un-authenticated backend to do the lookup, hence changes to the recipe-backend or recipe-search-backend
- [ ] Separate out "recipes by title/description" and "recipes by ingredient" and set up the search params appropriately (see https://github.com/guardian/recipe-search-backend/pull/13 for example)
- [ ] If a recipe card is linked to a `RecipeHit`, display the `score` field as a percentage
- [ ] Filter by diet (**do in another PR**)
- [ ] Search for contributors (**do in another PR**)

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fronts-client/src/components/feed/RecipeFeedItem.tsx
```diff
@@ -49,14 +53,19 @@ export const RecipeFeedItem = ({ recipe }: ComponentProps) => {
...
       hasVideo={false}
       isLive={true} // We do not yet serve preview recipes
       handleDragStart={handleDragStart}
       onAddToClipboard={onAddToClipboard}
       shouldObscureFeed={shouldObscureFeed}
-      metaContent={<ContentInfo>Recipe</ContentInfo>}
+      metaContent={<>
+        <ContentInfo>Recipe</ContentInfo>
+        <ContentExtra>
+          {recipe?.score && recipe.score<1 ?  `Relevance ${Math.ceil(recipe.score*100)}%` : ""}
```
jonathonherbert, line 66: Non-blocking, open question: do our users care about this? We're already seeing a sorted list of things. What does the user do with `Relevance: 82%`? (It feels useful for debugging – certainly something that might be useful to be switchable in `v2/features` for devs if we didn't include it on the main path)

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -55,16 +55,26 @@ enum FeedType {
 export const RecipeSearchContainer = ({ rightHandContainer }: Props) => {
   const [selectedOption, setSelectedOption] = useState(FeedType.recipes);
   const [searchText, setSearchText] = useState('');
-  const recipes: Record<string, Recipe> = useSelector((state: State) =>
-    recipeSelectors.selectAll(state)
-  );
+  // const recipes: Record<string, Recipe> = useSelector((state: State) =>
```
jonathonherbert, line 58:
```suggestion
```

fronts-client/src/services/__tests__/recipeQuery.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { updateImageScalingParams } from '../recipeQuery';
+
+describe("updateImageScalingParams", ()=>{
```
jonathonherbert, line 3: 👏 👏 !

fronts-client/src/services/recipeQuery.ts
```diff
@@ -0,0 +1,167 @@
...
+export interface DietSearchResponse {
+  "diet-ids": KeyAndCount[];
+}
+
+const widthParam = /width=(\d+)/;
+export const updateImageScalingParams = (url:string) => {
+  return url.replace(widthParam, "width=83")
+}
+
+const setupRecipeThumbnails = (recep:Recipe) => {
```
jonathonherbert, line 57: I gotta ask. This looks like a mispelling, but you're consistent with it. rec-E-P? As in, phonetic? 😅

jonathonherbert (approved):
Tested in CODE, looking great! A few non-blocking comments.

fronts-client/src/services/recipeQuery.ts
```diff
@@ -0,0 +1,167 @@
...
+  } catch(err) {
+    console.error(err);
+    return recep;
+  }
+}
+
+const recipeQuery = (baseUrl:string) => {
+  const fetchOne = async (href:string):Promise<Recipe|undefined> => {
+    const response = await fetch(`${baseUrl}${href}`);
+    const content = await response.json();
```
Divs-B, line 79: error handling or checking `response` we might need in case of `response` is of type of any error and not json?

fronts-client/src/services/recipeQuery.ts
```diff
@@ -0,0 +1,167 @@
...
+    }
+  } catch(err) {
+    console.error(err);
+    return recep;
+  }
+}
+
+const recipeQuery = (baseUrl:string) => {
+  const fetchOne = async (href:string):Promise<Recipe|undefined> => {
+    const response = await fetch(`${baseUrl}${href}`);
```
Divs-B, line 78: i think `pandafetch` method for fetch has been used widely, not sure if we can still use that here for consistent handling of these types of calls?

fronts-client/src/services/recipeQuery.ts
```diff
@@ -0,0 +1,167 @@
...
+  score: number;
+  href: string;
+}
+
+export type RecipeSearchHit = Recipe & {
+  score: number;
+}
+
+export interface DietSearchResponse {
+  "diet-ids": KeyAndCount[];
```
Divs-B, line 49: tiny one: can we try naming of params camelCase and not hyphen. not sure though if we need same name as coming in json?

fronts-client/src/types/Recipe.ts
```diff
@@ -19,6 +19,6 @@ export interface Recipe {
   title: string;
   canonicalArticle: string;
   difficultyLevel: string;
-  featuredImage: RecipeImage; // the latter is an old image format that appears in our test fixtures
+  featuredImage?: RecipeImage; // the latter is an old image format that appears in our test fixtures
```
Divs-B, line 22: any reason for making it optional? (sorry if I missed discussion relatd to this)

fronts-client/src/services/recipeQuery.ts
```diff
@@ -0,0 +1,167 @@
...
+    chefs: async (params:ChefSearchParams):Promise<ChefSearchResponse> => {
+      const args = [
+        params.query ? `q=${encodeURIComponent(params.query)}` : undefined,
+        params.limit ? `limit=${encodeURIComponent(params.limit)}` : undefined
+      ].filter(arg=>!!arg);
+
+      const queryString = args.length > 0 ? '?' + args.join("&") : "";
+      const url = `${baseUrl}/keywords/contributors${queryString}`;
+      const response = await fetch(url);
+      const content = await response.json();
```
Divs-B, line 112: same query as above that if we need checking response before?

Divs-B (changes requested):

fronts-client/src/components/feed/RecipeFeedItem.tsx
fredex42: Yes this is mainly for us and for the Data Science people to understand where to place a relevancy cut-off and to better understand why certain things are coming up.
I also think it's more useful if we start putting in new sort-ordering

fronts-client/src/services/__tests__/recipeQuery.spec.ts
fredex42: 😁 Yeah I noticed images were being v. slow to load and wondered if I had introduced a bug - turned out they _were_ very slow! For an obvious reason.

fronts-client/src/services/recipeQuery.ts
fredex42: Yeah it's just an abbrieviation that I like to use, as you know I type fast and it's really easy to mis-spell "recipe" as well as it being longer so for code I prefer `recep` (`ress - ep`) as it still conveys the meaning (I think!) and slows me down less.

fronts-client/src/services/recipeQuery.ts
fredex42: Heh, heh..... yeah that is a sticking point here.  We _do_ need the same name that's coming from the JSON and _that_ is coming _directly_ from the Elasticsearch index (it's an aggregation response). I forget the reason it's hyphenated there - I think it probably dates all the way back to Hatch - but we are kinda stuck with it for now I'm afraid 🤷

fronts-client/src/services/recipeQuery.ts
fredex42: No need for `panda` as this is an open API (intended for use by the app down the line)

fronts-client/src/types/Recipe.ts
fredex42: The reason I made it optional was that I came across a recipe in the "real" data where it was not set.  Now, I know that _shouldn't_ happen - but nonetheless it has.  Therefore, we need to code to deal with it because the lack of the field caused an exception.

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -55,16 +55,26 @@ enum FeedType {
...
+  //   recipeSelectors.selectAll(state)
+  // );
```
fredex42, line 60:
```suggestion
```

Divs-B:
thanks @fredex42 for your feedback on the comments. they are useful as got better perpective of implementation.
ready to go 👍

Divs-B (approved):

PR #1614, opened 2024-08-02
Update documentation on permissions

davidfurey:
## What's changed?

Makes it clear that CODE access can be granted by engineers on the team, and does not require CP or an EM.

I've chosen to call the team ed tools because I believe this term is still well understood and will outlive the current team name


paperboyo (approved):
👍

PR #1620, opened 2024-08-12
Fix for search breakage

fredex42:
## What's changed?

Previously, if any recipe failed to download, the entire search operation would abort.

Now it's fault-tolerant so a warning is logged but the app still works.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


jonathonherbert (approved):
Confirmed that this works by testing locally, blocking a single request, seeing the failure mode as described, and switching to this branch, where other recipes appear as normal. Be nice to have a test!

jonathonherbert (approved):

PR #1622, opened 2024-08-13
API endpoint to add collection to Front (editions/feast only)

jonathonherbert:
## What's changed?

Adds an API endpoint to add a new collection to a Front, at

`PUT editions-api/front/:frontId/collection?name=optional-name`

## Implementation notes

- I've factored out the nested row aggregation logic common to issue and fronts queries in dd57774336a9b7bba5066d153e1e3fd98514d354. Hopefully it's clear what the intermediate case class is used for.
- There are a few ways that DB queries can go wrong, and I was loathe to continue the tradition of modelling all of these failure modes in `Option[Result]`, so I've added three error states for missing things, logic invariants, and write errors. This enables the API to respond with the correct status code. Hopefully that looks useful and isn't overkill – feedback welcome!

## How to test

- The automated tests should pass. They should cover happy and unhappy paths. A few unhappy paths are difficult to test because of the transactional nature of the operation (they should never fail!)
- Try pinging the endpoint locally or in CODE (you'll need to borrow a cookie from a valid browser-based request – I'd recommend importing a valid API call with 'copy to cURL' and Postman). The endpoint should work as expected.

Tested locally with cURL, and verified in client.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/services/editions/db/CollectionsQueries.scala
+++ b/app/services/editions/db/CollectionsQueries.scala
@@ -193,8 +193,13 @@ trait CollectionsQueries {
 
   private case class GetCollectionsRow(collection: EditionsCollection, card: Option[DbEditionsCard])
 
-  protected def insertCollection(frontId: String, collectionIndex: Int, name: String, now: OffsetDateTime, user: User)(implicit session: DBSession) = 
-    Try { 
+  /**
+    * Insert a collection owned by the specified front.
+    *
+    * @return the Collection id
+    */
+  protected def insertCollection(frontId: String, collectionIndex: Int, name: String, now: OffsetDateTime, user: User)(implicit session: DBSession) =
+    Try {
       sql"""
         INSERT INTO collections (
           front_id,
--- a/app/services/editions/db/CollectionsQueries.scala
+++ b/app/services/editions/db/CollectionsQueries.scala
@@ -10,6 +10,8 @@ import services.editions.DbEditionsCard
 import services.editions.prefills.CapiQueryTimeWindow
 import model.editions.EditionsFeastCollection
 import play.api.libs.json.Writes
+import com.gu.pandomainauth.model.User
+import scala.util.Try
 
 trait CollectionsQueries {
 
@@ -191,4 +193,34 @@ trait CollectionsQueries {
 
   private case class GetCollectionsRow(collection: EditionsCollection, card: Option[DbEditionsCard])
 
+  protected def insertCollection(frontId: String, collectionIndex: Int, name: String, now: OffsetDateTime, user: User)(implicit session: DBSession) = 
+    Try { 
+      sql"""
+        INSERT INTO collections (
+          front_id,
+          index,
+          name,
+          is_hidden,
+          updated_on,
+          updated_by,
+          updated_email
+        ) VALUES (
+          $frontId
+          , ${collectionIndex}
+          , $name
+          , FALSE
+          , $now
+          , ${EditionsDB.getUserName(user)}
+          , ${user.email}
+        )
+        RETURNING id;
+      """.map(_.string("id"))
+        .single
+        .apply()
+        .toRight(EditionsDB.WriteError("Could not write new collection to database"))
+    }
+      .toEither
+      .left
+      .map { error => EditionsDB.WriteError(error.getMessage()) }
+      .flatten
 }
--- a/app/services/editions/db/CollectionsQueries.scala
+++ b/app/services/editions/db/CollectionsQueries.scala
@@ -12,6 +12,7 @@ import model.editions.EditionsFeastCollection
 import play.api.libs.json.Writes
 import com.gu.pandomainauth.model.User
 import scala.util.Try
+import model.editions.CuratedPlatform.Editions
 
 trait CollectionsQueries {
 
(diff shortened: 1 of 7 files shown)


Divs-B (approved):
Looks great 👍  I have tested by wiring it with client side implementation for adding new collection. can see new collection added in Front data on clicking of a button.
(Will attach PR for client side impl once raised)

app/services/editions/db/CollectionsQueries.scala
```diff
@@ -191,4 +193,39 @@ trait CollectionsQueries {
...
+          index,
+          name,
+          is_hidden,
+          updated_on,
+          updated_by,
+          updated_email
+        ) VALUES (
+          $frontId
+          , ${collectionIndex}
+          , $name
```
fredex42, line 215: Oooh - looks like this might be vulnerable to SQL injection?  For example, what would happen if I called the endpoint with `?name=test, FALSE, \"2024-01-02T03:04:05Z\", \"evil-hacker\", \"evil-hacker@pwned.net\"); drop table collections;` ? (appropriately URL-encoded, of course!)

Unless the `sql` modifier on the string is doing proper placeholder conversion of course?

app/services/editions/db/EditionsDB.scala
```diff
@@ -4,13 +4,50 @@ import java.time.{Instant, OffsetDateTime, ZoneOffset}
...
+      currentFront <- getFront(frontId).toRight(EditionsDB.NotFoundError(s"Front ${frontId} not found"))
+      collectionId <- insertCollection(
+        frontId = currentFront.id,
+        collectionIndex = collectionIndex.getOrElse(currentFront.collections.size),
+        name = name.getOrElse("New collection"),
+        user = user,
+        now = truncatedNow
+      )
+      updatedFront <- getFront(frontId).toRight(EditionsDB.InvariantError(s"Updated front ${frontId} not found in issue"))
+      newCollection <- updatedFront.collections.find(_.id == collectionId).toRight(EditionsDB.InvariantError(s"New collection ${collectionId} not found in updated front ${frontId}"))
```
fredex42, line 35: do we need `newCollection` here?

app/services/editions/db/FrontsQueries.scala
```diff
@@ -59,4 +73,86 @@ trait FrontsQueries extends Logging {
...
+          collections.path_type     AS collections_path_type,
+          collections.content_prefill_window_start       AS collections_content_prefill_window_start,
+          collections.content_prefill_window_end         AS collections_content_prefill_window_end,
+
+          cards.collection_id AS cards_collection_id,
+          cards.id            AS cards_id,
+          cards.card_type     AS cards_card_type,
+          cards.index         AS cards_index,
+          cards.added_on      AS cards_added_on,
+          cards.metadata      AS cards_metadata
```
fredex42, line 111: that's a lorra lorra important strings - I guess it's not possible to select on a `.` at line 119 and that's why we have to rename them all?

app/services/editions/db/IssueQueries.scala
```diff
@@ -201,42 +194,22 @@ trait IssueQueries extends Logging {
...
-            DbEditionsCard.fromRowOpt(rs, "cards_"))
-        }
+        val maybeIssue = EditionsIssue.fromRow(rs).toOption
+        val frontRow = FrontAndNestedEntitiesRow(
+          DbEditionsFront.fromRowOpt(rs, "fronts_"),
+          DbEditionsCollection.fromRowOpt(rs, "collections_"),
+          DbEditionsCard.fromRowOpt(rs, "cards_")
+        )
+
+        maybeIssue.map(issue => (issue, frontRow))
```
fredex42, line 204: Intrigued why this is now needed? (not saying it _isn't_, just intrigued!)

fredex42 (commented):
Cool - got a couple of questions inline though 😁

app/services/editions/db/CollectionsQueries.scala
jonathonherbert: _> Unless the sql modifier on the string is doing proper placeholder conversion of course?

👍 😀   [(docs)](https://scalikejdbc.org/documentation/sql-interpolation.html)

app/services/editions/db/EditionsDB.scala
jonathonherbert: Ooh, no, thanks! 
```suggestion
      _ <- updatedFront.collections.find(_.id == collectionId).toRight(EditionsDB.InvariantError(s"New collection ${collectionId} not found in updated front ${frontId}"))
```

app/services/editions/db/IssueQueries.scala
jonathonherbert: IIUC, this is related to 

> I've factored out the nested row aggregation logic common to issue and fronts queries in https://github.com/guardian/facia-tool/commit/dd57774336a9b7bba5066d153e1e3fd98514d354

in that `GetIssueRow` is replaced by `FrontAndNestedEntitiesRow` (decoupling the issue data in this ResultSet from the Fronts data), so we can use `FrontAndNestedEntitiesRow` elsewhere – does that help to explain? May have wrong end of stick!

app/services/editions/db/FrontsQueries.scala
jonathonherbert: I know eh 😢  – each joined table is scoped to its entity to prevent collisions. We could share this string with `IssueQueries`, though! Not sure I understand the `.` suggestion, but keen to reduce repetition if poss.

app/services/editions/db/CollectionsQueries.scala
fredex42: 😌

fredex42 (approved):
Can discuss the stuff about field-naming in person if you like, it's nothing to hold up right now 😀

PR #1623, opened 2024-08-13
API endpoint to remove collection from front (editions/feast only)

jonathonherbert:
## What's changed?
Building on #1622, adds an endpoint to remove a collection from a front, at 

`DELETE editions-api/front/:frontId/collection/:collectionId`

I think it makes sense to scope this action to the front, as we're removing a collection from the front, and reindexing the other collections as a result. It also means we can pass back the updated front to ensure the client can read its own writes.

## Implementation notes

- The automated tests should pass. They should cover happy and unhappy paths. A few unhappy paths are difficult to test because of the transactional nature of the operation (they should never fail!)
- Try pinging the endpoint locally or in CODE (you'll need to borrow a cookie from a valid browser-based request – I'd recommend importing a valid API call with 'copy to cURL' and Postman). The endpoint should work as expected.

Tested locally with cURL, and verified in client.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/EditionsController.scala
+++ b/app/controllers/EditionsController.scala
@@ -268,6 +268,19 @@ class EditionsController(db: EditionsDB,
     }
   }
 
+  def removeCollectionFromFront(frontId: String, collectionId: String) = EditEditionsAuthAction { req =>
+    db.removeCollectionFromFront(
+      frontId = frontId,
+      collectionId,
+      user = req.user,
+      now = OffsetDateTime.now()
+    ) match {
+      case Right(front) => Ok(Json.toJson(front))
+      case Left(EditionsDB.NotFoundError(message)) => NotFound
+      case Left(error) => InternalServerError(error.getMessage())
+    }
+  }
+
   private def getAvailableCuratedPlatformEditions: Map[String, List[CuratedPlatformDefinition]] = {
     val feastAppEditions = FeastAppTemplates.getAvailableTemplates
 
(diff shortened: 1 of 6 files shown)


app/controllers/EditionsController.scala
```diff
@@ -268,6 +268,19 @@ class EditionsController(db: EditionsDB,
     }
   }
 
+  def removeCollectionFromFront(frontId: String, collectionId: String) = EditEditionsAuthAction { req =>
```
fredex42, line 271: Question - do you think we should prevent the removal of a non-empty collection?

app/services/editions/db/CollectionsQueries.scala
```diff
@@ -143,6 +144,32 @@ trait CollectionsQueries {
...
+  Try {
+    collectionIds match {
+      case Nil => Right(())
+      case ids =>
+        sql"""
+          UPDATE collections
+          SET index=CASE
+            ${sqls.join(collectionIds.zipWithIndex.map {
+              case (id, index) => sqls"""WHEN id=${id} THEN ${index}"""
+            }, sqls.empty)}
```
fredex42, line 163: I am not familiar with how collections are currently ordered - it seems possible that this could re-order _any_ of them, depending on what the order of the incoming `collectionIds` is?

fredex42 (commented):
again, just got a couple of questions

app/services/editions/db/CollectionsQueries.scala
jonathonherbert: When you say any – do you mean any collectionId in the list (which is the intent), or any collectionId ever? The `WHERE` on L165 should prevent the latter.

fredex42: 👍

fredex42 (approved):
👍

PR #1627, opened 2024-08-14
Add option to create issue from last issue for Feast users

jonathonherbert:
## What's changed?

Gives an option to use the last available issue prior to the date of the issue being inserted when creating Feast issues.

copy

## Implementation notes

This is accomplished by finding the most recent issue prior to the selected date, and recreating it as a `Skeleton`, which strips unique information (e.g. uuids) from it. (The `Skeleton` types are there to create issues from templates.) The issue skeleton is then inserted as normal.

## How to test

- The automated tests should pass. They cover the happy path, and ensure that we don't get any duplicate data from issues prior to the select issue. They also cover the case where an issue isn't found. (This should be rare!)
- Have a play! Try creating an issue where no prior issue exists – you should get a reasonable error message:

Screenshot 2024-08-15 at 17 20 28

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


app/controllers/EditionsController.scala
```diff
@@ -110,17 +130,28 @@ class EditionsController(db: EditionsDB,
...
       (LocalDate.parse(dateFrom), LocalDate.parse(dateTo))
-    }.map {
-      case (localDateFrom, localDateTo) =>
-        db.listIssues(edition, localDateFrom, localDateTo) match {
-          case Right(issues) => Ok(Json.toJson(issues))
-          case Left(errors) => InternalServerError(s"Error listing issues: ${errors.mkString(", ")}")
-        }
-    }.getOrElse(BadRequest("Invalid or missing date"))
+    }.toEither.left.map { _ =>
+      BadRequest("Invalid or missing date")
```
emdash-ie, line 138: Should we include the details of the error here?

emdash-ie (approved):
Looks good! I haven’t tested it in CODE, but I can see you have tested it thoroughly so I’m happy.

app/controllers/EditionsController.scala
jonathonherbert: Yes, that's nicer! e95e5d86e7b9c9385fc79b0214fb9deade80bb61

app/controllers/EditionsController.scala
```diff
@@ -110,17 +130,28 @@ class EditionsController(db: EditionsDB,
   }
 
   def listIssues(edition: Edition) = EditEditionsAuthAction { req =>
-    Try {
+    val params = Try {
       val dateFrom = req.queryString.get("dateFrom").map(_.head).get
       val dateTo = req.queryString.get("dateTo").map(_.head).get
       (LocalDate.parse(dateFrom), LocalDate.parse(dateTo))
```
emdash-ie, line 136: Now that I think about it, this might be a another good usecase for [Validated](https://typelevel.org/cats/datatypes/validated.html) – I guess it’s only two dates, but if there’s a problem with both of them it’d be nice to be able to report the two errors instead of just the first.

jonathonherbert: Agreed! I do wonder whether getting these sorts of libraries in early means there's not this general slightly suboptimal approach to validation across the board. I'm 🚢  now due to time constraints (thereby extending this problem!) 😅

emdash-ie: Yup, no problem! Definitely more of a future option to look at than something to do in this PR.

PR #1626, opened 2024-08-14
Add new collection at clientside

Divs-B:
## What's changed?

This will help in adding new Collection on the Front (in Feast Edition).
This wire up the server-side implementation that are done in PR(s) #1622 and #1623 with the client.

Co-authored by @jonathonherbert 

## Implementation notes

Clicking on `Add New Collection` should add one new collection at a time, at the bottom of the existing collection list. 
We are requesting server to send us back the updated Front json response (that has newly generated Collection schema in it) which we are later passing back to state to reflect newly added collection on the screen.

### General
- [ ] 🤖 Relevant tests added
- [ ] ✅ CI checks / tests run locally
- [ ] 🔍 Checked on CODE

### Client
- [ ] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ ] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #1631, opened 2024-08-23
Remove collection implementation at client side

Divs-B:
## What's changed?
This will help in removing collection from the Front (in Feast Edition).
This wire up the server-side implementation that are done in https://github.com/guardian/facia-tool/pull/1623 with the client.

`Delete` button is now present in place of `Hide` button, this is only in Feast Edition. 
Clicking on the button should remove the particular collection at a time and behind the screen we are requesting server to send us back the updated Front json response which we are later passing back to state to reflect updated collection on the screen with removed collection.

## Implementation notes
For implementing the changes commonly, followed PR #1626  (that was done under pairing with @jonathonherbert).

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fronts-client/src/reducers/frontsReducer.ts
```diff
@@ -63,6 +63,20 @@ const reducer = (
         newState
       );
     }
+    case 'FETCH_FRONT_REMOVE_COLLECTION_SUCCESS': {
```
jonathonherbert, line 66: I can't see that this action is dispatched anywhere! Do we really need this case (and the above case), and the accompanying action types in `types/Actions.ts`?

jonathonherbert (commented):
A few minor things before we 🚢. There's not a lot of logic in this PR, but if you'd like to add a test, too, that'd be great!

fronts-client/src/components/FrontsEdit/CollectionComponents/Collection.tsx
```diff
@@ -217,13 +222,24 @@ class Collection extends React.Component<CollectionProps, CollectionState> {
...
+                    {isHidden ? 'Unhide' : 'Hide'}
+                  </HeadlineContentButton>
+                )}
+                {isFeast && (
+                  <HeadlineContentButton
+                    priority="default"
+                    onClick={() => this.removeFrontCollection()}
+                    title="Toggle the visibility of this container in this issue."
+                  >
+                    {isHidden ? 'Un-delete' : 'Delete'}
```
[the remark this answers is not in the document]
Divs-B: Absolutely not, thanks for checking, this happens when copy of code I did and forgot to update the relevant line , I will remove it now 😄👍

fronts-client/src/reducers/frontsReducer.ts
Divs-B: Yeah, infact that was my question i wanted to check with you as I didnt spot this action anywhere during running on browser. Thanks for confirming I will remove them 👍

Divs-B:
> A few minor things before we 🚢. There's not a lot of logic in this PR, but if you'd like to add a test, too, that'd be great!

I think I will get this in another PR to push these changes in first. Thanks @jonathonherbert

jonathonherbert (approved):

PR #1633, opened 2024-08-23
Chef search from recipes

fredex42:
## What's changed?

- Make search for chefs from recipe backend.  This means that:
  - we don't see anyone who has a tag but doesn't write recipes
  - chefs are also ordered by the number of recipes that they have written in the app
  - fuzzy matching is now present, so `ana` matches `anna` and `ama` (see the notes in https://github.com/guardian/recipe-search-backend/pull/16 for more details on this)

Chefs can be clipboarded, added, removed, as before.

test

## Implementation notes

- There are two ways of identifying contributors in Feast. The first is by profile tag; the second, for contributors who do NOT have a profile tag, is to put the name into the `byline` array.  At present, "byline" contributors **are ignored**. This will be addressed in a future PR.
- The presentation of the chef card has been changed to use `webTitle` for the display name instead of `firstname lastname`.  This is because it turns out that the `firstName` and `lastName` fields are not appropriately filled in for all tags - in fact, they are often downcased and the wrong way around (i.e. last in `firstName`, first in `lastName`). Chatting with CP it transpired that these are old tags from before these fields were mandatory in the tooling - and in fact they are not set at all in the tags tool (therefore something, somewhere, must be auto-filling those values for the benefit of other tooling).  Anyhow, `webTitle` is apparently the field that is "expected"  to be used for display.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/actions/__tests__/Collections.spec.ts
+++ b/fronts-client/src/actions/__tests__/Collections.spec.ts
@@ -8,8 +8,8 @@ import {
   stateWithCollection,
   capiArticle,
   stateWithDuplicateArticleIdsInCollection,
-  stateWithCollectionWithChefs,
-  stateWithCollectionWithDuplicateChefs,
+  // stateWithCollectionWithChefs,
+  // stateWithCollectionWithDuplicateChefs,
 } from 'fixtures/shared';
 import {
   getCollectionsApiResponse,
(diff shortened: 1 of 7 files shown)


jonathonherbert (approved):
Tested locally, looks good! Search is pleasing. Response time makes me realise we have not got loading or empty states in the search, I'll card that up.

PR #1635, opened 2024-08-27
Start nginx if it isn't already running

jonathonherbert:
## What's changed?

Start nginx if it isn't already running, [per Composer,](https://github.com/guardian/flexible-content/blob/ada1811f3e82b500916e479506075192e21f7a5f/dev-start.sh#L31) as it hasn't been clear what's wrong when `dev-start.sh` was running but the tool was inaccessible.

### DIFF
--- a/scripts/dev-start.sh
+++ b/scripts/dev-start.sh
@@ -22,7 +22,17 @@ hasCredentials() {
   fi
 }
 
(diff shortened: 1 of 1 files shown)


codingWithRach (approved):
Excellent!

PR #1636, opened 2024-08-27
Fix slideshow drag validation

dblatcher:
## What's changed?

Fixes a bug in `Card.tsx`'s  `checkDraggedImage` method - if the dragged "image" was another card which uses a slideshow, that card may not have the `meta.imageSrcWidth` and  `meta.imageSrcHeight` properties, so the validation would, and the function will show the "no replacement image found" alert to the user.

Now, if `meta.imageSlideshowReplace` is true for the card the image was dragged from, the method will validate using a new `validateSlideshowDimensions` function which:

- runs the existing `validateDimensions` function for every slide in the slideshow which has dimensions
- if all the results are passing, return a pass result ( ie `{ matchesCriteria: true }`)
- else, return a fail result with a message which says how many of the images have validation problems

## Implementation notes

`validateSlideshowDimensions` ignores slides (members of `CardType.meta.slideshow`) which do not have the `width` and `height` properties - these are typed as optional. I'm not sure if slides can in practise be created without dimensions, but if so, it seemed best to allow them to be copied by dragging.

## test:

- open a front,  
- create a card by dragging an item from the feed into a container slot
- click the 'slideshow' toggle for your new card
- add two or more images to the slideshow
- save you card
- try dragging the image element from your card onto another card

on main - you should get the "no replacement image found"
on this branch - you should be able to copy the slideshow to the other card.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


jonathonherbert (commented):
Would it be possible to add a test to prevent a regression? To avoid testing the entire component, perhaps a unit test for `checkDraggedSlideshow`, which looks pure and could perhaps be a helper fn?

dblatcher:
> 👏!
> 
> Would it be possible to add a test to prevent a regression? To avoid testing the entire component, perhaps a unit test for `checkDraggedSlideshow`, which looks pure and could perhaps be a helper fn?

Good point - I've moved the function to the validateImageSrc module and added a set of tests for it.

jonathonherbert (approved):
Tested locally, works as expected 👍

PR #1642, opened 2024-08-29
Move collection up and down, client side

jonathonherbert:
## What's changed?

Adds the ability to move a collection up and down.

move-collections

## Implementation notes

Previous implementations of client-side collection changes were relying on a response type from the server that was slightly faulty – the API was returning an `EditionsFront`, which contained `EditionsCollection`, not `EditionsClientCollection` instances, and thus `EditionsCard` not `EditionsClientCard` instances. This didn't matter when creating and deleting collections, because these were actions on empty or disappearing collections. When moving collections with content, however, this matters, because `EditionsCard` instances do not include a `cardType`, resulting in blank content.

This PR alters the backend to return `EditionsClientCollection[]` rather than `EditionsFront` across all the endpoints that alter collections in this way – we do not use the front data in any of those methods on rehydration, and there is no equivalent `EditionsClientFront`.

## How to test

- The automated tests should pass. They've been adjusted to reflect the new response types.
- Have a play moving collections up and down, perhaps in CODE. It should work as expected!

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/EditionsController.scala
+++ b/app/controllers/EditionsController.scala
@@ -10,7 +10,7 @@ import model.editions._
 import model.editions.templates.CuratedPlatformDefinition
 import model.forms._
 import net.logstash.logback.marker.Markers
-import play.api.libs.json.{JsObject, Json}
+import play.api.libs.json.{JsObject, Json, OFormat}
 import play.api.mvc.Result
 import services.Capi
 import services.editions.EditionsTemplating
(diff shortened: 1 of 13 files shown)


Divs-B (approved):
Tested locally and its working very smoothly. 
Good to see greyed out button if collection is at top or last, it is going very well without any issue even if we add new or remove the collections. 
great! 🙌

Divs-B:
The up/down icons are similar to open/close icons of collection but i think its all looking good, if someones picks it out to be different for both we can sort this out in future PR

PR #1647, opened 2024-08-30
Tidy up templates

jonathonherbert:
Removes templated collections from Feast templates – they're no longer necessary, as we can add and remove collections on the fly.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE


fredex42 (approved):

Divs-B (approved):
makes sense 👍

PR #1649, opened 2024-09-01
Only create issues from previous issues of that edition

jonathonherbert:
## What's changed?

When attempting to create an issue from a previous issue, filter by `edition` name to avoid creating e.g. Feast issues from Editions app issues 🤦 

## How to test

- The automated tests should be pass – I've added a test for this case.
- Try creating an issue from a previous Feast issue, where an Editions App issue would be more recent. On `main`, you should find your new issue populated with the Editions App issue. On this branch, you should find that the Editions App issue is ignored for the correct Feast issue.

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fredex42 (approved):
Good catch 🏏

PR #1657, opened 2024-09-17
Batch resolve recipes

fredex42:
## What's changed?

We have been having problems on the recipe search backend when we fire hundreds and hundreds of small requests at it. To avoid needing a lot of scaling, it's easier to use the APIGW endpoint to resolve the recipes. That, however needs an API key.

This PR adds an endpoint into the backend which proxies the requests through to the API backend.  It requires https://github.com/guardian/recipes-backend/pull/64 to be running in recipes-backend in order to have something to talk to.

It then updates the frontend to resolve batches of IDs through this endpoint rather than via the recipe search backend.  This should save us money in the short term as we don't have to worry about large load spikes as much.

This PR also adds a new script which can be used to generate CODE-compatible Feast fronts in either CODE or LOCAL.  It contains a list of potential titles (generated by ChatGPT 😉 ) which are then used as a search term on the recipe backend. Some results are dodgier than others, but it can generate a realistic load level which is important when testing this.

```
./recipe-front-generator.mjs
    --stage CODE
    --fronts-issue-id {your-id}
    --front-name 'All Recipes'
    --cookie "$AUTH_COOKIE"
    --collection-count 7
```

📣 **NOTE** Before merging ensure that the config keys are placed into `[facia-tool.application.secrets.PROD.conf]`

Keys have already been updated for LOCAL and CODE configurations

## How to test

- Deploy to CODE
- Ensure that the corresponding recipes-backend PR is also deployed to CODE
- Create a new Feast Issue in CODE
- Grab the issue ID from the address bar and the cookies from the Network tab (guAuth-assym is the one you're after; the others don't hurth though)
- Run the command as above. This will add 7 containers each with between 4 and 40 recipes in them.  If you're lucky the recipes will even be more than vaguely relevant (but then it _is_ CODE).  If you want do the same for 'Meat-Free' (if you're really enthusiastic, update the code to only allow vegetarian recipes on there 😁 )
- Reload the page to open the entire front. You'll see it takes a few seconds to actually properly load them in, but they will all appear.
- Compare to the behaviour with the `main` branch deployed, where it's hit-and-miss whether all of the recipes load or not.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


scripts/recipes/recipe-front-generator.mjs
```diff
@@ -0,0 +1,344 @@
...
+};
+
+class ContinueOnError extends Error {
+
+}
+
+const getFrontsUri = () => {
+    switch(stage.toLocaleUpperCase()) {
+        case "PROD":
+            throw new Error("Don't run this against PROD")
```
JustinPinner, line 151: Ah. Question answered!

scripts/recipes/recipe-front-generator.mjs
```diff
@@ -0,0 +1,344 @@
...
+        case "CODE":
+            return "https://fronts.code.dev-gutools.co.uk";
+        case "LOCAL":
+            return "https://fronts.local.dev-gutools.co.uk";
+        default:
+            throw new Error("--stage must be one of PROD, CODE or LOCAL")
```
JustinPinner, line 157: Maybe "must be one of CODE or LOCAL"?

JustinPinner (approved):
Seems reasonable to me 👍

scripts/recipes/recipe-front-generator.mjs
```diff
@@ -0,0 +1,344 @@
...
+    "Decadent Dinners: Treat Yourself Tonight",
+    "Savory Brunch Ideas for Lazy Mornings",
+    "Flour Power: Master the Art of Baking",
+    "Festive Feasts: Holiday Recipes for Celebration",
+    "Healthy Starts: Energizing Breakfast Recipes",
+    "Global Grains: A World of Delicious Grains",
+    "Perfect for Sharing: Family-Style Meals",
+    "Sweet & Savory Creations for Any Mood"
+]
+const recipeBase = "https://recipes.code.dev-guardianapis.com";
```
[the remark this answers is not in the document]
fredex42: Exactly. The intention is to use it to populate CODE data

PR #1660, opened 2024-09-19
add recrop button for trail images

twrichards:
_Related https://github.com/guardian/grid/pull/4331_

There appeared to be glaring omission in the functionality around 'overriding trail images' since the existing button only let you 'replace image' (taking you to grid homepage/search in the modal), unlike say composer's 'recrop button' on image elements which at least takes you to the picture that's already selected, meaning that to tweak the crop used for a trail image, the same image would need to be re-found amongst the millions in the grid.

## What's changed?
In this PR we add a new 'Recrop image' button alongside the 'Replace image' button to provide the missing functionality...
| Before | After |
| --- | --- |
| image | image |

... this takes the user to that specific image in crop mode (ready to create a new crop). Thanks to @waisingyiu's #1601 if the container uses 5:4 trails then the crop options will be limited to 5:4 (otherwise restricted to 5:3 when overriding trail images).

It now also passes a `seedCropId`, in anticipation of grid using it (https://github.com/guardian/grid/pull/4332) for the starting position/dimensions of the new crop (particularly useful when re-cropping a 5:3 into a 5:4).

Although, prompted by the 5:4 trail images situation, this feature is useful in its own right.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/inputs/InputImage.tsx
+++ b/fronts-client/src/components/inputs/InputImage.tsx
@@ -24,6 +24,7 @@ import {
   AddImageIcon,
   VideoIcon,
   WarningIcon,
+  CropIcon,
 } from '../icons/Icons';
 import imageDragIcon from 'images/icons/image-drag-icon.svg';
 import {
@@ -49,9 +50,10 @@ const AddImageButton = styled(ButtonDefault)<{ small?: boolean }>`
       small ? theme.colors.greyVeryLight : '#5e5e5e99'};
   }
   width: 100%;
-  height: 100%;
+  flex-grow: 1;
   padding: 0;
   text-shadow: 0 0 2px black;
+  display: inline-block;
 `;
 
 const ImageComponent = styled.div<{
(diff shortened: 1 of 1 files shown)


paperboyo:
Yay, thank you!

> if the container uses 5:4 trails then the crop options will be limited to 5:4 (otherwise restricted to 5:3 when overriding trail images)

And if container uses 4:5, it restricts to 4:5, right? (sorry for pedantism!)

twrichards:
> Yay, thank you!
> 
> > if the container uses 5:4 trails then the crop options will be limited to 5:4 (otherwise restricted to 5:3 when overriding trail images)
> 
> And if container uses 4:5, it restricts to 4:5, right? (sorry for being pedantic!)

It should yes. My comment was more an example that it preserves whatever logic was there already.

dblatcher (approved):
LGMT!
Not one for this PR, but it might be worth adding a recrop button for replacement images as well as on the original trail image.

PR #1662, opened 2024-09-19
Fix migration bug for missing data in the collection

Divs-B:
## What's changed?

We have recently encountered an issue on migration that one particular collection was missing the data (recipes) in it.
The collection title was `Season finale: pickles & preserves to make with late summer produce` in `curation.json` for date 14th Sept 24 whereas our migrated Front was showing empty collection with title `Season finale: pickles` for the same.
Before: 0 items in that collection 
image
After: 18 items in that collection
image

Reason was around new Collection creation via `fetch` call with one of parameter as `name=${title}` at the end, in this case, the part of the `title` was treated as `query string` and was getting dropper because of presence of `&` inside it.

The fix will encode the `title` to be considered as a valid part of the URI to avoid such case.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/scripts/recipes/insert-backdate-recipes.mjs
+++ b/scripts/recipes/insert-backdate-recipes.mjs
@@ -4,7 +4,7 @@ import { migrateIssue } from './migrate-issue.mjs';
 const dateFormat = 'YYYY-MM-DD';
 const usage = `Example usage is
 
-node ./insert-recipe-cards.mjs
+node ./insert-backdate-recipes.mjs
     --curation-path "northern"
     --from-date "2024-05-01"
     --to-date "2024-05-05"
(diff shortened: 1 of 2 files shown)


fredex42 (approved):
😀 Great find. Are there any other params that need url-encoding while we are at it? (can't see any on that line, just wondering if there might be others lurking somewhere)

Divs-B:
> 😀 Great find. Are there any other params that need url-encoding while we are at it? (can't see any on that line, just wondering if there might be others lurking somewhere)

Thanks @fredex42 for reviewing and suggesting for checking other places. I think this as of now should be the one where encoding needed so should be fine. thank you.

PR #1667, opened 2024-09-24
Fix frontend crash on chef search

fredex42:
## What's changed?

A user reported that the screen went white (frontend crash) when she clicked on the 'Chefs' search radio button.

Sentry traces showed that this was because of an unexpected `undefined` value for the chef id when rendering a chef card - https://the-guardian.sentry.io/issues/5820707430/?project=35467&referrer=issue-stream&statsPeriod=90d&stream_index=0&utc=true

Screenshot 2024-09-24 at 12 55 16

## Implementation notes
This obviously _shouldn't_ happen, but it did, so I added a `filter` before the `map` to strip out any null/undefined values before attempting to render cards

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/feed/RecipeSearchContainer.tsx
+++ b/fronts-client/src/components/feed/RecipeSearchContainer.tsx
@@ -117,7 +117,9 @@ export const RecipeSearchContainer = ({ rightHandContainer }: Props) => {
           <RecipeFeedItem key={id} id={id} />
         ));
       case FeedType.chefs:
-        return chefSearchIds.map((chefId) => (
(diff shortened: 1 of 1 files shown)


Divs-B (approved):
When any Chefs is undefined? i assume when any its deleted from TagManager? I feel when will be the scenerios when one contributor is getting deleted.
But this change anyway makes sense to filter non-existing values so good to merge. 👍

fredex42:
> When any Chefs is undefined? i assume when any its deleted from TagManager? I feel when will be the scenerios when one contributor is getting deleted. But this change anyway makes sense to filter non-existing values so good to merge. 👍

Genuinely, I have no idea. We _shouldn't_ enter into this state..... but we did somehow. As it was causing a problem for the users I figured it was better to do the quick fix rather than take ages trying to find the root cause

PR #1669, opened 2024-09-25
UI indication for no recipe

Divs-B:
## What's changed?

At present if the recipe is not available on the FRONT because of mostly it being taken down or rarely due to parsing or any other error, then current presentation is normal means can easily get un-noticed for which Editorial users can get confused why certains recipes are not available on Apps after they do publish.

This PR will change the ui presentation with clear message to get noticed by the editorial users for the recipes that are not found.

**Before:**

image

**After**:
**1)**

image

**2)**

image

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client
- [X] 📷 Screenshots / GIFs of relevant UI changes included


fronts-client/src/components/card/CardContent.tsx
```diff
@@ -1,9 +1,11 @@
 import styled from 'styled-components';
 import { CardSizes } from 'types/Collection';
+import { theme } from '../../constants/theme';
 
 const CardContent = styled.div<{
   displaySize?: CardSizes;
   textSize?: CardSizes;
+  isContentNotFound?: boolean;
```
fredex42, line 8: I would suggest to specify the "warning" state as a seperate CSS class (something like

```typescript
const NotFoundWarningStyle = css`
  background-color: theme.colors.greyMediumLighter;
...
 `

```

and then link that into the `CardContent` div below using the `css` prop, i.e.

```tsx
<CardContent css={isContentNotFound ? NotFoundWarningStyle : undefined}>

```

That makes it much easier to evaluate in test classes, because you can inspect the class-names on the div from within your test

fronts-client/src/components/card/CardHeading.tsx
```diff
@@ -8,9 +8,11 @@ import { CardSizes } from 'types/Collection';
 const Wrapper = styled.span<{
   displaySize?: CardSizes;
   showLargeHeadline?: boolean;
+  isContentNotFound?: boolean;
 }>`
   font-family: TS3TextSans;
-  font-weight: normal;
+  font-weight: ${({ isContentNotFound }) =>
+    isContentNotFound ? 'bold' : 'normal'};
   padding: 2px 0 0;
```
fredex42, line 16: As above

fronts-client/src/components/card/recipe/RecipeCard.tsx
```diff
@@ -63,27 +64,48 @@ export const RecipeCard = ({
...
           <CardHeadingContainer size={size}>
-            <CardHeading data-testid="headline" html>
-              {recipe?.title ?? 'No recipe found'}
+            <CardHeading
+              data-testid="headline"
+              html
+              isContentNotFound={isContentNotFound}
+            >
+              {recipe?.title ??
+                'This recipe won’t load in the app, please select an alternative.'}
```
fredex42, line 101: Suggested wording change - from "won't load" to "might not load", as we can't be 100% sure it's not a bug on our side 🤭

fredex42 (commented):
Looking good, I'd like to request changing the "warning state" to its own CSS class.

That will then enable you to build a test-case where you can pass in an invalid ID and ensure that the "warning state" class is present on the item

fredex42 (approved):
Great! Sorry it took so long.... :grinning:

Divs-B:
Thankyou @fredex42 for tidy up and help in adding the test.

PR #1599, opened 2024-06-24
Add collection IDs to the SNS message

waisingyiu:
_co-authored-by:_ @twrichards 

## What's changed?

We have been exploring ways to put more information on fronts updates events to send to Heatphan via Fastly Fanout.

This pull request add the IDs of modified collections to SNS messages when fronts are updated.  We will have a pull request on Fastly Fanout (guardian/fastly-content-fanout#20) to extract these collection IDs and pass on to Fanout to include them in Server Sent Events.

## Implementation notes

It changes the SNS messages in the backend so it is not expected to affect the front end of the fronts tool and change any user-facing behaviour.

### General
- [ ] 🤖 Relevant tests added
- [X] ✅ CI checks / tests run locally
- [X] 🔍 Checked on CODE

### Client
- [X] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [X] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #1671, opened 2024-09-27
Filter out irrelevant collection IDs from front update events

waisingyiu:
@jonflynng noticed there were occasionally some unexpected collectionIds coming through the SNS events since #1599 - with some help from @davidfurey we realised facia press is sometimes triggered with a big list of collection IDs (e.g. in config tool, when for example collections are added/removed from a front) this in turn triggers SNS events for all fronts where the collections ids occur, which meant if any collections were shared then all those fronts would receive the full list of collection ids from the front where there were the actual changes - not helpful downstream.

## What's changed?
This PR attempts to address that by filtering the list of collectionIds for each event to only those actually present on the front that the event is for.

We've tested this (before and after this PR was deployed to CODE) by making a change to the order of collections on a front and seeing that the list of collectionIds delivered over SNS (one SNS event per front) is now limited to the changed collectionIDs which actually belong on the front in question.

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/services/FaciaPress.scala
+++ b/app/services/FaciaPress.scala
@@ -66,6 +66,8 @@ class FaciaPress(val faciaPressTopic: FaciaPressTopic, val configAgent: ConfigAg
         path <- configAgent.getConfigsUsingCollectionId(id)
       } yield path
 
+      val pathToCollectionIdsLookup = configAgent.getConfigCollectionMap
+
       def sendEvents(pressType: PressType) = Future.traverse(
         paths.filter(_ => pressType match {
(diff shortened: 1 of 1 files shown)


jonflynng (approved):

twrichards:
@jonflynng is this behaving itself now? (i.e. is the data you're receiving more what you were originally expecting when you started relying on this)

jonflynng:
@twrichards  Yep!

PR #1673, opened 2024-09-30
Fix - Chefs more than 10 does not get loaded on refresh

Divs-B:
## What's changed?

Solution suggested by: Andy Gallagher <fredex42@gmail.com>

If more than 10 chefs are present in a container, the 11th and subsequent tend to show as “No Chef Found”.
It has been found that the Fronts tool issues HTTP request proxied to CAPI for contributor tags on which the response is getting split across two pages due to pageSize being 10.
Solution to try pageSize to be 100 as we doubt we'll need more than 100 chefs in a container!

**Before**:

image

**After**:

image

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client
- [X] 📷 Screenshots / GIFs of relevant UI changes included

### DIFF
--- a/fronts-client/src/bundles/__tests__/chefsBundle.spec.ts
+++ b/fronts-client/src/bundles/__tests__/chefsBundle.spec.ts
@@ -8,124 +8,161 @@ interface MockedResponse {
   response: any;
 }
 
-const createStoreAndFetchMock = (fetchResponses:MockedResponse[]) => {
-  fetchResponses.forEach((r)=>{
-    fetchMock.once(r.pattern, r.response, {overwriteRoutes: true});
-  })
+const createStoreAndFetchMock = (fetchResponses: MockedResponse[]) => {
+  fetchResponses.forEach((r) => {
+    fetchMock.once(r.pattern, r.response, { overwriteRoutes: true });
+  });
   return configureStore();
 };
 
 const annaMockRecipe = {
-  "hits": 7,
-  "results": [
+  hits: 7,
+  results: [
     {
-      "contributorType": "Profile",
-      "nameOrId": "profile/anna-jones",
-      "docCount": 182
+      contributorType: 'Profile',
+      nameOrId: 'profile/anna-jones',
+      docCount: 182,
     },
     {
-      "contributorType": "Byline",
-      "nameOrId": "Anna Haugh",
-      "docCount": 6
+      contributorType: 'Byline',
+      nameOrId: 'Anna Haugh',
+      docCount: 6,
     },
     {
-      "contributorType": "Byline",
-      "nameOrId": "Anna Tobias",
-      "docCount": 4
+      contributorType: 'Byline',
+      nameOrId: 'Anna Tobias',
+      docCount: 4,
     },
     {
-      "contributorType": "Profile",
-      "nameOrId": "profile/anna-del-conte",
-      "docCount": 3
+      contributorType: 'Profile',
+      nameOrId: 'profile/anna-del-conte',
+      docCount: 3,
     },
     {
-      "contributorType": "Byline",
-      "nameOrId": "Anna Higham",
-      "docCount": 2
+      contributorType: 'Byline',
+      nameOrId: 'Anna Higham',
+      docCount: 2,
     },
     {
-      "contributorType": "Byline",
-      "nameOrId": "Anna Jones",
-      "docCount": 1
+      contributorType: 'Byline',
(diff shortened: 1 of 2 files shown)


fronts-client/src/bundles/chefsBundle.ts
```diff
@@ -33,52 +35,69 @@ export const fetchChefs =
...
 
-  //we could be called as the second part of a two-stage fetch, OR we could be called directly.
-  //if called directly, indicate that we started a fetch. Otherwise, the first part should
-  if(!alreadyStarted) dispatch(actions.fetchStart(tagIds));
+    try {
+      const chefTags = await liveCapi.chefs({
+        ids: tagIds.join(','),
+        'show-elements': 'image',
+        'show-fields': 'all',
+        'page-size': 100,
```
Divs-B, line 76: This is the only singe line change dedicated for this PR. 
Other all were Lint errors that got fixed at my IntelliJ ESLint and hence getting merged in here.

fronts-client/src/bundles/__tests__/chefsBundle.spec.ts
```diff
@@ -8,124 +8,161 @@ interface MockedResponse {
...
 
-  it("fetchChefsById should look up CAPI details for profile tags", async ()=>{
+  it('fetchChefsById should look up CAPI details for profile tags', async () => {
     const store = createStoreAndFetchMock([
       {
-        pattern: "/api/live/tags?type=contributor&ids=profile%2Fanna-jones%2Cprofile%2Fanna-del-conte&show-elements=image&show-fields=all",
-        response: tagLookupResponse
-      }
+        pattern:
+          '/api/live/tags?type=contributor&ids=profile%2Fanna-jones%2Cprofile%2Fanna-del-conte&show-elements=image&show-fields=all&page-size=100',
```
Divs-B, line 157: Another change for this PR: included pageSize.
Others are lint errors that got fixed by INtellJ.

fronts-client/src/bundles/__tests__/chefsBundle.spec.ts
```diff
@@ -8,124 +8,161 @@ interface MockedResponse {
...
+        response: annaMockRecipe,
       },
       {
-        pattern: "/api/live/tags?type=contributor&ids=profile%2Fanna-jones%2Cprofile%2Fanna-del-conte&show-elements=image&show-fields=all",
-        response: tagLookupResponse
-      }
-      ]);
-    await store.dispatch(fetchChefs({query: "anna"}) as any);
+        pattern:
+          '/api/live/tags?type=contributor&ids=profile%2Fanna-jones%2Cprofile%2Fanna-del-conte&show-elements=image&show-fields=all&page-size=100',
```
Divs-B, line 142: same here

fronts-client/src/bundles/chefsBundle.ts
JustinPinner, line 76: Would you not want the `page-size` here to use the `pageSize` parameter value? Or perhaps `pageSize` is an instruction to the renderer of the retrieved chefs, in which case ignore this comment :)

fredex42: hmmm I tend to agree, we should use the `pageSize` parameter value here and then explicitly set that when doing the initial load

Divs-B: Ya sure i will try to check now. Thanks both.

fronts-client/src/bundles/chefsBundle.ts
```diff
@@ -33,52 +35,69 @@ export const fetchChefs =
...
 
-  //we could be called as the second part of a two-stage fetch, OR we could be called directly.
-  //if called directly, indicate that we started a fetch. Otherwise, the first part should
-  if(!alreadyStarted) dispatch(actions.fetchStart(tagIds));
+    try {
+      const chefTags = await liveCapi.chefs({
+        ids: tagIds.join(','),
+        'show-elements': 'image',
+        'show-fields': 'all',
+        'page-size': pageSize,
```
Divs-B, line 76: Keeping signature of this method as it is except pageSize to be  default to 100 as the suggested fix for maximum chefs to get loaded.

fredex42 (approved):
👍 Great! Happy with this - please do the checks outlined in the PR description, though (no regressions, no obvious console warnings etc.), and I think we can argue that it _is_ covered by a test case since you had to add `limit=` to the expected URL in the tests

PR #1688, opened 2024-10-14
Downgrade ubuntu version to match PROD

emdash-ie:
## What's changed?

Github’s standard ubuntu version for Github Actions runners has recently changed to 24.04. This appears to have broken CI on this repository because it broke the ChromeHeadless usage in tests.

To fix it, we could revert to 22.04, which was being used before the recent change. However, it appears fronts is using 20.04 in PROD (based on [the amigo recipe](https://amigo.gutools.co.uk/recipes/editorial-tools-focal-java11-ARM-WITH-cdk-base) specified [in riff-raff.yaml](https://github.com/guardian/facia-tool/blob/6e0804dca3eed7a7d1dee8238fad8b0bba6d0e71/riff-raff.yaml#L14)). So instead I think it’s probably best to revert to 20.04.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/.github/workflows/improvedci.yml
+++ b/.github/workflows/improvedci.yml
@@ -10,7 +10,7 @@ on:
 jobs:
     generate_build_number:
         name: Generate build number
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-20.04
         outputs:
             BUILD_NUMBER: ${{ steps.create-build-number.outputs.BUILD_NUMBER }}
         steps:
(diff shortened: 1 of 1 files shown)


Fweddi (approved):

emdash-ie:
Unfortunately I missed the snyk workflow (which of course I should’ve realised uses sbt 🤦🏻 )! Will make a second PR shortly

PR #1693, opened 2024-10-14
WIP: Downgrade ubuntu to 22.04 in snyk workflow

emdash-ie:
### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -10,6 +10,7 @@ on:
 
 jobs:
   snyk-V2-and-scala:
+    runs-on: ubuntu-22.04
     uses: guardian/.github/.github/workflows/sbt-node-snyk.yml@main
     with:
       ORG: guardian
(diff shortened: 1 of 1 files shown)


emdash-ie:
Hmmm, Snyk failed when run on this branch: https://github.com/guardian/facia-tool/actions/runs/11326490208

Looks like I’ve used the wrong syntax. Maybe you can’t use `runs-on` with `uses`, and we actually just need the sbt-node-snyk workflow to update its ubuntu version?

emdash-ie:
If I’m reading https://docs.github.com/en/actions/sharing-automations/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow right, we can’t use `runs-on` when calling a reusable workflow (as we do here). So I think we’ll have to update [sbt-node-snyk](https://github.com/guardian/.github/blob/main/.github/workflows/sbt-node-snyk.yml) instead.

PR #1694, opened 2024-10-14
Explicit Sentry logging for error responses from the recipe search api

fredex42:
## What's changed?

Add explicit Sentry logging for error responses from the recipe search api.

If there are any transient issues with the search api, we need to know about them before we start trying to integrate to the main app. Since this is the only place it's in use at the moment, we want to get detailed information about any potential failures.

## Implementation notes

n/a

### General
- [ ] 🤖 ~~Relevant tests added~~ n/a
- [ ] 🔍 Checked on CODE

### Client
- [ ] 📷 ~~Screenshots / GIFs of relevant UI changes included~~ n/a

### DIFF
--- a/fronts-client/src/services/recipeQuery.ts
+++ b/fronts-client/src/services/recipeQuery.ts
@@ -1,5 +1,6 @@
 import url from '../constants/url';
 import { Recipe, RecipePartialIndexContent } from '../types/Recipe';
+import Raven from 'raven-js';
 
 interface KeyAndCount {
 	key: string;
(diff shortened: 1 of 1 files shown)


fronts-client/src/services/recipeQuery.ts
```diff
@@ -135,21 +144,21 @@ const recipeQuery = (baseUrl: string) => {
 			const queryString = args.length > 0 ? '?' + args.join('&') : '';
 			const url = `${baseUrl}/keywords/contributors${queryString}`;
 			const response = await fetch(url);
-			const content = await response.json();
+			const content = await response.text();
 			if (response.status == 200) {
-				return content as ChefSearchResponse;
+				return JSON.parse(content) as ChefSearchResponse;
```
JustinPinner, line 149: Is there an increased risk of json parse errors here? Should we wrap these in a try and catch the errors there too?

fredex42: In _principle_, no, because all 200 responses come from the (search) app and they are all definitely JSON (this is how it was working before, calling `await response.json()`).
On the other hand, it's not a bad idea just to be on the safe side.....

fredex42: Actually, no, I take that back; all of the _callers_ of these functions wrap in `try`/`catch` and report the error up that way, so we should not need to worry about it here.

JustinPinner (approved):
👍

PR #1699, opened 2024-10-16
Add pre-commit hook that runs prettier

emdash-ie:
## What's changed?

As a convenience, this PR adds a hook which checks the formatting of your files, formatting any which it’s safe to.

## Implementation notes

I’ve copied the one I implemented in https://github.com/guardian/datawrapper-import/pull/12 and tweaked it to account for the differences between prettier and scalafmt.

In particular, the hook is careful not to overwrite any unstaged changes: if a file has both staged and unstaged changes, running a formatter on it could (if something goes wrong) cause you to lose changes that git is unaware of, and they might be difficult to recover.

## To Test

- Run the setup script, or manually run `git config --local core.hooksPath scripts/git-hooks`.
- Test the behaviour of the hook with combinations of:
  - typescript files with only staged changes which need formatting
  - typescript files with only unstaged changes which need formatting
  - typescript files with staged and unstaged changes, which need formatting
  - non-typescript files with staged or unstaged changes

### DIFF
--- a/scripts/git-hooks/pre-commit
+++ b/scripts/git-hooks/pre-commit
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+
+if command -v fronts-client/node_modules/.bin/prettier > /dev/null; then
+    echo "Checking file formatting with prettier…"
+    echo ""
+    # For files which have staged changes and no unstaged changes, let prettier format them
+    filesWithOnlyStagedChanges=$(
+        comm -1 -3 \
+            <(git diff --name-only --diff-filter=ACMR | grep "\.tsx\?\$" | sort) \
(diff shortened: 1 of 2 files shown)


fredex42 (approved):
😎 where did you get the script from originally?

emdash-ie:
> 😎 where did you get the script from originally?

Nowhere, I wrote it! It’s my first time using `comm`. I think I ran shellcheck on the script too? Will double-check.

emdash-ie:
Ran shellcheck, no problems reported

emdash-ie:
Ok, ran shellcheck with `-oall` and got some extra complaints about masked return values in pipelines (e.g. of git and comm) – not sure whether I want to fix those or not.

PR #1701, opened 2024-10-17
Fix top elements disappearance when selecting Overview

Divs-B:
## What's changed?

We observed when we select any collection name in `Overview` the top elements present in the Front (above the containers) are getting disappeared means shifts at top and go out of the viewport.

### Before:

https://github.com/user-attachments/assets/ddd1a37d-f67d-4498-8d5a-bfa3c524c15e

### After

https://github.com/user-attachments/assets/65b59edd-2e22-43b9-91a3-2419e5a189ed

## Implementation notes
The block: 'nearest' option will scroll the element to the nearest edge of the viewport, instead of always forcing it to the very top.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fredex42 (approved):
🥳

PR #1704, opened 2024-10-21
Date ordering preference for recipes

fredex42:
## What's changed?

- Compatibility with https://github.com/guardian/recipe-search-backend/pull/48, which adds the ability to prefer most recent recipes in the search results
- Fix long-standing bug where the search results disappear when recipes are loaded in from the clipboard or a container
- Adds the option to completely override the feed item body in `FeedItem`
- Optionally shows the modified and last published times for recipes in the Feed

date_priority_search

## Implementation notes
See the recipe-search-backend PR for an explanation of what "prefer" actually means

### General
- [ ] 🤖 ~~Relevant tests added~~ - _None of the touched code currently has tests_
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/feed/RecipeFeedItem.tsx
+++ b/fronts-client/src/components/feed/RecipeFeedItem.tsx
@@ -67,7 +67,7 @@ export const RecipeFeedItem = ({ id, showTimes }: ComponentProps) => {
 					) : undefined}
 					{recipe?.publishedDate && showTimes ? (
 						<ContentExtra>
-							First published {renderTimestamp(recipe.publishedDate)}
+							Published {renderTimestamp(recipe.publishedDate)}
 						</ContentExtra>
 					) : undefined}
 				</>
--- a/fronts-client/src/components/feed/RecipeFeedItem.tsx
+++ b/fronts-client/src/components/feed/RecipeFeedItem.tsx
@@ -33,46 +33,46 @@ export const RecipeFeedItem = ({ id }: ComponentProps) => {
 		);
 	}, [recipe]);
 
-  const shortenTimestamp = (iso: string) => {
-    const parts = iso.split('T');
-    return parts[0];
-  };
+	const shortenTimestamp = (iso: string) => {
+		const parts = iso.split('T');
+		return parts[0];
+	};
 
-  return (
-    <FeedItem
-      type={CardTypesMap.RECIPE}
-      id={recipe.canonicalArticle}
-      title={recipe.title}
-      thumbnail={recipe.previewImage?.url ?? recipe.featuredImage?.url ?? ""}
-      liveUrl={`https://theguardian.com/${recipe.canonicalArticle}`}
-      hasVideo={false}
-      isLive={true} // We do not yet serve preview recipes
-      handleDragStart={handleDragStartForCard(CardTypesMap.RECIPE, recipe)}
-      onAddToClipboard={onAddToClipboard}
-      shouldObscureFeed={shouldObscureFeed}
-      metaContent={
-        <>
-          <ContentInfo>Recipe</ContentInfo>
-          <ContentExtra>
-            {recipe?.score && recipe.score < 1
-              ? `Relevance ${Math.ceil(recipe.score * 100)}%`
-              : ''}
-            <br />
-            {recipe?.lastModifiedDate
-              ? `M ${shortenTimestamp(recipe.lastModifiedDate)}`
-              : undefined}
-            <br />
-            {recipe?.publishedDate
-              ? `P ${shortenTimestamp(recipe.publishedDate)}`
-              : undefined}
-            <br />
-            {recipe?.firstPublishedDate
-              ? `F ${shortenTimestamp(recipe.firstPublishedDate)}`
-              : undefined}
-            <br />
-          </ContentExtra>
-        </>
-      }
-    />
-  );
+	return (
+		<FeedItem
+			type={CardTypesMap.RECIPE}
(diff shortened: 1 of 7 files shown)


fronts-client/src/components/inputs/HoverActionButtonWrapper.tsx
```diff
@@ -45,6 +45,7 @@ interface WrapperProps {
 	toolTipAlign: 'left' | 'center' | 'right';
 	urlPath: string | undefined;
 	renderButtons: (renderProps: ButtonProps) => JSX.Element;
+	noPinboard?: boolean;
```
emdash-ie, line 48: Calling this `noPinboard` feels like a bit of a double negative: `noPinboard={true}` takes me a second to think about. Could we make it `showPinboard` or `enablePinboard`?

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -84,20 +95,37 @@ export const RecipeSearchContainer = ({ rightHandContainer }: Props) => {
...
+					decay: 0.95,
+					dropoffScaleDays: 90,
+					offsetDays: 7,
+				};
+			case 'forceful':
+				return {
+					decay: 0.7,
+					dropoffScaleDays: 180,
+					offsetDays: 14,
+				};
```
emdash-ie, line 125: These values seem a bit surprising (e.g. why a bigger offset for forceful?), but I haven’t looked into it much and I assume you have.

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -84,20 +95,37 @@ export const RecipeSearchContainer = ({ rightHandContainer }: Props) => {
...
 	const chefsPagination: IPagination | null = useSelector((state: State) =>
 		chefSelectors.selectPagination(state),
 	);
 
 	const hasPages = (chefsPagination?.totalPages ?? 0) > 1;
 
+	const getUpdateConfig = () => {
+		switch (orderingForce) {
+			case 'default':
+				return undefined;
```
emdash-ie, line 113: To be more explicit about how unknown values are handled, could we add a default case here?

```suggestion
			case 'default':
			default:
				return undefined;
```

fronts-client/src/components/feed/RecipeSearchContainer.tsx
```diff
@@ -144,11 +180,99 @@ export const RecipeSearchContainer = ({ rightHandContainer }: Props) => {
...
+								onChange={(evt) => {
+									console.log(evt.target);
+									setDateField(
+										evt.target.value === 'Relevance'
+											? undefined
+											: (evt.target.value as DateParamField),
+									);
+								}}
+							>
+								<option value={'Relevance'}>
```
emdash-ie, line 215: Using an uppercase R here feels inconsistent. I suppose it’s signalling that this option isn’t a field name? I think I’d still prefer lowercase.

emdash-ie (approved):
A couple of questions, but happy for you to merge!

Divs-B:
I tested on `Most recent published` with search word `Salad` but result is not changing for me the one I got earlier is the same I am getting on selecting recent published too and it seems they are not in expected order for published?:

image

Network payload for search endpioint can be seen as : {queryText: "Salad", uprateByDate: "publishedDate", noStats: true}
and with response as:
image

Let me know if I have missed anything. thanks.
Note: this is I checked locally so pointing to CODE recipe.

fronts-client/src/components/feed/RecipeSearchContainer.tsx
fredex42: the bigger offset means that anything in the last 14 days won't get down-rated at all; it's therefore pushing recent content up the list more forcefully than dropping the score from 7 days.  It does behave in the intended way when testing though it's all a bit hard to get one's head around 😵‍💫

fronts-client/src/components/feed/RecipeFeedItem.tsx
```diff
@@ -45,6 +57,22 @@ export const RecipeFeedItem = ({ id }: ComponentProps) => {
 			handleDragStart={handleDragStartForCard(CardTypesMap.RECIPE, recipe)}
 			onAddToClipboard={onAddToClipboard}
 			shouldObscureFeed={shouldObscureFeed}
+			noPinboard={true}
+			bodyContent={
+				<>
+					<FeedItemTitle>{recipe.title}</FeedItemTitle>
```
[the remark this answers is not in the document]
fredex42: no; `data-testid` is discouraged anyway, it's only needed in cases where there is absolutely no other way to find the element in a test

fronts-client/src/components/inputs/HoverActionButtonWrapper.tsx
emdash-ie: You could have the prop default to `true` if not specified, no?

Divs-B:
When `always show date` is ON, few recipes are not showing dates under their titles.
One recipe where date is not mentioned it is not available also when clicked to view(we need to check this if recipe is not available then how we can show that recipe in the feed it shud be under `Recipe not found` i think, but can check in detail).
Here recipe: `Grilled nectarine & burrata salad with pickled onions, tomato and mint` which is not available when click to view.
image
And this one: this has no dates as well but this recipe is available when click to view.
image

fredex42:
@Divs-B regarding the various salads, I'll try that, what happens if you try to prioritise time over relevance?

PR #1732, opened 2024-11-19
Recipe filters

fredex42:
## What's changed?

Adds options to specify "celebration" and "suitable for diet" filters.

## Implementation notes

These are currently simple dropdown options, we should probably look at something more generic and sophisticated but this works at the moment

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/actions/__tests__/fixtures/Editions.fixture.ts
+++ b/fronts-client/src/actions/__tests__/fixtures/Editions.fixture.ts
@@ -764,6 +764,15 @@ export const initialState = {
 		loadingIds: [],
 		updatingIds: [],
 	},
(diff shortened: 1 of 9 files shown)


jonathonherbert (approved):
LGTM! An endpoint per keyword seems a lot, but perhaps we're worried about how this data grows, and it eliminates a class of problem.

It'd be lovely to display the doc counts alongside the option values, to give users an idea of how many recipes they're likely to see for a given option — guiding filtering behaviour and potentially saving some clicks. (Ideally, they'd be running totals that would update for combinations of options, but that's probably a larger piece of work.)

PR #1738, opened 2024-12-03
Indicate highlights container should use square crop

Georges-GNM:
## What's changed?

Images in the highlights container are expected to use an aspect ratio of 1:1 (i.e. square). We've previously switched to [5:4 aspect ratios](https://github.com/guardian/facia-tool/pull/1601) for the appropriate collections and here are making similar changes for the highlights container to use 1:1.

So similarly to that previous PR...

- the `InputImage` component should open the Grid view with a 1:1 square crop when users replace the trail image
- The `ThumbnailSmall` component should use a 1:1 image crop
- The trail image preview in the `ArticleMetaForm` should use a 1:1 image (the width/height and alignment have been slightly adjusted as otherwise the image ends up taking too much space)
- Various components (`Card`, `ArticleCard`, etc.) have been updated to pass on the image criteria.

The only function I've not been able to test is `Recrop image` which - for me - consistently returns "image not found" regardless of the container type, but I'm assuming that's environment related and the functionality should work as intended. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/card/article/ArticleBody.tsx
+++ b/fronts-client/src/components/card/article/ArticleBody.tsx
@@ -36,7 +36,7 @@ import ImageAndGraphWrapper from 'components/image/ImageAndGraphWrapper';
 import { getPaths } from 'util/paths';
 import { getMaybeDimensionsFromWidthAndHeight } from 'util/validateImageSrc';
 import { Criteria } from 'types/Grid';
-import { landscape5To4CardImageCriteria } from 'constants/image';
+import { landscape5To4CardImageCriteria, squareImageCriteria } from 'constants/image';
 
 const ThumbnailPlaceholder = styled(BasePlaceholder)`
 	flex-shrink: 0;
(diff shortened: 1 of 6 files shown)


fronts-client/src/components/card/article/ArticleBody.tsx
```diff
@@ -326,6 +333,7 @@ const articleBodyDefault = React.memo(
 									isDraggingImageOver={isDraggingImageOver}
 									isPortrait={thumbnailIsPortrait}
 									showLandscape54={showThumbnailInLandscape54}
+									showSquareThumbnail={showSquareThumbnail}
```
Georges-GNM, line 336: This leads to the component using similar styling implemented by the feast team. I reckon the thumbnail ends up being a bit small due to the values set [here](https://github.com/guardian/facia-tool/blob/2f0f6c1180415c68748964f082fb2c882355d58a/fronts-client/src/constants/theme.ts#L157) - if others agree I'm happy to change that.

Georges-GNM:
> Ignore the comments about the refactoring, I think that should be done separately! This looks great

Yeah, the thought crossed my mind at some point, but probably best to leave for another PR. Have added a ticket to the back of the backlog to keep as a reminder!

PR #1739, opened 2024-12-06
Add link to live front

Fweddi:
## What's changed?
Alongside the Preview button, show a link to the corresponding Live front: 
image

These buttons should collapse to just their icons if the front has a width smaller than 600px:
image


fronts-client/src/components/FrontsEdit/FrontSection/front-section.css
```diff
@@ -0,0 +1,17 @@
+/*
+	TODO: Move this logic into FrontSection.tsx when styledComponents is migrated to v6
+    styledComponents v6 will support container queries, which will allow us to remove this CSS file
+   */
+.front-header {
+	container-type: inline-size;
+}
```
Fweddi, line 7: For those new to container queries: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries

dblatcher (approved):

fronts-client/src/components/FrontsEdit/FrontSection/FrontSection.tsx
```diff
@@ -201,19 +215,36 @@ class FrontSection extends React.Component<
...
+											<LinkButtonText className="visible-based-on-front-header-width">
+												Preview
+											</LinkButtonText>
+										</FrontHeaderButton>
+									</Link>
+									<Link
+										href={`${
+											shouldUseCODELinks ? urls.liveUrlCODE : urls.liveUrlPROD
+										}${this.props.frontId}`}
+										target="live"
```
twrichards, line 238: I wonder if the `target` should be the name/path of the front? so that the tab can be re-used (as it is now) but can be different if you're working on multiple fronts at once

PR #1746, opened 2025-01-08
Enable limiting the number of stories in flexible general containers

Georges-GNM:
## What's changed?

Part of this [fairground ticket](https://trello.com/c/ACnDaVW8/697-flexible-general-tooling-set-story-limit-in-config-tool)

This allows editorial to set the number of stories in a flexible general container, by making use of an existing value, `maxItemsToDisplay`, nested in the `DisplayHints` section of a collection's config. 

The input has a placeholder of 9, which will also be a default/fallback number used throughout the codebases, to replicate existing behaviour. At the most there can be up to 20 stories, at the least there can be a single one.

Aside from adding the input, this PR uses the value to adjust the dividing line separation in the fronts tool (see screenshot); this is currently used by editorial as an indicator of which stories will appear - those below the line are effectively queued up, and so not intended to be displayed. 

As the amount of stories actually displayed on the platforms is determined in their respective codebases, follow up work is in process in frontend/mapi so they can also pick up on the desired value and ensure the right amount of stories gets displayed. 

## `maxItemsToDisplay` usage note

As mentioned this is an existing value, but it is currently only used for email fronts (which don't have access to flexible general containers), and has a similar intention to limit the number of visible stories. I would normally feel uneasy about piggybacking like this, but it feels like there's enough similarities while also having sufficient separation between the two use cases that this can be a working solution. 

This wasn't my first choice though, I intended to add a separate field (through this [facia scala client PR](https://github.com/guardian/facia-scala-client/pull/340/files)) but implementing this proved to be non trivial, hence this approach. 

## Fronts tool (v2) implementation note

The dividing lines are getting adjusted depending on the `maxItemsToDisplay` value, which for the most part seems to be working as intended, but the behaviour can be temperamental when manipulating splash cards (e.g. adding one or more, or moving a splash card to the list of standard stories, which seems to shift the dividing line up or down rather than keeping it at the desired number of stories). 

There's a separate ticket to limit the [number of splash cards to one story](https://trello.com/c/HAnWsQvN/353-flexible-general-restrict-splash-to-one-story), which feels best to tackle first and separately. I would expect the undesired behaviours to be resolved as part of that PR (or possibly a follow up). 

## Screenshots

Config tool:
image

Fronts tool - showing the dividing line appearing below the 5th standard story (with 1 splash)
image

## How to test

In the config tool, on a front, create a new flexible general container. You should see the field appear with the placeholder 9, and if you save the container, the tool should display a dividing line at 9 standard stories (you may need to refresh the tool once or twice). If you then edit the field in the config tool to a lower value, the dividing line should go up to the relevant spot. 

If these [frontend](https://github.com/guardian/frontend/pull/27708) and [DCR](https://github.com/guardian/dotcom-rendering/pull/13135) PRs are also deployed to code, you can check that the value is getting correctly used and updated by the web platform. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/StoriesVisibleController.scala
+++ b/app/controllers/StoriesVisibleController.scala
@@ -22,7 +22,7 @@ class StoriesVisibleController(
   def storiesVisible(containerType: String) =
     AccessAPIAuthAction(parse.json[StoriesVisibleRequest]) { implicit request =>
       val storiesVisible =
-        containerService.getStoriesVisible(containerType, request.body.stories)
+        containerService.getStoriesVisible(containerType, request.body.stories, collectionConfigJson = null)
 
       logger.info(
         s"got stories-visible=$storiesVisible for containerType=$containerType"
--- a/app/controllers/StoriesVisibleController.scala
+++ b/app/controllers/StoriesVisibleController.scala
@@ -25,7 +25,7 @@ class StoriesVisibleController(
         containerService.getStoriesVisible(
           containerType,
           request.body.stories,
-          collectionConfigJson = null
+          collectionConfigJson = None
         )
 
       logger.info(
--- a/app/controllers/StoriesVisibleController.scala
+++ b/app/controllers/StoriesVisibleController.scala
@@ -22,7 +22,11 @@ class StoriesVisibleController(
   def storiesVisible(containerType: String) =
     AccessAPIAuthAction(parse.json[StoriesVisibleRequest]) { implicit request =>
       val storiesVisible =
-        containerService.getStoriesVisible(containerType, request.body.stories, collectionConfigJson = null)
+        containerService.getStoriesVisible(
+          containerType,
+          request.body.stories,
+          collectionConfigJson = null
+        )
 
       logger.info(
         s"got stories-visible=$storiesVisible for containerType=$containerType"
--- a/app/services/CollectionService.scala
+++ b/app/services/CollectionService.scala
@@ -101,15 +101,15 @@ object CollectionService {
       CollectionService.getStoriesForCollectionStages(collectionId, collection)
     config.collections.get(collectionId).flatMap(_.`type`) match {
       case Some(cType) =>
-        val cConfigJson = config.collections(collectionId)
+        val cConfigJson = config.collections.get(collectionId)
         Some(
           StoriesVisibleByStage(
             containerService
-              .getStoriesVisible(cType, stages._1, Some(cConfigJson)),
+              .getStoriesVisible(cType, stages._1, cConfigJson),
             containerService.getStoriesVisible(
               cType,
               stages._2,
-              Some(cConfigJson)
+              cConfigJson
             )
           )
         )
--- a/app/services/CollectionService.scala
+++ b/app/services/CollectionService.scala
@@ -97,14 +97,15 @@ object CollectionService {
       config: ConfigJson,
       containerService: ContainerService
   ): Option[StoriesVisibleByStage] = {
+		val cConfigJson = config.collections.get(collectionId).get
     val stages =
       CollectionService.getStoriesForCollectionStages(collectionId, collection)
     config.collections.get(collectionId).flatMap(_.`type`) match {
       case Some(cType) =>
         Some(
           StoriesVisibleByStage(
-            containerService.getStoriesVisible(cType, stages._1),
-            containerService.getStoriesVisible(cType, stages._2)
+            containerService.getStoriesVisible(cType, stages._1, cConfigJson),
+            containerService.getStoriesVisible(cType, stages._2, cConfigJson)
           )
         )
       case None => None
--- a/app/services/CollectionService.scala
+++ b/app/services/CollectionService.scala
@@ -97,7 +97,7 @@ object CollectionService {
       config: ConfigJson,
       containerService: ContainerService
   ): Option[StoriesVisibleByStage] = {
-		val cConfigJson = config.collections.get(collectionId).get
+    val cConfigJson = config.collections.get(collectionId).get
     val stages =
       CollectionService.getStoriesForCollectionStages(collectionId, collection)
     config.collections.get(collectionId).flatMap(_.`type`) match {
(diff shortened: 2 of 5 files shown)


app/controllers/StoriesVisibleController.scala
```diff
@@ -22,7 +22,11 @@ class StoriesVisibleController(
   def storiesVisible(containerType: String) =
     AccessAPIAuthAction(parse.json[StoriesVisibleRequest]) { implicit request =>
       val storiesVisible =
-        containerService.getStoriesVisible(containerType, request.body.stories)
+        containerService.getStoriesVisible(
+          containerType,
+          request.body.stories,
+          collectionConfigJson = null
```
davidfurey, line 28: If this is a valid situations where `collectionConfigJson` is not provided can you make the field optional and pass `None` instead of `null`?

app/services/CollectionService.scala
```diff
@@ -97,14 +97,15 @@ object CollectionService {
       config: ConfigJson,
       containerService: ContainerService
   ): Option[StoriesVisibleByStage] = {
+    val cConfigJson = config.collections.get(collectionId).get
```
davidfurey, line 100: This will throw an exception of the collection is not found, please refactor to handle the failure case and avoid the `.get`

app/services/ContainerService.scala
```diff
@@ -48,7 +53,10 @@ class ContainerService(val containers: Containers) {
         )
 
       case Flexible(container) =>
-        val numberVisible = container.storiesVisible(stories)
+        val numberVisible = container.storiesVisible(
+          stories,
+          collectionConfigJson: CollectionConfigJson
```
davidfurey, line 58: Why is the type annotation required here?

app/slices/FlexibleContainer.scala
```diff
@@ -1,24 +1,39 @@
...
     val numOfSplash = splash.size min 1
     val numOfStandard = stories.size - numOfSplash
+    val defaultStandardStoryLimit = 8
 
-    numOfSplash + (numOfStandard min 8)
+    val standardStoryLimit = collectionConfigJson.displayHints
+      .flatMap(_.maxItemsToDisplay)
+      .getOrElse(defaultStandardStoryLimit)
+
+    numOfSplash + (numOfStandard min standardStoryLimit)
```
davidfurey, line 28: I think it would be easier to read as numOfSplash + min(numOfStandard, standardStoryLimit)

davidfurey (changes requested):
I think using `maxItemsToDisplay` is reasonable, although slightly unfortunate that it is nolonger the actual max but in fact the max standard. Still, I don't think it justifies a new field for now.

I had no memory of adding this field 9 years ago. I'm slightly surprised that no further display hints have been added.

app/services/CollectionService.scala
```diff
@@ -97,15 +97,20 @@ object CollectionService {
       config: ConfigJson,
       containerService: ContainerService
   ): Option[StoriesVisibleByStage] = {
-    val cConfigJson = config.collections.get(collectionId).get
     val stages =
       CollectionService.getStoriesForCollectionStages(collectionId, collection)
     config.collections.get(collectionId).flatMap(_.`type`) match {
       case Some(cType) =>
+        val cConfigJson = config.collections(collectionId)
```
davidfurey, line 104: This will through an exception if the collection is not present, and is exactly equivalent to the `config.collections.get(collectionId).get` you had before.

Instead, you could do `val cConfigJson = config.collections.get(collectionId)` and drop the `Some(...)` that you have wrapped `cConfigJson` in below.

app/slices/FlexibleContainer.scala
```diff
@@ -21,9 +21,13 @@ object FlexibleGeneral extends FlexibleContainer {
...
-    val standardStoryLimit = collectionConfigJson.displayHints
-      .flatMap(_.maxItemsToDisplay)
-      .getOrElse(defaultStandardStoryLimit)
+    val standardStoryLimit = collectionConfigJson match {
+      case Some(config) =>
+        config.displayHints
+          .flatMap(_.maxItemsToDisplay)
+          .getOrElse(defaultStandardStoryLimit)
+      case None => defaultStandardStoryLimit
+    }
```
davidfurey, line 30: You've used two different approaches to handling `Option`s here, pattern matching and `flatMap`. It is typically easy to read code that sticks to one approach.

```suggestion
    val standardStoryLimit = collectionConfigJson
      .flatMap(_.displayHints)
      .flatMap(_.maxItemsToDisplay)
      .getOrElse(defaultStandardStoryLimit)
```

Georges-GNM:
> I think using `maxItemsToDisplay` is reasonable, although slightly unfortunate that it is nolonger the actual max but in fact the max standard.

Yeah, I hesitated about that, and thought that since the number of splash stories was already limited, that really it was the number of standard stories we wanted to affect. 

However, from what I can tell the platforms, when they determine how many stories to show, don't easily have a way to access information about splash and standard stories, which we would need in order to get the right number of stories to come up (in frontend for example, the proposed implementation looks [like this](https://github.com/guardian/frontend/pull/27708/files#diff-3cd4553bc97f5296782fc33885ea69e6fc43a79c60ded9a0fb03595aa4ba80c6L334)), so in the end I think the best way is for the value to indeed reflect the actual max. Just in the process of revisiting this and testing it on code.

app/services/CollectionService.scala
Georges-GNM: I do prefer the look of that suggestion, but just to check my understanding - wouldn't this code only get executed after line 102, which should guarantee the collection is present at that point?

davidfurey: That is an excellent point. I missed that you already know that the collection exists. But much like me, the complier isn't clever enough to detect that your actually accessing it in a safe place way.

Looking at this again, you could refactor:
```
    for {
      cConfigJson <- config.collections.get(collectionId)
      cType <- cConfigJson.`type`
    } yield {
        StoriesVisibleByStage(
          containerService
            .getStoriesVisible(cType, stages._1, Some(cConfigJson)),
          containerService.getStoriesVisible(
            cType,
            stages._2,
            Some(cConfigJson)
          )
        )
    }
```

Georges-GNM: Thanks, that does look a bit more concise and seems to work!

davidfurey (approved):
Great work!

PR #1747, opened 2025-01-08
Upgrade to Panda v7 - support key rotation

davidfurey:
This upgrades Panda from v4 to v7, allowing us to use [key rotation](https://github.com/guardian/pan-domain-authentication/pull/150).

Follows the guidance from https://github.com/guardian/pan-domain-authentication/issues/160 and based on https://github.com/guardian/atom-workshop/pull/361

### DIFF
--- a/app/controllers/BaseFaciaController.scala
+++ b/app/controllers/BaseFaciaController.scala
@@ -1,10 +1,9 @@
 package controllers
 
 import java.util.Locale
-
 import com.gu.pandomainauth.action.{AuthActions, UserRequest}
 import com.gu.pandomainauth.model.AuthenticatedUser
-import com.gu.pandomainauth.{PanDomain, PanDomainAuthSettingsRefresher}
+import com.gu.pandomainauth.{PanDomain, PanDomainAuthSettingsRefresher, S3BucketLoader}
 import com.gu.permissions.{PermissionsConfig, PermissionsProvider}
 import conf.ApplicationConfiguration
 import logging.Logging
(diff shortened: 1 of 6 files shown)


davidfurey:
I've resolved two compiler warnings in a way which might change some behaviour that I'm unfamiliar with

@jonathonherbert - can you review https://github.com/guardian/facia-tool/pull/1747/commits/4a41ba54e98d0f4f8950afb2cfcaaeaea72f2b3e ?

@Divs-B - can you review https://github.com/guardian/facia-tool/pull/1747/commits/ba504988a7764ac3e3552252198d0b15d6631372

app/model/editions/client/EditionsClientCollection.scala
```diff
@@ -109,14 +109,14 @@ object EditionsSupportingClientCard {
...
     supportingCard.cardType match {
-      case Some(CardType.Recipe) =>
-        EditionsRecipe(supportingCard.id, supportingCard.frontPublicationDate)
       case Some(CardType.Chef) =>
         EditionsChef(
           supportingCard.id,
           supportingCard.frontPublicationDate,
           supportingCard.meta.map(_.toChefMetadata)
         )
+      case _ =>
```
Divs-B, line 118: Hi @davidfurey, thanks for asking, 
If I remember, we do need case matching for Recipe and Chef only, we don't want this to work for Feast-Collection(FC), because we don't have yet supporting feature of FC (present inside another FC) to drag into some another FC.
Can we have undefined or something similar if we really need to do case _?

davidfurey:
Tested on CODE - auth with an existing panda cookie works, and logging in from incognito window also works

jonathonherbert:
> @jonathonherbert - can you review [4a41ba5](https://github.com/guardian/facia-tool/commit/4a41ba54e98d0f4f8950afb2cfcaaeaea72f2b3e) ?

This is fine, nothing but recipes should be in FeastCollection meta, and `collect` is the idiomatic way to do this 👍

jonathonherbert (approved):

app/model/editions/client/EditionsClientCollection.scala
Divs-B: Thanks for discussion @davidfurey to discuss on ways to solve this. 
I will make change in later PR and attach here as reference.

Divs-B (approved):

PR #1753, opened 2025-01-23
Restrict splash to one story

Georges-GNM:
## What's changed?
Closes [this fronts+curation ticket](https://trello.com/c/HAnWsQvN/353-flexible-general-restrict-splash-to-one-story)

As per the ticket description, we want the tool to reflect how the frontend renders so that fronts editors will have clarity over what will appear to users. In this case specifically, as the splash can only contain up to a single story, we want to make sure this is limited in the tool as well. 

The approach I've taken is to update the `handleInsert` and `handleMove` functions called when a card is dragged to a collection/group, such that if we're not adding a card to a group with the name splash, nothing has changed and the pre-existing logic keeps running to insert as normal. This only concerns the new fairground containers, so it didn't feel necessary to ensure this is implemented retroactively. 

If we are adding to a splash, we check how many cards there already are in the group and which position we're inserting to:
- if there's no cards in the group already, we insert as normal
- if there's a card in the group already, and we're inserting to the bottom of it, the card we're inserting is instead moved to the top of the standard stories group
- if there's a card in the group already, and we're inserting to the top of it, the new story is inserted into the splash, and the previous story is moved to the top of the standard stories group.

Some new properties to relevant components have been added to enable checking these conditions (e.g. the ids of the groups in the collection, and the name and cards currently in the relevant group).

Manual testing would involve checking these four scenarios, for both dragging from the feed of stories and moving between groups. 

The jest tests have just been updated to pass, but not take into account these conditions yet - happy to try adding these variations though. 

The branch is currently up on code. 

## Video

https://github.com/user-attachments/assets/9c8e8c4e-c70a-49bd-b1dd-b94d29711aa7

### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/FrontsEdit/Collection.tsx
+++ b/fronts-client/src/components/FrontsEdit/Collection.tsx
@@ -114,7 +114,6 @@ interface ConnectedCollectionContextProps extends CollectionContextProps {
 	handleBlur: () => void;
 	lastDesktopArticle?: string;
 	lastMobileArticle?: string;
-	groupIds: string[];
 	updateCardMeta: (id: string, meta: CardMeta) => void;
 	addImageToCard: (uuid: string, imageData: ValidationResponse) => void;
 }
@@ -137,7 +136,6 @@ class CollectionContext extends React.Component<ConnectedCollectionContextProps>
 			removeSupportingCard,
 			lastDesktopArticle,
 			lastMobileArticle,
-			groupIds,
 			updateCardMeta,
 			addImageToCard,
 		} = this.props;
@@ -153,7 +151,7 @@ class CollectionContext extends React.Component<ConnectedCollectionContextProps>
 					canPublish={browsingStage !== 'live'}
 					browsingStage={browsingStage}
 				>
-					{(group, isUneditable, showGroupName) => (
+					{(group, isUneditable, groupIds, showGroupName) => (
 						<div key={group.uuid}>
 							<GroupDisplayComponent
 								key={group.uuid}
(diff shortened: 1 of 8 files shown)


fronts-client/src/components/FrontsEdit/Collection.tsx
```diff
@@ -114,6 +114,7 @@ interface ConnectedCollectionContextProps extends CollectionContextProps {
 	handleBlur: () => void;
 	lastDesktopArticle?: string;
 	lastMobileArticle?: string;
+	groupsIds: string[];
```
abeddow91, line 117: Nit : sorry, could this be renamed to groupIds rather than groupsIds?

Georges-GNM: Yep, that's better - done!

fronts-client/src/components/FrontsEdit/Collection.tsx
```diff
@@ -258,7 +264,10 @@ const createMapStateToProps = () => {
 			collectionSet: props.browsingStage,
 		});
 
+		const groupIds = state.collections.data[props.id]?.draft || [];
```
abeddow91, line 267: I think this can be refactored such that we pass the group ids though in the collection component, rather than needing to add it to the store here. 

we have access to groups here https://github.com/guardian/facia-tool/blob/main/fronts-client/src/components/FrontsEdit/CollectionComponents/Collection.tsx#L207

My thought is we can map over this to retrieve a list of arrays, and then drill them through to the group here https://github.com/guardian/facia-tool/blob/5d6c14b39527982aac3147ab5b978124023cc4ea/fronts-client/src/components/FrontsEdit/CollectionComponents/Collection.tsx#L341

Georges-GNM: Thanks, done - looks like it still works 🎉

abeddow91 (commented):
It looks like the functionailty is good here but I find the nesting if else statement quite difficult to read. I think early returns would help improve the readibility here. Let me know if you'd like to work together to refactor this.

abeddow91 (approved):
Great work!

PR #1756, opened 2025-02-05
Adjust visual hierarchy of containers depending on primary/secondary

Georges-GNM:
## What's changed?

Closes this f+c [ticket](https://trello.com/c/0rpuA4HV/932-improve-the-visual-hierarchy-of-primary-secondary-containers-in-the-fronts-tooling) to Improve the visual hierarchy of primary & secondary containers in the fronts tooling. 

We've shown this to a couple of stakeholders in editorial who've said they're happy with the look. 

image

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/FrontsEdit/CollectionOverview.tsx
+++ b/fronts-client/src/components/FrontsEdit/CollectionOverview.tsx
@@ -17,6 +17,7 @@ import {
 } from 'selectors/shared';
 import EditModeVisibility from 'components/util/EditModeVisibility';
 import { createSelectCollectionIdsWithOpenForms } from 'bundles/frontsUI';
+import { css } from 'styled-components';
 
 interface FrontCollectionOverviewContainerProps {
 	frontId: string;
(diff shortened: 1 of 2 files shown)


fronts-client/src/components/FrontsEdit/CollectionOverview.tsx
```diff
@@ -127,9 +145,19 @@ const CollectionOverview = ({
 				openCollection(collection.id);
 			}}
 			isSelected={isSelected}
+			isSecondaryContainer={
+				collection.metadata?.some((tag) => tag.type === 'Secondary') ?? false
```
abeddow91, line 149: We could use a double negation here (and the other two points we run this expression) to force a true / false outcome rather than having a fallback of false. This is up to you though, a personal preference judgement call


```suggestion
				!!collection.metadata?.some((tag) => tag.type === 'Secondary')
```

abeddow91 (approved):
Nice solution. I like that you are coercing the secondary value so we don't have to handle an optional in the styles. 

One syntax suggestion but its non-blocking and personal preference.

PR #1762, opened 2025-02-21
Handle the move of suppressImages to displayHints

Fweddi:
## What's changed?
In https://github.com/guardian/facia-scala-client/pull/342 we moved the `suppressImages` field to DisplayHints in order to free up the Case Class limit on the CollectionConfig. This PR handles that change in the Fronts Tool.

## Testing
In the Fronts config, using a Beta container, check the `suppressImages` checkbox. Refresh the config page. The field should persist. 

image

Navigate to the Fronts Tool proper and open a front with the relevant container. You should see `Images suppressed` appear at the top of the container:
image

If you are deployed to CODE you can check the CODE front using the `See live` button. The images for the container should be hidden.

Uncheck the checkbox in the config; refresh, the `Images suppressed` label should disappear in the Fronts tool and the relevant container should now show images on the CODE front.


public/src/js/models/config/collection.js
```diff
@@ -46,12 +46,12 @@ export default class ConfigCollection extends DropTarget {
...
                 'userVisibility',
-                'targetedTerritory',
-                'suppressImages'
+                'targetedTerritory'
             ]),
             {
                 displayHints: asObservableProps([
-                    'maxItemsToDisplay'
+                    'maxItemsToDisplay',
+					'suppressImages'
```
davidfurey, line 54:
```suggestion
                    'maxItemsToDisplay',
                    'suppressImages'
```

davidfurey (approved):
lgtm

PR #1763, opened 2025-02-21
Add editable maxItems field in Config tool

Fweddi:
## What's changed?
Add an editable maxItems field per 'group' (e.g. splash, standard) in a given container.

This allows the new `flexible/general` container to have different card groups with variable card limits.

Intended to replace the use of the `maxItemsToDisplay` field. Note this PR does not enforce any card limits in the Fronts Tool, this will be carried out in a future PR.

Note you will need to 'reset' the container to see the `groupsConfig` defaults pull through. So if you are looking at a `flexible/general` container in the config, you will need to change the container type to something else and then back to `flexible/general`.

| Before | After | 
| --- | --- |
|  |  |

### DIFF
--- a/app/model/frontsapi.scala
+++ b/app/model/frontsapi.scala
@@ -421,8 +421,8 @@ trait UpdateActionsTrait extends Logging {
       collectionId: String,
       collectionJson: CollectionJson
   ): CollectionJson = {
-    configAgent.getConfig(collectionId).flatMap(_.groups) match {
-      case Some(groups) if groups.groups.nonEmpty => collectionJson
+    configAgent.getConfig(collectionId).flatMap(_.groupsConfig) match {
+      case Some(groupsConfig) if groupsConfig.groups.nonEmpty => collectionJson
       case _ =>
         collectionJson.copy(
           live = collectionJson.live.map(removeGroupsFromTrail),
--- a/fronts-client/src/fixtures/shared.ts
+++ b/fronts-client/src/fixtures/shared.ts
@@ -1,3 +1,5 @@
+import { CollectionConfig } from '../types/FaciaApi';
+
 const capiArticle = {
 	id: 'world/live/2018/sep/13/florence-hurricane-latest-live-news-updates-weather-path-storm-surge-north-carolina',
 	type: 'liveblog',
@@ -591,11 +593,25 @@ const normalisedCollectonWithPreviously = {
 	excludedRegions: [],
 };
 
-const collectionConfig = {
+const collectionConfig: CollectionConfig = {
 	id: 'exampleCollection',
 	displayName: 'Example Collection',
 	type: 'any',
 	groups: ['large', 'medium', 'small'],
+	groupsConfig: [
+		{
+			name: 'large',
+			maxItems: 10,
+		},
+		{
+			name: 'medium',
+			maxItems: 10,
+		},
+		{
+			name: 'small',
+			maxItems: 10,
+		},
+	],
 };
 
 const collectionWithSupportingArticles = {
--- a/public/src/css/style.css
+++ b/public/src/css/style.css
@@ -1930,6 +1930,15 @@ ol.search_list > li::before {
     height: 50px;
 }
 
+.capitalize-first-letter {
+    /*needed for :first-letter to work*/
+    display: inline-block;
+}
+
+.capitalize-first-letter:first-letter {
+    text-transform: capitalize;
+}
+
 .cnf-form .action-buttons {
     height: 40px;
 }
--- a/public/src/js/models/config/persistence.js
+++ b/public/src/js/models/config/persistence.js
@@ -65,7 +65,8 @@ function flattenModel (model) {
     return _.reduce(model, function (accumulator, value, key) {
         var x = flattenValue(value);
 
-        if (x) {
+        // We want to drop falsy values, but not 0
+        if (x || x === 0) {
             accumulator[key] = x;
         }
 
--- a/public/src/js/models/config/persistence.js
+++ b/public/src/js/models/config/persistence.js
@@ -53,6 +53,9 @@ function flattenModel (model) {
 
     function flattenValue(value) {
         if (_.isFunction(value)) {
+            if (_.isArray(value())) {
+              return value().map(flattenValue);
+            }
             return value();
         } else if (_.isObject(value)) {
             var flattened = flattenModel(value);
--- a/public/src/js/utils/observable-numeric.js
+++ b/public/src/js/utils/observable-numeric.js
@@ -8,7 +8,7 @@ export default function(initialValue) {
         },
         write(newValue) {
             var parsedValue = parseFloat(newValue);
-            actual(isNaN(parsedValue) ? newValue : parsedValue);
+            actual(isNaN(parsedValue) ? newValue : Math.abs(parsedValue));
         }
     });
 }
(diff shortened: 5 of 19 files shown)


app/model/frontsapi.scala
```diff
@@ -421,8 +421,8 @@ trait UpdateActionsTrait extends Logging {
       collectionId: String,
```
Fweddi, line 421: I'm confused about the relation between the config `groups` field and the trail metadata `group` field.

fronts-client/src/fixtures/shared.ts
```diff
@@ -591,11 +593,22 @@ const normalisedCollectonWithPreviously = {
 	excludedRegions: [],
 };
 
-const collectionConfig = {
+const collectionConfig: CollectionConfig = {
 	id: 'exampleCollection',
 	displayName: 'Example Collection',
 	type: 'any',
-	groups: ['large', 'medium', 'small'],
```
abeddow91, line 598: Existing containers have a concept of groups. Is the plan to move to groupsConfig for all containers rather than just for "beta" containers?

abeddow91: ahh ignore the above, i see that we're supporting both here https://github.com/guardian/facia-scala-client/pull/345/files 👍

Fweddi: am keeping groups for now as we'll need to support clients who haven't upgraded to the latest model 👍

public/src/js/models/config/persistence.js
```diff
@@ -65,7 +68,8 @@ function flattenModel (model) {
     return _.reduce(model, function (accumulator, value, key) {
         var x = flattenValue(value);
 
-        if (x) {
+        // We want to drop falsy values, but not 0
```
Fweddi, line 71: Cf. https://github.com/guardian/facia-tool/pull/682
> This pr also cleans all falsy values and empty arrays from the meta object to correspond with what is saved in the meta by v1 and avoid any obscure future bugs.

There are a couple of places in the tests where we are explicitly testing falsy values are filtered out, so this seemed the minimum change to allow us to support 0 as a value.

0 as a value is useful as it lets config editors know that a group (e.g. Very Big) has explicitly been set to 0 rather than it is unset and waiting to be filled in.

public/src/css/style.css
```diff
@@ -1930,6 +1930,15 @@ ol.search_list > li::before {
     height: 50px;
 }
 
+.capitalize-first-letter {
+    /*needed for :first-letter to work*/
+    display: inline-block;
+}
+
+.capitalize-first-letter:first-letter {
```
Fweddi, line 1938: This means we can write 'very big' as 'Very big' (and not 'Very Big' which `text-transform: capitalize` would do if left to its own devices).

Fweddi: (Stakeholder request)

Fweddi:
It would be good to have future PRs to:
(1) remove all references to groups in place of groupsConfig
(2) allow some groupConfig maxItems to not be editable - specifically, splash, which should only ever be one card*
(3) fix `suppressImages` (a boolean) being instantiated with `observableNumeric` - it works, because `observableNumeric` is slightly odd, but it shouldn't be instantiated this way

*At least, at this current moment...

public/src/js/utils/observable-numeric.js
```diff
@@ -8,7 +8,7 @@ export default function(initialValue) {
         },
         write(newValue) {
             var parsedValue = parseFloat(newValue);
-            actual(isNaN(parsedValue) ? newValue : parsedValue);
+            actual(isNaN(parsedValue) ? newValue : Math.abs(parsedValue));
```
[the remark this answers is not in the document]
Fweddi: It prevents us from modelling negative numbers. This code is invoked in the few cases we expect numbers, and we have already checked at this stage that the variable definitely _is_ a number. Happy to jettison it or move it elsewhere though.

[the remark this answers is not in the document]
davidfurey: We plan to EOL this app so it is not really a problem for this API to be confusing, let's leave as is.
(with apologies to me or other future developer who gets bitten by this in 5 years time when we are still using this code)

public/src/js/widgets/columns/fronts-config.html
```diff
@@ -259,10 +259,12 @@
...
             <!-- /ko -->
 
-			<!-- ko if: meta.type() === "flexible/general"-->
-			<label for="flexGenStoryLimit">Max stories to display</label>
-			<input id="flexGenStoryLimit" type="number" data-bind="value: meta.displayHints.maxItemsToDisplay" attr: { max: 20, min: 1 } max="20" min="1" placeholder="9">
-			<!-- /ko -->
+            <!-- ko if: meta.type() === "flexible/general"-->
+              <div data-bind="foreach: meta.groupsConfig">
+                <label for="flexGenSplashStoryLimit"><span data-bind="text: name" class="capitalize-first-letter"></span> stories</label>
+                <input id="flexGenSplashStoryLimit" type="number" data-bind="value: maxItems" attr: { max: 20, min: 0 } max="20" min="0">
```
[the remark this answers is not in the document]
Fweddi: Actioned

Fweddi:
> > It would be good to have future PRs to:
> > 
> > 1. remove all references to groups in place of groupsConfig
> > 2. allow some groupConfig maxItems to not be editable - specifically, splash, which should only ever be one card*.
> > 3. fix `suppressImages` (a boolean) being instantiated with `observableNumeric` - it works, because `observableNumeric` is slightly odd, but it shouldn't be instantiated this way
> > 
> > *At least, at this current moment...
> 
> When we do `1.` we need to make sure that groupsConfig is populated without someone having to go through each container and changing the type backwards/forwards.
> 
> For `2.`, if rendering platforms can handle multi splash cards, JH has no concerns about the tool allowing it. Even if the platforms cannot handle it, he is still fairly relaxed about this.

Thank you for clarifying 2 as I'd expected us to want to keep the hard limit introduced in https://github.com/guardian/facia-tool/pull/1753.

public/src/js/utils/observable-numeric.js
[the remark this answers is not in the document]
Fweddi: Perhaps we can address this when we fix the `suppressImages` being treated as an `observableNumeric`.

davidfurey:
> Thank you for clarifying 2 as I'd expected us to want to keep the hard limit introduced in #1753.

It is fine for the hard-limit to be controlled by configuration, rather than hardcoded.

public/src/js/widgets/columns/fronts-config.html
```diff
@@ -259,10 +259,12 @@
...
             <!-- /ko -->
 
-			<!-- ko if: meta.type() === "flexible/general"-->
-			<label for="flexGenStoryLimit">Max stories to display</label>
-			<input id="flexGenStoryLimit" type="number" data-bind="value: meta.displayHints.maxItemsToDisplay" attr: { max: 20, min: 1 } max="20" min="1" placeholder="9">
-			<!-- /ko -->
+            <!-- ko if: meta.type() === "flexible/general"-->
+              <div data-bind="foreach: meta.groupsConfig">
+                <label data-bind="attr: { for: name }"><span data-bind="text: name" class="capitalize-first-letter"></span> stories</label>
+                <input type="number" data-bind="value: maxItems, attr: { id: name, max: 20, min: 0 }" max="20" min="0">
```
davidfurey, line 265: I don't think `id`s can have whitespace. Perhaps this will work (untested):
`attr: { for: 'maxItems-' + $index }`

PR #1766, opened 2025-02-28
Restriction on one card in splash group should not apply to sublinks

davidfurey:
## What's changed?

https://github.com/guardian/facia-tool/pull/1753 restricted splash groups to only contain one story, but it also inadvertently limited splash cards to only have one sublink.

This PR adds an additional guard so that the singular limit on splash groups only applies if second story is being added to the group rather than to the card (as a sublink).

### DIFF
--- a/fronts-client/src/components/FrontsEdit/FrontContent.tsx
+++ b/fronts-client/src/components/FrontsEdit/FrontContent.tsx
@@ -259,7 +259,11 @@ class FrontContent extends React.Component<FrontProps, FrontState> {
 
 		// if we are inserting an article into any group that is not the splash, then we just insert
 		// we also just insert if we're in the splash and there's no other article already in the splash
-		if (to.groupName !== 'splash' || numberOfArticlesAlreadyInGroup === 0) {
+		if (
+			to.type !== 'group' ||
+			to.groupName !== 'splash' ||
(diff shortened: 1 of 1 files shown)


Georges-GNM (commented):
This does fix the sublink issue (and we may want to add the same guard in the handle move function above when moving cards around)

But it looks to me like the logic to restrict the number of cards in the splash is no longer working? As in inserting a story above or below the existing one allows it to come in to the splash group rather than moving a story to the standard one. 

image

Just having a think as to what that could be due to. Also thinking that logic may be getting revisited as part of the groups config work next week, so maybe making sure the right number of cards is limited in the group can be left to one of the PRs raised for that work?

davidfurey:
> This does fix the sublink issue (and we may want to add the same guard in the handle move function above when moving cards around)
> 
> But it looks to me like the logic to restrict the number of cards in the splash is no longer working? As in inserting a story above or below the existing one allows it to come in to the splash group rather than moving a story to the standard one.

This was a bug. I was checking for the value 'collection' but should have been checking for 'group'

Should work now.

Georges-GNM (approved):
Tested locally and fixes the issue (which I probably should have caught when writing the original PR, but don't think I tested thoroughly with sublinks) 

Think the failing test is just flaky, so this looks good to me 👍

PR #1767, opened 2025-03-04
Set default boost level for groups in flexible general containers when dragging from feed

Georges-GNM:
## What's changed?

Part of this [F+C ticket](https://trello.com/c/a16rDa8J/1012-add-group-limits-to-fronts-tool) implementing the new group config limits in the v2 client. 

This adds a defaultGroupBoostLevel applied to a card when it is dropped in a flexible general container group. A good portion of the changes is just passing props around so we can check the name of the group being added to (via the `to` prop) and access the collectionId to check the collectionType. 

This doesn't seem to work when moving a story from the clipboard or between groups, but those points will be addressed in a subsequent PR. 

## Testing

Check out the branch, set up a flexible general container which should then have the 4 groups, splash, very big, big and standard. When dragging from the feed (not using the clipboard or moving from one group to another), you should observe the following behaviour:
- going into a splash or standard group, there should be no boostlevel applied
- going into a very big, it should be megaboost
- going into a big it should be boost. 

## Screenshot

Shows the settings were applied correctly (just on dragging and dropping) and are coming through on DCR. 

image

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/util/card.ts
+++ b/fronts-client/src/util/card.ts
@@ -59,7 +59,7 @@ const createCard = (
 	isEditionsApp: boolean,
 	{
 		cardType = 'article',
-		boostLevel = 'megaboost',
+		boostLevel = 'default',
 		imageHide = false,
 		imageReplace = false,
 		imageCutoutReplace = false,
--- a/fronts-client/src/util/card.ts
+++ b/fronts-client/src/util/card.ts
@@ -173,9 +173,15 @@ const getCardEntitiesFromDrop = async (
 	state: State,
 	to: PosSpec,
 ): Promise<TArticleEntities> => {
-	const collectionType = to.collectionId && selectCollectionType(state, to.collectionId);
+	const collectionType =
+		to.collectionId && selectCollectionType(state, to.collectionId);
 	if (drop.type === 'CAPI') {
-		return getArticleEntitiesFromFeedDrop(drop.data, isEdition, to, collectionType);
+		return getArticleEntitiesFromFeedDrop(
+			drop.data,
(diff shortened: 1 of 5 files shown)


fronts-client/src/util/card.ts
```diff
@@ -55,6 +59,7 @@ const createCard = (
 	isEditionsApp: boolean,
 	{
 		cardType = 'article',
+		boostLevel = 'megaboost',
```
Georges-GNM, line 62: don't think this is needed?

fronts-client/src/util/card.ts
```diff
@@ -318,10 +332,30 @@ const getFeastCollectionFromFeedDrop = (
 const getArticleEntitiesFromFeedDrop = (
 	capiArticle: CapiArticle,
 	isEdition: boolean,
+	to?: PosSpec,
+	collectionType?: string,
```
abeddow91, line 336: non-blocking: are we able to tighten this type to a list of collectionTypes (FLEXIBLE_GENERAL_NAME, FLEXIBLE_SPECIAL_NAME) etc rather than string?

Georges-GNM: Hm, yeah, good point. I think that would probably mean introducing all the collection types though and ideally updating the other parts of the application where it gets used?
Screenshot 2025-03-05 at 15 28 57

Makes me think it's best left to another PR.

abeddow91 (approved):
Great work! A very elegant solution to setting defaulted boost levels

PR #1770, opened 2025-03-07
Add boost toggles for groups

abeddow91:
## What's changed?
Refactors the boost toggles available on flexible container cards. This allows us to provide preset toggle groups to boost groups. The reason for this is to simplify the boosting process for editors. We override the labels in some groups ( eg very big only has the option of megaboost but this is labelled default) so they make more sense in the context of the collection.

The groupings are as follows:

  ### Flexible General (4 Groups):
  - **Group 0 (Standard)**: Default, Boost, MegaBoost
  - **Group 1 (Big)**: Boost, MegaBoost
  - **Group 2 (Very Big)**: MegaBoost
  - **Group 3 (Splash)**: Default, Boost, MegaBoost, GigaBoost
 
  ### Flexible Special (2 Groups):
  - **Group 0 (Standard)**: Default, Boost, MegaBoost
  - **Group 1 (Splash)**: Default, Boost, MegaBoost, GigaBoost
 
### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client


fronts-client/src/components/form/ArticleMetaForm.tsx
```diff
@@ -530,15 +538,6 @@ class FormComponent extends React.Component<Props, FormComponentState> {
 			);
 		};
 
-		const allowGigaBoost = () =>
```
abeddow91, line 533: this is no longer required as the renderBoostToggles function manages show / hiding of toggles

fronts-client/src/components/form/ArticleMetaForm.tsx
```diff
@@ -172,26 +169,26 @@ const CaptionControls = styled.div`
...
+	margin-bottom: 60px;
+	color: ${(props: { invalid: boolean }) =>
+		props.invalid ? error.primary : 'default'};
+	border-color: ${(props: { invalid: boolean }) =>
+		props.invalid ? error.primary : 'default'};
+	:focus {
+		border-color: ${(props: { invalid: boolean }) =>
+			props.invalid ? error.primary : 'default'};
+	}
 `;
```
abeddow91, line 192: drive-by formatting fixes

fronts-client/src/components/form/BoostToggles.tsx
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * */
+const CollectionToggles: Record<string, Record<number, Toggle[]>> = {
+	'flexible/general': {
+		0: [TOGGLES.default, TOGGLES.boost, TOGGLES.megaboost],
+		1: [
+			{ ...TOGGLES.boost, label: 'default' },
+			{ ...TOGGLES.megaboost, label: 'boost' },
+		],
+		2: [{ ...TOGGLES.megaboost, label: 'default' }],
```
abeddow91, line 67: Given there is only one option here, we might want to change this to an empty array so that nothing renders rather than an unchangeable "default" radio button. 

We will need feedback from editorial stakeholders before making this change so for now i have left this toggle in and will fix forward if required.

abeddow91: This has been discussed and editorial would like to keep the default toggle visible.

fronts-client/src/components/form/BoostToggles.tsx
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * ### Flexible General (4 Groups):
+ * - **Group 0 (Standard)**: Default, Boost, MegaBoost
+ * - **Group 1 (Big)**: Boost, MegaBoost
+ * - **Group 2 (Very Big)**: MegaBoost
+ * - **Group 3 (Splash)**: Default, Boost, MegaBoost, GigaBoost
+ *
+ * ### Flexible Special (2 Groups):
+ * - **Group 0 (Standard)**: Default, Boost, MegaBoost
+ * - **Group 1 (Splash)**: Default, Boost, MegaBoost, GigaBoost
```
Georges-GNM, line 57: Discussed in person, but just a quick note here that the grouping in flexible special is... special, and we need to allow gigaboost for standard stories too (so splashes have the correct options). 

In the future we can revisit the grouping in flexible special so we have something a bit more explicit and similar to the flexible general groups though!

Georges-GNM (approved):
Tested locally, looks good to me 👏

PR #1771, opened 2025-03-09
Update logic to set boost level when moving cards

Georges-GNM:
## What's changed?

When we only had the splash and standard groups in flexible general containers, we had some logic that triggered when moving a card from the splash to the standard group and set the boostlevel to megaboost if it had previously been set to gigaboost. 

That logic stays in place and we've now introduced the very big and big groups, which have similar requirements - when moving a card to the very big, they need to be set to megaboost; similarly, cards going to the big group need to be set to boosted.

Standard and splash groups both allow megaboost and boost levels, so we don't make any changes when moving a card with those properties to either group.

## Recording

https://github.com/user-attachments/assets/3ea3b169-8f58-4322-a22b-d164bf4489cb

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/actions/Cards.ts
+++ b/fronts-client/src/actions/Cards.ts
@@ -225,11 +225,12 @@ const mayLowerCardBoostLevelForDestinationGroup = (
 		const { collection } = selectGroupCollection(state, groupId);
 		const group = selectGroups(state)[groupId];
 		if (collection?.type === FLEXIBLE_GENERAL_NAME) {
-			if (!group || group.id === null) {
-				return;
-			}
 			// if we move a gigaboosted card to a standard group, we set megaboost
-			if (parseInt(group.id) === 0 && card.meta.boostLevel === 'gigaboost') {
+			if (
+				group &&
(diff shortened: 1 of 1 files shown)


fronts-client/src/actions/Cards.ts
```diff
@@ -222,6 +225,7 @@ const mayLowerCardBoostLevelForDestinationGroup = (
 		const { collection } = selectGroupCollection(state, groupId);
 		const group = selectGroups(state)[groupId];
 		if (collection?.type === FLEXIBLE_GENERAL_NAME) {
+			// if we move a gigaboosted card to a standard group, we set megaboost
 			if (
```
abeddow91, line 229: We could lift up an early return here to say 
`if (!group || group.id === null)` {
return
}`

this would allow us to remove this repetition form lines 229, 244, and 254

abeddow91 (commented):
looks good! could we make a slight tweak to add an early return to make the reasoning logic easier to read please?

PR #1772, opened 2025-03-10
Adjust boost levels when moving between groups

abeddow91:
## What's changed?
Refactors boost levels when a card is moved between groups. It implements the following rules:

- If the card is moved to a splash (group 3) group, we always set the boost level to default
- If the card's boost level is lower than the minimum required for the group, set it to the minimum group boost level
- If the card already has the minimum group boost level, we don't want to go any lower, so we use the existing boost level
- Otherwise, we decrease the boost level by 1. 
   This results in the follow flow of levels:
   **Gigaboost -> MegaBoost -> Boost -> Default**
 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/actions/Cards.ts
+++ b/fronts-client/src/actions/Cards.ts
@@ -251,7 +251,7 @@ const getBoostLevel = (groupId: number, boostIndex: number) => {
 	return boostLevels[boostIndex - 1];
 };
 
-const mayLowerCardBoostLevelForDestinationGroup = (
+const mayAdjustCardBoostLevelForDestinationGroup = (
 	state: State,
 	to: PosSpec,
 	card: Card,
@@ -261,7 +261,6 @@ const mayLowerCardBoostLevelForDestinationGroup = (
 		const groupId = to.id;
 		const { collection } = selectGroupCollection(state, groupId);
 		const group = selectGroups(state)[groupId];
-		console.log(card.meta.group);
 
 		if (collection?.type === FLEXIBLE_GENERAL_NAME) {
 			if (!group) {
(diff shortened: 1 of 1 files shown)


fronts-client/src/actions/Cards.ts
```diff
@@ -214,7 +214,44 @@ const updateCardMetaWithPersist = (persistTo: PersistTo) =>
...
+	return 'default';
+};
+
+/**
+ * If the card is moved to a splash (group 3) group, we always set the boost level to default
+ * If the card already has the minimum boost level for the group it is moving to,
+ * we don't want to go any lower.
+ * Otherwise, we decrease the boost level by 1.
+ */
+const getBoostLevel = (groupId: number, boostIndex: number) => {
```
Georges-GNM, line 233: Maybe saying it's the current index is a tad clearer? 

```suggestion
const getBoostLevel = (groupId: number, currentBoostIndex: number) => {
```

fronts-client/src/actions/Cards.ts
```diff
@@ -224,44 +261,25 @@ const mayLowerCardBoostLevelForDestinationGroup = (
...
-					card.uuid,
-					{
-						boostLevel: 'boost',
-					},
-					{ merge: true },
-				);
-			}
+			const groupId = parseInt(group.id ?? '0');
+
+			const currentBoostLevel = boostLevels.indexOf(
```
Georges-GNM, line 271: Similarly, maybe:

```suggestion
			const currentBoostLevelIndex = boostLevels.indexOf(
```

Georges-GNM (approved):
Nitpicky non blocking suggestion about a value name, but otherwise really neat solution 👏 

Tested on code and all seemed to be working as intended 🎉

PR #1780, opened 2025-03-19
Add radio buttons to select media

Fweddi:
## What's changed?

Replaces various toggle switches for media selection (Show video, Slideshow) with a group of radio buttons.

The new `Video` radio button will have more features in future. Currently it can only be selected if the main media of a given article is a video.

| Before | After | 
| --- | --- |
|  |  | 

### Testing
Do the toggle switches work? Are their changes reflected in the pressed Front?

There is an existing issue where Video articles cannot become Slideshows when rendered as Splash cards in Flexible General containers. This PR does not address this bug.

I have also seen a new intermittent issue where Slideshows are being rendered with empty fields, and so the slideshow cannot be edited. I am not clear of the cause, and I have not seen it for a while. It may have been caused by a broken front rather than any changes from this PR. Let me know if you see it - we wouldn't want to release this PR if it causes slideshows to intermittently break.

### Demo
<video src="https://github.com/user-attachments/assets/7a1116a7-3ecd-419f-b706-2f64da138ba0" width="400px"/>

### DIFF
--- a/fronts-client/src/components/Row.tsx
+++ b/fronts-client/src/components/Row.tsx
@@ -1,11 +1,16 @@
 import { styled } from 'constants/theme';
 
-const Row = styled.div<{ flexDirection?: string; gutter?: number, rowGap?: number }>`
+const Row = styled.div<{
+	flexDirection?: string;
+	gutter?: number;
+	rowGap?: number;
+}>`
 	display: flex;
 	margin: 0 ${({ gutter = 10 }: { gutter?: number }) => `${-(gutter / 2)}px`};
 	flex-wrap: wrap;
 	// row-gap used when wrapped
-	row-gap: ${({ rowGap }: { rowGap?: number }) => rowGap ? `${rowGap}px` : '0'};
+	row-gap: ${({ rowGap }: { rowGap?: number }) =>
+		rowGap ? `${rowGap}px` : '0'};
 	flex-direction: ${({ flexDirection }) =>
 		flexDirection ? flexDirection : ''};
 `;
--- a/fronts-client/src/components/Row.tsx
+++ b/fronts-client/src/components/Row.tsx
@@ -1,8 +1,11 @@
 import { styled } from 'constants/theme';
 
-const Row = styled.div<{ flexDirection?: string; gutter?: number }>`
+const Row = styled.div<{ flexDirection?: string; gutter?: number, rowGap?: number }>`
 	display: flex;
 	margin: 0 ${({ gutter = 10 }: { gutter?: number }) => `${-(gutter / 2)}px`};
+	flex-wrap: wrap;
+	// row-gap used when wrapped
+	row-gap: ${({ rowGap }: { rowGap?: number }) => rowGap ? `${rowGap}px` : '0'};
 	flex-direction: ${({ flexDirection }) =>
 		flexDirection ? flexDirection : ''};
 `;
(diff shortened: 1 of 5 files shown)


fronts-client/src/components/Row.tsx
```diff
@@ -1,8 +1,16 @@
...
+	flexDirection?: string;
+	gutter?: number;
+	rowGap?: number;
+}>`
 	display: flex;
 	margin: 0 ${({ gutter = 10 }: { gutter?: number }) => `${-(gutter / 2)}px`};
+	flex-wrap: wrap;
+	// row-gap used when wrapped
+	row-gap: ${({ rowGap }: { rowGap?: number }) =>
+		rowGap ? `${rowGap}px` : '0'};
```
Fweddi, line 13: This adds a more semantic vertical gap between the Trail Image and Select Media sections when the row is wrapped (Select Media beneath Trail Image).

fronts-client/src/components/form/ArticleMetaForm.tsx
```diff
@@ -1080,6 +1128,7 @@ const createMapStateToProps = () => {
 				article && selectFormFields(state, article.uuid, isSupporting),
 			kickerOptions: article ? selectArticleTag(state, cardId) : defaultObject,
 			imageSlideshowReplace: valueSelector(state, 'imageSlideshowReplace'),
+			showMainVideo: valueSelector(state, 'showMainVideo'),
```
Fweddi, line 1131: We need to access the Redux form value to bind to the checked status of the Select Video radio button. We already have the imageSlideshowReplace selector in the line above.

fronts-client/src/components/inputs/InputRadio.tsx
```diff
@@ -95,18 +122,28 @@ export default ({
...
 	<>
 		<InputContainer data-testid={dataTestId}>
-			<RadioButtonContainer>
+			<RadioButtonContainer usesBlockStyling={usesBlockStyling}>
 				<Switch>
-					<RadioButton type="radio" {...inputRest} {...rest} id={id} />
+					<RadioButton
+						type="radio"
+						{...inputRest}
+						checked={checked}
```
Fweddi, line 137: `checked` needs to be supplied directly, the spread operator can drop this value. The value is needed to manually drive the radio button, which doesn't play well with Redux forms (https://github.com/redux-form/redux-form/issues/124).

fronts-client/src/components/inputs/InputRadio.tsx
```diff
@@ -9,17 +9,33 @@ import { theme } from 'constants/theme';
...
+	${(props) => props.usesBlockStyling && BlockStylingMixin}
 `;
 
 const Label = styled(InputLabel)`
 	padding-left: 5px;
-	color: ${(props) => props.theme.input.colorLabel};
 	line-height: 15px;
 	flex: 1;
-	cursor: pointer;
+	cursor: inherit;
```
Fweddi, line 38: cursor and color styles lifted to the container, which can be conditionally styled depending on the descendant input state.

fronts-client/src/components/form/ArticleMetaForm.tsx
```diff
@@ -763,45 +764,91 @@ class FormComponent extends React.Component<Props, FormComponentState> {
...
-											component={InputCheckboxToggleInline}
-											label="Show video"
-											id={getInputId(cardId, 'show-video')}
-											type="checkbox"
-											onChange={() => this.changeImageField('showMainVideo')}
+										<Field
+											component={InputRadio}
+											disabled={
+												editableFields.indexOf(this.getImageFieldName()) === -1
+											}
```
Fweddi, line 792: Previously unusable buttons were hidden; now we prefer to disable them. We could in future include a label explaining why the button is disabled.

PR #1781, opened 2025-03-19
Rewrite handle move function

Georges-GNM:
Co-authored-by: Anna Beddow <anna.beddow@guardian.co.uk>

## What's changed?

Resolves [this ticket](https://trello.com/c/a16rDa8J/1012-add-group-limits-to-fronts-tool)

A previous [PR](https://github.com/guardian/facia-tool/pull/1753) updated the `handleMove` function to consider the new flex gen groups (especially when those groups already had the maximum number of items) when moving cards around, through recursion. Unfortunately this led to a couple of bugs and unexpected behaviour. 

This PR rewrites that function, removing the recursive nature and instead using a `queue` to identify which cards need to be moved, and then call the `moveCard` function on them. 

When we try to move a card, there's various scenarios to consider: 

- if the group we're moving the card to is not full, or we're moving the card within the same group, or the group we're moving to is the standard group, then we just carry on with the move operation. 
- if this isn't the case, i.e. we're moving to a different flex gen group (a splash, very big, or big) which is already full, then we set up a `queue`. 

Again, there's a couple of variations depending on: 
- if we're moving the card to the top of a full group, in which case the story that was already in the group should drop down to the next group. In this case, `buildMoveQueue` adds the original card with the original move data, and loops over the groups via `groupsData` to identify which other cards need to be moved from one group to the next. 
- if we're moving the card to the bottom of a full group, in which case the card we're moving should be the one to drop down to the next group. In this case, `buildMoveQueue` amends the original card so that it is moved to the correct location, and then loops over the appropriate groups to identify what other cards need to be moved. 

In both cases, if the next group is full, similar behaviour should occur and the last pre-existing story in the group should drop down to the next, until we reach a standard group, in which case we just insert the story at the top. 

## Recording

https://github.com/user-attachments/assets/ba69a7ca-284d-4d41-9409-e130da5b692b

### General
- [X] 🤖 Relevant tests added
- [X] 🔍 Checked on CODE

### Client


fronts-client/src/lib/dnd/Level.tsx
```diff
@@ -206,12 +206,14 @@ class Level<T> extends React.Component<Props<T>, State> {
...
-					index,
-				};
+			const from =
+				forceClone && isMove(this.props.parents, parents)
+					? {
+							type: parentType,
+							id: parentId,
+							index,
+						}
+					: null;
```
abeddow91, line 216: This refactor allows us to always have the from type as 

```	from: null | PosSpec;```
as opposed to 
```from: false | PosSpec;``` and then coercing false to null elsewhere

File: fronts-client/src/components/FrontsEdit/FrontContent.tsx [hunk 2]
```diff
@@ -94,6 +94,113 @@ interface FrontState {
...
+function getNextGroupTarget(
+	currentGroupId: string,
+	groupIds?: string[],
+	groupsData?: Group[],
+) {
+	if (!groupIds || !groupsData) {
+		return;
+	}
+	const currentIndex = groupIds.findIndex((id) => id === currentGroupId);
+	const nextGroupId = groupIds[currentIndex + 1];
```
Fweddi, line 106: `groupsIds` could be an empty array, in which case this might cause a runtime error?

fronts-client/src/components/FrontsEdit/FrontContent.tsx
```diff
@@ -94,6 +94,113 @@ interface FrontState {
...
+	const nextGroup = groupsData.find((group) => group.uuid === nextGroupId);
```
Fweddi, line 107: If `nextGroup` is undefined, should we return early? This one might be okay.

fronts-client/src/components/FrontsEdit/FrontContent.tsx
```diff
@@ -94,6 +94,113 @@ interface FrontState {
...
+	 *  Begin cascading shifts for full groups.
+	 *  Start at the target group (or one after, if we adjusted for a bottom insert),
+	 *  and continue shifting the last card of each full group into the next group.
+	 * */
+
+	const startIndex = to.groupIds?.indexOf(to.id) ?? 0;
+
+	for (
+		let index = isBottomInsert ? startIndex + 1 : startIndex;
+		to.groupsData && index < to.groupsData.length - 1;
```
Fweddi, line 162: Does this `-1` offset correspond to the potential `+1` in the line above?

fronts-client/src/components/FrontsEdit/FrontContent.tsx
```diff
@@ -178,127 +285,37 @@ class FrontContent extends React.Component<FrontProps, FrontState> {
...
-		// or any group that is either empty or doesn't have the max items already,
-		// or if we're moving the article within the same group,
-		// then we just move the article to the location
+		const numberOfCardsInGroup = move.to.cards?.length ?? 0;
+		const targetGroupIsFull = move.to.groupMaxItems === numberOfCardsInGroup;
+
+		/**
+		 *  If:
+		 * - the target group isn't full
+		 * - or the card is already in the target group
```
Fweddi, line 296: i.e. the card is moving within the target group?

fronts-client/src/components/FrontsEdit/FrontContent.tsx
```diff
@@ -178,127 +285,37 @@ class FrontContent extends React.Component<FrontProps, FrontState> {
...
+		 * - or it's a standard group
```
Fweddi, line 297: Does this assume the `standard` is the bottom group? If so we might want to make this logic more abstract.

Fweddi (approved):

fronts-client/src/components/FrontsEdit/FrontContent.tsx
Georges-GNM: Not quite, it was there to avoid the cascading logic to trigger when we've reached the last group - but we've refactored to make it more explicit.

PR #1793, opened 2025-04-07
Allow any card to have a video

Fweddi:
The core of this PR is to allow a user to replace the trail image / main media atom with a media atom. This means that any card can now have a video, previously only cards with main media atoms could display video.

To test, you will need to enable the 'Replacement Video' feature switch using the `/v2/features` endpoint. If running locally, you will also need to pull the latest config by running the `setup.sh` script.

| Before | After |
|--------|--------|
| image | image |
| image | image |

#### Video Selection
https://github.com/user-attachments/assets/93a30274-9331-44c9-9e6b-bdda110e5072

#### Media Atom Maker integration
https://github.com/user-attachments/assets/6e42dcba-3edb-487e-a2e5-9a4c69e95e5b

#### Preview video
https://github.com/user-attachments/assets/d486408e-ae33-48b9-9da2-28af37e85325

### Features
image

* Allow users to provide a replacement media atom using a 'video url' field. This field can parse:
	1. A CAPI url (e.g. https://content.guardianapis.com/atom/video/[ATOM-ID])
	2. A Media Atom Maker url (e.g. https://video.gutools.co.uk/videos/[ATOM-ID])
* Implement an 'Replace video' button which opens a Media Atom Maker modal - similar to Grid for images[^1]
	* This is an alternative way of providing a video
	* If a video is provided by this modal the 'video url' field is kept in sync
* Implement a  'Preview video' button which opens a Preview modal for the current video. The current video is either:
	* The main media video, if one is present
	* The 'replacement' video
* If no main media video is present, default to the replacement video (but don't refer to it as a 'replacement' in the UI)
* If there is both a main media and replacement video, provide a 'Use Replacement Video' toggle button
	* This button defaults to _on_ when a replacement video is added by the user, but can be toggled _on_ and _off_
	* This button is disabled if no replacement video is provided (with an explainer: _Replacement video required_)
* Allow the 'replacement' video to be removed from the card. This can be done either by:
	1. Changing / clearing the 'video url' field
	2. Using the two-stage delete button (the same button used to disassociate images)
* Prevent saving the card when a valid video is not provided (with a warning: _You need to provide a valid video_)
* Add a tooltip to explain what happens if the selected video cannot be played
* Change the styling around the Media Select area (better padding and flex behaviour, improved hover / focus)
* Increase the bounding box for block-style radio buttons (now any part can be clicked)
* Warn users not to upload self-hosted video (as it is currently not supported)

[^1]: For Panda / rights reasons, in CODE and local Fronts Tool we need to open CODE MAM, which provides atoms from CODE CAPI. Unfortunately Fronts Tool reads from PROD CAPI (following the departmental principle to connect CODE services to PROD services where possible). This leads to errors locally and in CODE. For my own two cents, it would probably be best if we connected CODE Fronts Tool to CODE CAPI and ported over data from PROD to CODE CAPI on a very frequent basis. The ideological argument for connecting to PROD services encounters too many real world counterexamples for it to be worthwhile. In the meantime - expect the MAM modal in this PR to fail on local and CODE stages.


Fweddi:
Closing in favour of a group of smaller PRs.

PR #1804, opened 2025-04-29
Retain boost level if moving to the same group type

Fweddi:
## What's changed?
We had the following report from Editorial: 
> Currently in News when you move a story with a boost on it in a standard slot to a new slot by dragging it it loses it's boost and you need to manually tick boost again. It should be retaining

We realised this was a reproducible issue. In our logic for changing boost when moving cards, the default case is to reduce the boost level. This PR adds a new case - when we are moving between groups of the same type (e.g. standard -> standard, very big -> very big), we retain the boost level.

I have added some unit tests to give us increased confidence.

### How to test
In a `flexible/general` or `flexible/special` container, move a card with a boost from a Standard group to another Standard group, or the same Standard group. The boost level should remain the same.

Other boost level adjustments should be unaffected.


domlander (approved):
Looks good, thanks for adding those tests

PR #1806, opened 2025-04-29
Add tools audit pixel tracking

waisingyiu:
## What's changed?

This PR adds the [Tools Audit tracking pixel](https://github.com/guardian/editorial-tools-user-telemetry-service?tab=readme-ov-file#tracking-pixel) to allow us to understand more how the facia tool is being used by different users. The results will be visible in the [tools audit dashboard](https://metrics.gutools.co.uk/d/aegzv59r9eghsd/tool-audit?orgId=1&var-domain=birthdays.gutools.co.uk&var-path=All&var-interval=1d&var-chosen_apps=All).

## Implementation notes

We aim to add pixel tracking to all pages that can be served to authenticated users.  

Please note that some of the pages are changed via client-side JS so we cannot rely completely on the browser to send the tracking data.  We added a new hook on top of `useLocation` to send the tracking data when the location path has changed.  

This `useLocation` is available for react-router 5.1.0 or above so we also did a version bump on it.  Despite its major version change, it is [a minor release](https://github.com/remix-run/react-router/releases/tag/v5.0.0).

The changes have been validated on CODE.  I could see the requests to the telemetry service as a result of the pixel tracking when I loaded the fronts tool, the config tool and troubleshoot page.

| Requests |
| --- |
|  |
|  |
|  |

### General
- [ ] 🤖 Relevant tests added
- [X] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/views/V2App/app.scala.html
+++ b/app/views/V2App/app.scala.html
@@ -65,13 +65,7 @@
         <script async defer src="@pinboardUrl"></script>
     }
 
-	@maybeTelemetryUrl.map { telemetryUrl => {
-        <script type="module">
-			const image = new Image();
-			image.src = `@Html(telemetryUrl)&path=${window.location.pathname}`;
-		</script>
-		}
-	}
+	@templates.pixel_tracking(maybeTelemetryUrl)
 </body>
 
 </html>
--- a/app/views/V2App/app.scala.html
+++ b/app/views/V2App/app.scala.html
@@ -5,7 +5,8 @@
     faviconLocation: String,
     clientConfigJson: String,
     isDev: Boolean,
-    maybePinboardUrl: Option[String]
+    maybePinboardUrl: Option[String],
+	maybeTelemetryUrl: Option[String]
 )
 <!DOCTYPE html>
 <html lang="en">
@@ -52,17 +53,25 @@
             }
             window.$RefreshSig$ = () => (type) => type
             window.__vite_plugin_react_preamble_installed__ = true
-        </script>
+		</script>
         <script type="module" src="http://localhost:5173/assets/fronts-client-v2/@@vite/client"></script>
         <script type="module" src="http://localhost:5173/assets/fronts-client-v2/src/index.tsx"></script>
     } else {
-        <link rel="stylesheet" href="@cssLocation">
+		<link rel="stylesheet" href="@cssLocation">
         <script type="module" src="@jsLocation" type="application/javascript"></script>
     }
 
     @maybePinboardUrl.map { pinboardUrl =>
         <script async defer src="@pinboardUrl"></script>
     }
+
+	@maybeTelemetryUrl.map { telemetryUrl => {
+        <script type="module">
+			const image = new Image();
+			image.src = `@Html(telemetryUrl)&path=${window.location.pathname}`;
+		</script>
+		}
+	}
 </body>
 
 </html>
--- a/app/views/V2App/app.scala.html
+++ b/app/views/V2App/app.scala.html
@@ -64,8 +64,6 @@
     @maybePinboardUrl.map { pinboardUrl =>
         <script async defer src="@pinboardUrl"></script>
     }
-
-	@templates.pixel_tracking(maybeTelemetryUrl)
 </body>
 
 </html>
--- a/app/views/priority.scala.html
+++ b/app/views/priority.scala.html
@@ -1,4 +1,4 @@
-@(identity: Option[com.gu.pandomainauth.model.User] = None, stage: String, isDev: Boolean, displayV2Message: Boolean = false, priority: String = "")
+@(identity: Option[com.gu.pandomainauth.model.User] = None, stage: String, isDev: Boolean, displayV2Message: Boolean = false, priority: String = "", maybeTelemetryUrl: Option[String] = None)
 
 @import switchboard.SwitchManager
 
@@ -36,5 +36,13 @@
             </div>
         </main>
     }
+
+	@maybeTelemetryUrl.map { telemetryUrl => {
+        <script type="module">
+			const image = new Image();
+			image.src = `@Html(telemetryUrl)&path=${window.location.pathname}`;
+		</script>
+		}
+	}
     </body>
 </html>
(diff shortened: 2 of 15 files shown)


app/views/priority.scala.html
```diff
@@ -36,5 +36,13 @@
...
         </main>
     }
+
+	@maybeTelemetryUrl.map { telemetryUrl => {
+        <script type="module">
+			const image = new Image();
+			image.src = `@Html(telemetryUrl)&path=${window.location.pathname}`;
+		</script>
+		}
+	}
```
SHession, line 46: Would it be possible to move this into a reusable Twirl component to avoid the repetition? I'm also curious, do we need to the `type="module"` attribute?

SHession (commented):
Nice job! This looked like a tricky on to integrate the tracking into. 

I spotted one case where we the tracking pixel doesn't fire, when transitioning between pages i.e. navigating to https://fronts.code.dev-gutools.co.uk/v2/ -> [Any of the listed links]

app/views/priority.scala.html
waisingyiu: Great idea!  I've moved the snippet into a reusable twirl component in the next commit.

If we set the attribute `type="module"`, the code inside will be executed as a ES module and so the variables/constants created there will not be in the global scope.  I think it could minimize the risk of accidentally breaking the existing JS code (though it is highly unlikely there is really a constant `image` in the global scope).

waisingyiu:
> Nice job! This looked like a tricky on to integrate the tracking into.
> 
> I spotted one case where we the tracking pixel doesn't fire, when transitioning between pages i.e. navigating to https://fronts.code.dev-gutools.co.uk/v2/ -> [Any of the listed links]

Yeah @SHession I've noticed it too.  It appears that clicking those links on the `v2` homepage does not cause page reload.  Apparently, it is handled by the [client side JS](https://github.com/guardian/facia-tool/blob/main/fronts-client/src/components/App.tsx#L151-L156) so the browser does not run the pixel tracking code again, as mentioned in the implementation notes.

Any suggestions are great.

app/views/priority.scala.html
SHession: Good point, I hadn't thought of that

waisingyiu:
> > Yeah @SHession I've noticed it too. It appears that clicking those links on the `v2` homepage does not cause page reload. Apparently, it is handled by the [client side JS](https://github.com/guardian/facia-tool/blob/main/fronts-client/src/components/App.tsx#L151-L156) so the browser does not run the pixel tracking code again, as mentioned in the implementation notes.
> > Any suggestions are great.
> 
> Sorry I missed your comment in the description. I've not very familiar with the fronts tool, but it looks we are using React router here:
> 
> https://github.com/guardian/facia-tool/blob/309162a28472478d0b264a71dca7bedd236d6575/fronts-client/src/index.tsx#L82-L91
> 
> I wonder if we might be able to add a hook to App component which requests the tracking pixel when the location changes, similar to this example:
> 
> https://v5.reactrouter.com/web/api/Hooks/uselocation
> 
> Let me know what you think or if it's worth pairing.

Hey @SHession that's a great suggestion, and it can give us very useful insights on how the tool is used.  

The `useLocation` is available for react-router v5.1.0 or above, so I've upgraded it from v4.2.0 to this version.  Despite the major ver bump, it is only [a minor release](https://github.com/remix-run/react-router/releases/tag/v5.0.0)

fronts-client/src/components/App.tsx
```diff
@@ -139,28 +141,44 @@ const BackgroundHeader = styled.div`
...
-);
+function usePixelTracking() {
+	const location = useLocation();
+	React.useEffect(() => {
+		if (pageConfig.telemetryUrl) {
+			void fetch(`${pageConfig.telemetryUrl}&path=${location.pathname}`, {
+				credentials: 'include',
+			}).catch(() => {
+				console.error('Fail to send pixel tracking to telemetry service.');
+			});
```
SHession, line 152: Nothing wrong with making the call in this way, but making a request via an image element may be a simpler way to make a dumb GET call.

app/views/V2App/app.scala.html
```diff
@@ -52,17 +53,19 @@
...
-        <link rel="stylesheet" href="@cssLocation">
+		<link rel="stylesheet" href="@cssLocation">
         <script type="module" src="@jsLocation" type="application/javascript"></script>
     }
 
     @maybePinboardUrl.map { pinboardUrl =>
         <script async defer src="@pinboardUrl"></script>
     }
+
+	@templates.pixel_tracking(maybeTelemetryUrl)
```
SHession, line 68: I think we should be able to remove this component as the tracking will be fired by the client script, causing duplicate visit events.

fronts-client/src/components/App.tsx
```diff
@@ -139,28 +141,44 @@ const BackgroundHeader = styled.div`
...
-			<AppFonts />
-			<OptionsModal />
-		</DropDisabler>
-	</ThemeProvider>
-);
+function usePixelTracking() {
+	const location = useLocation();
+	React.useEffect(() => {
+		if (pageConfig.telemetryUrl) {
+			void fetch(`${pageConfig.telemetryUrl}&path=${location.pathname}`, {
```
SHession, line 148: It looks like `location.pathname` return the relative path, meaning in the case we are dropping the `/v2`:

image

I expect if we were to call `window.location.pathname` here it would work as we would hope.

SHession (commented):
This looks like it's working nicely, just a couple of little things to address.

> The useLocation is available for react-router v5.1.0 or above, so I've upgraded it from v4.2.0 to this version. Despite the major ver bump, it is only [a minor release](https://github.com/remix-run/react-router/releases/tag/v5.0.0)

Haha, everyone makes mistakes!

app/views/V2App/app.scala.html
waisingyiu: A great spot!

fronts-client/src/components/App.tsx
waisingyiu: Yeah, `window.location.pathname` works for us.  You are right that we need to use `useLocation` hook here.

dblatcher (approved):
I checked the changes made since Sam's reivew and can see the latest version works on CODE - LGTM 👍

PR #1809, opened 2025-05-07
Fix issue moving sublinks out of flex general cards

davidfurey:
## What's changed?

ID is only a group id if the type of the entity is a group.

Adding a check for `type === 'group'` before extracting the id fixes a console error "Cannot read properties of undefined (reading 'id') cause by `selectGroups(state)[maybeFromGroupId]` returning undefined (not null) when you try to index the map of groups using an item id


abeddow91 (approved):
Good catch! Checked on CODE and all seems good 👌

PR #1812, opened 2025-05-19
Log breaking news payload for better debugging

aracho1:
## What's changed?

At the moment when facia sends a breaking news, it does not log any information about the content.

image

This makes it difficult for us track which notifications were sent and debug if any of them were not correctly processed by the downstream services e.g. mobile notifications. 

This PR therefore updates the logging messages to include the notification payload so that we can use the information to debug any breaking news alert issues in the future.

## I sent the following test notification on facia code.
image

### Facia logs
Now the Facia logs show information about the notification id, headline and capi id.
image

### Mobile logs 
We can check if the corresponding notification was correctly processed in our mobile services (for instance by querying the id).
image
### General
- [ ] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/updates/BreakingNewsUpdate.scala
+++ b/app/updates/BreakingNewsUpdate.scala
@@ -282,23 +282,23 @@ class BreakingNewsClientImpl(ws: WSClient, host: String, apiKey: String)
       .post(body)
       .map { response =>
         if (response.status >= 200 && response.status < 300) {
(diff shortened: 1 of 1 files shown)


Georges-GNM (approved):
Haven't tested myself, but this looks good and much more helpful 👏

PR #1814, opened 2025-05-21
Update to latest facia-scala-client

Fweddi:
### Changelist

* Allow users to provide a replacement media atom using a 'video url' field. This field can parse:
	* A CAPI url (e.g. [https://content.guardianapis.com/atom/video/[ATOM-ID]](https://content.guardianapis.com/atom/video/%5BATOM-ID%5D))
	* A Media Atom Maker url (e.g. [https://video.gutools.co.uk/videos/[ATOM-ID]](https://video.gutools.co.uk/videos/%5BATOM-ID%5D))
* Update to latest facia-scala-client (bringing in https://github.com/guardian/facia-scala-client/pull/355)
* Extract video base URL (i.e. Media Atom Maker) from config - this field has been added to local, CODE and PROD configs in S3


fronts-client/src/components/video/VideoControls.tsx
```diff
@@ -0,0 +1,70 @@
...
+	videoBaseUrl,
+	mainMediaVideoAtom,
+	replacementVideoAtom,
+	showMainVideo,
+	showReplacementVideo,
+	changeField,
+	changeMediaField,
+	form,
+	replacementVideoControlsId,
+	warningsContainerId,
```
Fweddi, line 36: Not all of this fields are relevant here, but will be used in subsequent PRs.

fronts-client/src/util/CAPIUtils.ts
```diff
@@ -81,6 +82,14 @@ function getThumbnail(
...
 		return meta.slideshow[0].src;
+	} else if (meta.videoReplace) {
+		const atomProperties = extractAtomProperties(meta.replacementVideoAtom);
+		if (
+			atomProperties !== undefined &&
+			atomProperties.trailImage !== undefined
+		) {
+			return atomProperties.trailImage;
+		}
 	}
```
abeddow91, line 93: I might be wrong but I think we're still unsure as to if we want the trail image of the video. It makes sense to show this on the expanded card view where the replaced video is, but I'm not sure we should be using it on the collapsed card view.

fronts-client/src/components/image/ImageMetaDataContainer.tsx
```diff
@@ -11,19 +11,44 @@ interface ImageMetadataContainerProps {
...
+) => {
+	if (imageSlideshowReplace) return 'Slideshow';
+	if (imageReplace) return 'Image replaced';
+	if (imageCutoutReplace) return 'Cutout replaced';
+	if (showMainVideo) return 'Main video';
+	if (hasMainVideo && videoReplace) return 'Video replaced';
+	if (videoReplace) return 'Show video';
+	return null;
+};
+
```
abeddow91, line 35: 👌

fronts-client/src/components/video/VideoControls.tsx
```diff
@@ -0,0 +1,70 @@
...
+					component={VideoUriInput}
+					type="text"
+					onChange={(e: any) => {
+						const videoUri = e.currentTarget.value;
+
+						if (
+							videoUri !== undefined &&
+							videoUri !== null &&
+							videoUri !== ''
+						) {
```
abeddow91, line 56: Could we simplify this to the below? It will also treat 0, false, and NaN as falsy too but I think thats ok here.

```suggestion

						if (videoUri) {
```

fronts-client/src/util/extractAtomId.ts
```diff
@@ -0,0 +1,136 @@
...
+
+	return undefined;
+};
+
+const extractAtomId = (videoUri: string | undefined): string => {
+	/*
+	 * This method works for both PROD and CODE paths, irrespective of the stage of the Fronts Tool.
+	 *
+	 * We could choose to pass a stage parameter, but I think it's better to be flexible, given
+	 * CODE Fronts Tool can be pointed to PROD or CODE CAPI.
```
abeddow91, line 27: 👍

fronts-client/src/util/extractAtomId.ts
```diff
@@ -0,0 +1,136 @@
...
+		urlConstants.base.capi,
+		urlConstants.video.capiAtomPath,
+	);
+	if (capiAttempt !== undefined)
+		return `${urlConstants.video.capiAtomPath}${capiAttempt}`;
+
+	return '';
+};
+
+const extractAssetId = (atom: Atom): string | undefined => {
```
abeddow91, line 52: I don't think this does return undefined, it will either return the id string or throw an error. The next two functions for extracting the platform and video trail seem to prefer swallowing this error with logging and returning an undefined. Do we want to throw here?

fronts-client/src/util/extractAtomId.ts
```diff
@@ -0,0 +1,136 @@
...
+		assets === undefined ||
+		assets.length === 0 ||
+		assets[0] === undefined ||
+		assets[0].id === undefined
+	) {
+		throw new Error(`No assets found for atom ${atom.id}`);
+	} else {
+		return assets[0].id;
+	}
+};
```
abeddow91, line 64: Non-blocking suggestion - we could simplify this with optional chaining eg
```suggestion
  const assetId = atom.data?.media?.assets?.[0]?.id;
  if (!assetId) {
    throw new Error(`No assets found for atom ${atom.id}`);
  }
  return assetId;
};
```

fronts-client/src/util/extractAtomId.ts
```diff
@@ -0,0 +1,136 @@
...
+		imageAssets.length === 0 ||
+		imageAssets[0] === undefined ||
+		imageAssets[0].file === undefined
+	) {
+		console.error(`No trail image found for atom ${atom.id}`);
+		return undefined;
+	} else {
+		return imageAssets[0].file;
+	}
+};
```
abeddow91, line 82: See above for re simplifying with optional chaining

fronts-client/src/util/extractAtomId.ts
```diff
@@ -0,0 +1,136 @@
...
+	if (
+		mediaAssets === undefined ||
+		mediaAssets.length === 0 ||
+		mediaAssets[0] === undefined ||
+		mediaAssets[0].platform === undefined
+	) {
+		console.error(`No media assets found for atom ${atom.id}`);
+		return undefined;
+	} else {
+		return mediaAssets[0].platform;
```
abeddow91, line 96: See above for re simplifying with optional chaining

abeddow91 (approved):
Looks great, a few questions and some optional feedback but non blocking 👍

fronts-client/src/util/CAPIUtils.ts
Fweddi: I think both images should be the same. The collapsed view should show whatever the current media is set to. If you look at the 'Replace image' as an analogy, we show the Replacement image in both the collapsed and expanded view. I think we should do the same for video.

PR #1823, opened 2025-06-02
Add a shortcut to close the Overlay modal

Fweddi:
## What's changed?
Adds a shortcut to close the Overlay modal - press the Escape key.

iFrames steal keypress events when focussed - so we need to post events up from the iFrame in these contexts: https://github.com/guardian/media-atom-maker/pull/1204

### To test
Toggle the 'Enable replacement video' feature switch on (e.g. https://fronts.local.dev-gutools.co.uk/v2/features).

Find / create a card with a video. Open the Video Preview modal. You should be able to exit using the Escape key.

This will also work with the Video Replace modal (which opens Media Atom Maker) and the Image Replace modal (which opens Grid), so long as the iFrame does not have focus.

https://github.com/guardian/media-atom-maker/pull/1204 fixes this for Media Atom Maker by passing the keypress event up. Grid fix pending.

### DIFF
--- a/fronts-client/src/components/modals/OverlayModal.tsx
+++ b/fronts-client/src/components/modals/OverlayModal.tsx
@@ -1,7 +1,7 @@
 import { styled } from '../../constants/theme';
 import Modal from 'react-modal';
 import ButtonDefault from '../inputs/ButtonDefault';
-import React from 'react';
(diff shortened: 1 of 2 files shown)


Georges-GNM (approved):
Working locally, so lgtm 👍

PR #1839, opened 2025-06-18
Add tests for resetting boost levels

abeddow91:
## What's changed?
@Georges-GNM quite rightly pointed out that there are no tests for the logic that determines if a cards boost level should be reset - this PR adds test coverage for this logic.

As part of writing these tests, there has been a small refactor to lift the check to ensure the card is in a flexible general collection check outside of the mayResetBoostlevel function. This helps simplify the function and also ensure we are in a flexible general container before attempting to reset the card's boost level.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/actions/Cards.ts
+++ b/fronts-client/src/actions/Cards.ts
@@ -14,12 +14,7 @@ import {
 	maybeAddFrontPublicationDate,
 	copyCardImageMeta,
 } from 'actions/CardsCommon';
-import {
-	selectCards,
-	selectCard,
-	selectArticleGroup,
-	selectGroups,
-} from 'selectors/shared';
+import { selectCards, selectCard, selectArticleGroup } from 'selectors/shared';
 import { ThunkResult, Dispatch } from 'types/Store';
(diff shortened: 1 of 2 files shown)


fronts-client/src/actions/Cards.ts
```diff
@@ -228,19 +224,15 @@ const isFlexibleGeneralContainer = (state: State, collectionId: string) => {
...
 	card: Card,
 	persistTo: 'collection' | 'clipboard',
 ) => {
-	if (to.type !== 'group' || !to.collectionId || persistTo !== 'collection')
-		return;
+	if (to.type !== 'group' || persistTo !== 'collection') return;
 	if (from?.id === to.id) return;
-	if (!isFlexibleGeneralContainer(state, to.collectionId)) return;
-	const toGroup = selectGroups(state)[to.id];
-	const groupName = toGroup?.name ?? 'standard';
```
abeddow91, line 243: We can access the group name from `to` so we don't need to make a separate check to retrieve it.

PR #1800, opened 2025-04-23
Add check for opinion section and hide image if there's no cutout

Georges-GNM:
## What's changed?

Closes this [ticket](https://trello.com/c/pZUNi59j/1140-opinion-toggle-change)

This updates some logic when the content is being pulled in to the tool: for opinion articles, the default is to choose the cutout photo - but if there's no cutout, we default to hide image rather than fall back to a trail image. 

## Video

https://github.com/user-attachments/assets/4ec1da8b-f275-43ae-aeec-2fb4436e8bec

### General
- [ ] 🤖 Relevant tests added
- [ ] ✅ CI checks / tests run locally
- [ ] 🔍 Checked on CODE

### Client
- [ ] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ ] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #1843, opened 2025-06-24
Pass imageHide and imageCutoutReplace when creating an article from drop

Georges-GNM:
Resolves this [F&C ticket](https://trello.com/c/Ya4Wy5n6/1285-opinion-hide-image-bug)

## What's changed?

We [previously added logic ](https://github.com/guardian/facia-tool/pull/1800) to determine if an opinion card should have the image hidden by default. This worked for the most part, except in the specific case of adding an opinion article to the clipboard by pasting it in the application (rather than adding it from the feed). 

That is to say, when an editor added an opinion article to their clipboard through ctrl+v and then moved it to an actual collection, the card generated which should have had its image hidden had the `imageHide` property set to false, which led it to unexpectedly display an image on platforms. Confusingly, the tool's UI indicated the setting was correctly set. 

On investigation, this seems to be because when creating the card through this method, we were just setting a number of defaults, rather than using the meta generated from fetching the content from CAPI. The default for `imageHide` was set to false, so it was always coming through as false. Now, when creating the article, we point to the default meta settings, which seems to allow the desired value for `imageHide` to correctly come through. 

## Note

To be honest, I'm still unsure as to why the article in the tool correctly had the imageHide setting to true when the data being sent to the platforms had it as false. I tried to figure this out, and I'm guessing it's related to state and an attempt to denormalise card data. Looking at the state when a card gets added to a collection, for example, shows a lengthy list of cards, some of which with the same internal code but different uuids, and some of which seem quite out of date, such as cards that I've removed from a collection. 

This gave me the sense that the object representing the card in the UI is in effect a different one than what gets added to the request when updating a collection, and although I'd assume the theory is that those two objects should generally be the same, in this case they didn't appear to be. It does feel worth investigating that to at least understand the tool better, but felt like it was beginning to be a rabbit hole (that was especially hard to follow due to the opaque ways to refer to a card like the uuids). 

TLDR: I'm not a fan of redux. 

## Testing

- add an opinion article with a trail image (not a cutout) to the clipboard by pasting it in (command+v). The UI should show the image grayed out and the Hide Image setting toggled. 
- If you add the article to a collection and launch the front, does the image appear on the front? If not, then this logic is working as intended. 

Also worth checking expected behaviour with other cards is still working (e.g. using an opinion article with a cutout should use the cutout, a regular, non-opinion article with a trail image should display the image). 

## Video

image

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


abeddow91 (approved):
Tested with 
a clipboard card that had been dragged from the feed
a clipboard card that had been pasted
a feed card


all working as expected. Good find!

PR #1845, opened 2025-07-17
quickfix: Remove a space to ensure there's no possible crash on iOS

alexduf:
## What's changed?

Removed a prefixing space character that unfortunately lead to an iOS crash. We'll take extra measures to ensure this is mitigated downstream as well, but this should be fixed here too.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


Divs-B (approved):
That looks good to me 👍 
CI seems failed just try to re-run

PR #1851, opened 2025-08-11
Correctly place overmatter line in Flexible General container

Fweddi:
Paired on with @Georges-GNM 
____

## What's changed?
We have not been passing the collection config to the StoriesVisibleController, leading to a glut of `NoSuchElementException` errors (see https://github.com/guardian/facia-tool/pull/1848). We now correctly inject the collection config, which requires passing the collection id from the client to the server. This should place the overmatter line (derived from the stories visible endpoint response) in the right place.

While investigating the overmatter line, we noticed errors leading to inaccurate line placement. We have refactored the logic to correctly reflect the number of items in each group (capping each group at its max items, rather than taking the group's max items).

## What's the overmatter line?
The four horizontal lines demarcating which stories are visible:

image

### DIFF
--- a/app/slices/FlexibleContainer.scala
+++ b/app/slices/FlexibleContainer.scala
@@ -13,15 +13,30 @@ object FlexibleGeneral extends FlexibleContainer {
       stories: Seq[Story],
       maybeCollectionConfigJson: Option[CollectionConfigJson]
   ): Int = {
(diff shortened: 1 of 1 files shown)


Georges-GNM (approved):
Looks good to me, from testing locally.

PR #1852, opened 2025-08-14
Remove dynamicslowmpu container as this is no longer used.

abeddow91:
## What does this change?
Removes all code related to the dynamic/slow-mpu container.

## Why?
This container is no longer supported or in use on production fronts.

### DIFF
--- a/app/thumbnails/ContainerThumbnails.scala
+++ b/app/thumbnails/ContainerThumbnails.scala
@@ -183,9 +183,6 @@ class ContainerThumbnails(val fixedContainers: FixedContainers) {
       case "dynamic/package" =>
         Some(Seq(ThreeQuarterTallQuarter2))
 
-      case "dynamic/slow-mpu" =>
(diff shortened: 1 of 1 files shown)


app/thumbnails/ContainerThumbnails.scala
```diff
@@ -221,5 +218,15 @@ class ContainerThumbnails(val fixedContainers: FixedContainers) {
...
+    // Default placeholder so callers never get None
+    svgOpt.orElse(
+      Some(
+        <svg xmlns="http://www.w3.org/2000/svg" width="100" height="40">
+			  <rect x="0" y="0" width="100" height="40" style="fill:#eee;stroke:#ddd;stroke-width:1"/>
+			  <text x="50" y="22" text-anchor="middle" style="font:10px Arial;">No preview</text>
+		  </svg>
+      )
+    )
+
```
Fweddi, line 230: assuming this is a drive-by cleanup op 👍

PR #1859, opened 2025-09-22
Support subtitles track

Fweddi:
## What's changed?

Brings in new subtitles asset type from https://github.com/guardian/content-atom/pull/178.

Add the ability to preview subtitles in the Preview Video modal. Rather than using `hls.js` (which would be another dependency to maintain), use a `video` html tag with a source for the mp4 and a source for the vtt (as we do in Media Atom Maker).

### Screenshots

image

____

image

### DIFF
--- a/fronts-client/src/components/video/VideoControls.tsx
+++ b/fronts-client/src/components/video/VideoControls.tsx
@@ -83,6 +83,7 @@ const MarginWrapper = styled.div`
 const Video = styled.video`
 	width: 100%;
 	height: min-content;
+	max-height: 90vh;
 `;
 
 export const VideoControls = ({


fronts-client/src/util/extractAtom.ts
```diff
@@ -71,44 +62,54 @@ const extractVideoImage = (atom: Atom): string | undefined => {
...
-	if (atomProperties === undefined) {
-		return undefined;
-	}
-
-	return atomProperties?.platform === 'youtube'
-		? `https://www.youtube.com/embed/${atomProperties.assetId}`
-		: atomProperties?.assetId;
+const extractAssetsByMimeType = (atomAssets: AtomAsset[]) => {
+	return {
+		m3u8: findAssetByMimeType(atomAssets, 'application/x-mpegURL'),
```
gnm-notona, line 80: In the scala code we're using `application/vnd.apple.mpegurl`

Fweddi: _great_ catch

PR #1868, opened 2025-10-17
no v2 banner on breaking news

dblatcher:
## What's changed?

Updates the arguments sent by the ViewsController to the admin_main template to set `displayV2Message` to false for the breaking-news pages.

The "breaking news tool" is not replicated in the v2 tool, so rendering the banner suggesting user should not be using it is somewhat misleading and might lead to confusion for new users.

|before|after|
|---|---|
|.Screenshot 2025-10-20 at 09 43 30  |Screenshot 2025-10-17 at 15 12 33|

## Implementation notes

If we actually do produce an alternative to the current breaking news page, we could update the template to display a more specific banner and eventually put in a redirect to the new URL when the the old version can be removed.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/FeatureSwitches.scala
+++ b/app/model/FeatureSwitches.scala
@@ -26,18 +26,10 @@ object PageViewDataVisualisation
       enabled = true
     )
 
-object TenImageSlideshows
-    extends FeatureSwitch(
-      key = "ten-image-slideshows",
-      title = "Allow slideshows to contain 10 images rather than 5",
-      enabled = false
-    )
-
 object FeatureSwitches {
   val all: List[FeatureSwitch] = List(
     ObscureFeed,
-    PageViewDataVisualisation,
-    TenImageSlideshows
+    PageViewDataVisualisation
   )
 
   def updateFeatureSwitchesForUser(
--- a/fronts-client/src/actions/__tests__/fixtures/Editions.fixture.ts
+++ b/fronts-client/src/actions/__tests__/fixtures/Editions.fixture.ts
@@ -173,11 +173,6 @@ export const initialState = {
 					enabled: true,
 				},
 				{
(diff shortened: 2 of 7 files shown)


fronts-client/src/declarations/modules.d.ts
```diff
@@ -130,3 +130,8 @@ declare module 'react-router-redux' {
 		path: string,
 	): (state: { router: RouterState }) => match | null;
 }
+
+declare module 'history' {
```
dblatcher: I was getting a typescript compile error on `yarn build` that there was no types for the 'history' package - added this stub to resolve that

public/test/config.js
```diff
@@ -50,7 +50,7 @@ export default {
 		baseUrls: {
 			mediaBaseUrl: 'https://media.test.dev-gutools.co.uk',
 			apiBaseUrl: 'https://api.media.test.dev-gutools.co.uk',
-			videoBaseUrl: 'https://video.code.dev-gutools.co.uk',
-		},
+			videoBaseUrl: 'https://video.code.dev-gutools.co.uk'
+		}
     }
```
dblatcher: the linter for the project was warning about trailing commas - this isn't actually a syntax concern anymore (other projects will be configured to require them) but thought it best to reduce the noise by compiling with the linting in place.

emdash-ie:
I can see the v2 banner was originally not shown on the breaking news pages, according to [this commit as part of the PR introducing it](https://github.com/guardian/facia-tool/pull/922/commits/e777acbc61b6a74c5ff8103cac7cc6481aa2586a). I wonder why that stopped being the case? (I am not curious enough to dig through the history yet)

PR #1881, opened 2025-10-28
breaking news: display text length guidance, with correct count for breaking news

dblatcher:
## What's changed?

Restores a broken feature to show when the headline field is over the maximum length when editing the text from the clipboard and include a revert/undo button to reset the text the previous version (since last saved).

The character count and warning elements were already in the UI, but were being left as "visibility:hidden" in the css as they were never nested in an element with the ".attention" class, which seems to have been removed from the mark-up, but was still expected in the css. The main change was to remove the requirement for ".attention".

The character counts we using a limit of 120 for the headlines - this is set in "article-meta-fields" module, which defines an array of objects configuring the options(including `maxLength`) to use for editing the article meta fields., however, for breaking news, the limit should be 90 characters (defined as the `restrictedHeadlineLength` in "constants/defaults"). To have the headline editor on the breaking news pages  use the `restrictedHeadlineLength` as the `maxLength` for the headline field , there is a new function which returns a modified version of the article meta fields array if the page is a breaking news page.

For extra clarity, the warning message now includes the limit.

Also adds title and aria attributes to the word count related elements - this might make some minor accessibility improvements, but the pages are likely not accessible overall.

**UPDATE**

following feedback from stakeholders and input from @Chloe-Guardian , this PR now:
 - removes the "pop up" alert that prevented users from dragging articles from the clipboard if the headline was >90 characters
 - shows a character count for the text fields, rather than a "characters left" countdown
 - shows the warning text when the users go over 90 character but only highlight the text area and character count in red if they go over 120
 - does not include the "revert" button mentioned above
 - has the same warning / character count ui on the clipboard and "containers" on the right

## Implementation notes

The  pink highlight one the text only shows when the headline editor is in the 'active' state - so not in the items in the feed, even if the "default" headlines would be too long if used. This seemed to be the intent in UI. I looked into changing this, but there seemed to be some performance issues with actively checking the lengths of every item in the list - it also made the page very "noisy" as most items in the feed would initially be over the 90 character limit.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

## images
under 90 characters
Screenshot 2025-11-05 at 16 52 01

exactly 90 characters
Screenshot 2025-11-05 at 16 52 21

over 90 characters
Screenshot 2025-11-05 at 16 52 37

over 120 characters
Screenshot 2025-11-05 at 16 52 57

### DIFF
--- a/public/test/spec/config.front.spec.js
+++ b/public/test/spec/config.front.spec.js
@@ -238,7 +238,7 @@ describe('Config Front', function () {
             var imageUrl = images.path('square.png');
             dom.type('.metadata--provisionalImage', imageUrl);
 
-            return wait.ms(500).then(() => {
+            return wait.ms(1000).then(() => {
                 var front = frontWidget.pinnedFront();
                 expect(persistence.front.update).toHaveBeenCalledWith(front);
                 expect(front.props.webTitle()).toBe('Nicer title');


public/src/css/style.css
```diff
@@ -1254,25 +1254,20 @@ select {
...
-    display: none;
-}
-
-.attention .length--alert .editor__alert__message {
-    display: inline;
 }
 
 .editor__length {
     display: none;
 }
```
SHession: We have two rules for `editor__length`. Should we remove this top one?

SHession (commented):
The changes make sense! 

I wonder if the 120 / 90 character count discrepancy might account for some of the user confusion.

dblatcher:
> The changes make sense!
> 
> I wonder if the 120 / 90 character count discrepancy might account for some of the user confusion.

CP are checking in with the digi-ed's about the change - it seems that users will sometime cut the headline in the clipboard to <90 just so they can get it onto the container, then put back the longer version (>90) and send it out. I don't think we discovered this point in the fact-finding. 

From the code, it doesn't actually look like the character limits are enforced when sending the alerts, unless I'm missing something. 

In short, we might need to re-consider this PR - 90 characters might not actually be the 'correct' limit.

public/src/js/utils/modify-fields-for-breaking-news.js
```diff
@@ -0,0 +1,31 @@
...
+
+const metaFieldsForBreakingNews = Object.freeze(
+    metaFields.map(field => {
+        if (field.key === 'headline') {
+            return Object.assign({}, field, {
+                showCharacterCountInsteadOfCharactersLeft: true,
+                noRevertButton: true,
+                lengthWarningMessages: [
+                    [CONST.restrictedHeadlineLength, `${CONST.restrictedHeadlineLength} characters maximum`]
+                ]
```
SHession: It feels like we could make the messaging here a little clearer. This isn't such a maximum as a suggested limit. 
Also is there any reason to represent `lengthWarningMessages` as an array?

dblatcher: It was an array so we'd have the option of setting different warnings at different character counts - however, the designs send through had "90 characters maximum" at >90 and >120

PR #1961, opened 2026-03-18
Update Breaking News and Sports push notification titles

SHession:
Tests were asserting `"Editors' pick"` while the implementation returns `"Editors' picks"`, causing CI failures.

## Changes

- Updated test assertions in `BreakingNewsUpdateTest` from `"Editors' pick"` → `"Editors' picks"` to align with the value returned by `createPayload`
- Updated test description strings to consistently reflect the plural form


dblatcher (approved):
LGTM - there is a slight Product concern that the change won't be apparent to users, but  that can probably be handled with comms.

If this was a concern, an option would be to add a temporary notice banner in the UI to remind users of the change in title.

PR #1974, opened 2026-04-10
Improve type safety of createAsyncResourceBundle

davidfurey:
# Improve type safety

## Summary

This PR improves type safety across the `createAsyncResourceBundle` utility and its consumers, and fixes several related type issues in bundle definitions and fixtures.

### `createAsyncResourceBundle` — stronger generic types

The bundle factory previously used a single `Resource` type parameter. This meant `selectAll` returned `any`, `State.data` was typed as a union with `any`, and the distinction between indexed (`Record<string, Resource>`) and non-indexed (`Resource`) bundles was not reflected in the types.

The function now takes a second type parameter `T extends boolean` that corresponds to `indexById`, allowing the data type to be inferred as either `Record<string, Resource>` (when `T` is `true`) or `Resource` (when `T` is `false`):

- `State<Resource>.data` is now simply typed as `Resource` — no more `| any`
- `initialData` is now required, ensuring the initial state is always well-typed
- `indexById` is now required (no default)
- `selectAll` returns `DataType` instead of `any` or a union type
- Added a typed `isResourceMap` type guard that replaces the old `!indexById` boolean checks in the reducer, giving TypeScript the information it needs to narrow the data type

### Bundle call sites updated

All bundle definitions have been updated to pass the new required type arguments and `initialData`

### `Collection` type — optional region fields

`targetedRegions` and `excludedRegions` on the `Collection` interface were previously required, causing spurious type errors at call sites where they are absent. They are now optional, and the `selectCollectionTargetedRegions` selector has been updated with a corresponding null check.

### `collectionsBundle`

The 3 `stages` are explicitly defined in tightly typed array to avoid a type error.

### Fixtures cleaned up

The `externalArticles` fixture in `initialState.ts` has been updated to conform to the `CapiArticle` type, avoiding type errors that were previously masked by `any`.


jonathonherbert:
This looks great. Pleasing to see fixtures changing as a result of the better typesafety. Can't see that this changes any code consuming selector output, presumably because outside of the context of the bundle code the assertions on the selectors were (mostly) doing the job.

On the API: using the second type parameter as the means of specifying whether a resource is `T` or `Record<id, T>` is I think surprising and also requires repetition. When I write:

```typescript
const bundle = createAsyncResourceBundle<Thing>('things', {
	indexById: true
});
```

We get the error, **`Expected 2 type arguments, but got 1.`** When I look at the type argument in-situ it's `T`. Perhaps this could be more descriptive; I have to understand how type parameter inference works to move from T to `indexById: T`. AFAIK, there are no comments we can write against the type parameter that might provide inline documentation to explain how to use this. I then have to repeat myself and write `true` in two places.

(If as a developer I'm wondering why I need to repeat myself, I also need to understand that the reason we must also specify the second parameter when we provide the first, even though Typescript will happily compile a call to `createAsyncResourceBundle` with no type parameters at all, is because type inference is all-or-nothing in Typescript; you either provide none of them and Typescript will attempt to infer, or provide all of them explicitly.)

Happy to approve as-is on the basis that improved type safety is a good thing to have, but I'd recommend we refine to find a cleaner API. one suggestion might be to keep the underlying implementation private, and provide separate `create` functions for each resource type (`createAsyncResourceBundle` `createIndexedAsyncResourceBundle`, I'm sure there are better names) that contain comments outlining their use cases. No extra type parameters, clear documentation. WDYT?

jonathonherbert (approved):

davidfurey:
> because type inference is all-or-nothing in Typescript;

Which is sad.

I've pushed some changes based on your feedback. I think it is an improvement. It also has the side benefit that initialState can be optional in some cases.

jonathonherbert (approved):

PR #1979, opened 2026-04-24
Interactive atom search

tomrf1:
Interactive atoms can be added to a front collection as a snap link.
This is done by dragging a url into the clipboard on the left, and then dragging from there into a collection.
Screenshot 2026-04-24 at 16 04 35

This mechanism is used for thrashers.
The card then flows through to Frontend/MAPI as a snap link.

This PR proposes an alternative way to achieve the same thing in the tool. There are two motivations behind this:
1. improved UX for finding and adding interactive atoms, since they are in CAPI.
2. a stepping stone towards supporting other types of atoms. E.g. we are considering implementing thrashers differently (natively in dotcom-rendering, rather than as interactive atoms)

An "Include atoms" checkbox is added to the content search settings:
Screenshot 2026-04-24 at 15 53 51

If ticked, the results will include interactive atoms from CAPI as well as content. They will appear at the bottom of the results list:
Screenshot 2026-04-24 at 15 58 55

In the client, any interactive atoms retrieved this way are modelled as a `INTERACTIVE_ATOM` rather than as a `SNAP_LINK`.
When the user drags an interactive atom from the list into a collection it is still persisted as a snap. This means the only change here is to the client side of the tool.
Screenshot 2026-04-24 at 16 00 47

Downstream systems (e.g. Frontend/MAPI) should be unaffected.


fronts-client/src/constants/cardTypes.ts
```diff
@@ -4,6 +4,7 @@ export const CardTypesMap = {
 	RECIPE: 'recipe',
 	CHEF: 'chef',
 	FEAST_COLLECTION: 'feast-collection',
+	INTERACTIVE_ATOM: 'interactive-atom',
```
davidfurey, line 7: What do you think about making this just `ATOM: 'atom'` rather than also specifying that it is interactive?

tomrf1: This enum value is only used client-side in the tool. The shape of interactive atoms differs from other atoms in CAPI, so I think we'd still need some way of handling them differently in the client.
It's a bit hard to say right now, but I'm inclined to be more specific here, and we can revisit this later when we introduce more atom types to the tool.

fronts-client/src/lib/createAsyncResourceBundle/index.ts
```diff
@@ -21,13 +21,13 @@ interface FetchStartAction {
 	payload: { ids?: string[] | string };
 }
 
-interface FetchSuccessAction<Resource> {
+interface FetchSuccessAction<Resource, Order = string> {
```
tomrf1: The new `Order` type param in this file enables us to process the new `FeedEntry` type, which models both articles + atoms. This means we know how to render each item when processing the feed

fronts-client/src/bundles/capiFeedBundle.ts
```diff
@@ -44,17 +58,34 @@ const {
...
+
+	const atomResults: CapiInteractiveAtom[] = atomsResponse
+		? atomsResponse.response.results
+		: [];
 
 	return {
-		results: checkIsContent(response) ? [response.content] : response.results,
+		results: checkIsContent(response)
+			? [response.content]
+			: [...response.results, ...atomResults],
```
davidfurey, line 88: Would it be difficult to sort the results here so that atoms and articles are in one list? It seems unintuitive for atoms to be at the bottom with no UX hint that this is the case

fronts-client/src/bundles/capiFeedBundle.ts
```diff
@@ -136,14 +167,16 @@ export const fetchPrefill =
 		try {
 			const { response } = await getPrefills(id);
 			if (!checkIsContent(response)) {
+				const filteredResults = response.results.filter(isNonCommercialArticle);
 				dispatch(
-					prefillActions.fetchSuccess(
-						response.results.filter(isNonCommercialArticle, {
+					prefillActions.fetchSuccess(filteredResults, {
+						order: filteredResults.map((item) => item.id),
+						pagination: {
```
davidfurey, line 174: Presumably the pagination of atoms and articles is independent? So given that we publish many more articles than interactive atoms, page 5 will continue quite fresh articles but quite old atoms?

davidfurey (approved):
I've added some feedback but I don't think we should let that block us getting this feature out. 

Before merging we should discuss a comms strategy with CP

fronts-client/src/bundles/capiFeedBundle.ts
tomrf1: Yes good point, perhaps the UX needs a rethink here. If the user is looking for an atom, do they even want to see articles and atoms mixed together?

Georges-GNM:
Just for info - I've picked on up on this branch and [spiked](https://github.com/guardian/facia-tool/pull/2007) an expansion to the functionality to include surfacing more atoms (those created in atom workshop), as part of discovery work for the event kit mission (and thanks a ton @tomrf1 - your work was really helpful) . We're planning to chat to CP about this relatively soon, so I'm happy to bring up the question of UX and release comms with them then.

PR #1989, opened 2026-05-22
Setup dev containers

davidfurey:
## What's changed?

Uses https://github.com/guardian/devenv to setup devcontainers in this project.

See https://github.com/guardian/devenv/blob/main/docs/containerised-development/containerised-development.md for why devcontainers are a good thing.


.devcontainer/devenv.yaml
```diff
@@ -0,0 +1,44 @@
...
+# To disable, comment out or remove items from this list
+modules:
+  - mise
+  - docker-in-docker
+  - scala
+  - node
+
+# Optional: Ports to forward
+# forwardPorts:
+#   - 8080  # same port on host and container
```
waisingyiu, line 22: Do we need to forward the port on which the service is listening on?

scripts/setup.sh
```diff
@@ -118,7 +118,11 @@ main() {
   check_yarn_installed
   install_v1_deps
   install_v2_deps_and_build
-  setup_nginx
+  if [ "$USER" = "vscode" ]; then
```
waisingyiu, line 121: So, do we expect developers to execute "setup.sh" in host environment first (to set up nginx and other stuff) and then run "setup.sh" again in dev container (to copy config files and install deps etc)?

.devcontainer/devenv.yaml
davidfurey: This happens automatically

scripts/setup.sh
davidfurey: For now, yes. But it might make sense to split the script up in future

waisingyiu (approved):
It looks great to me!

PR #1990, opened 2026-05-26
Allow localstack S3 for local development

davidfurey:
## What's changed?
While working on the config tool, I often corrupt some of the CODE fronts config files. Versioning is enabled, so it is easy to fix this and no one has noticed (yet).

But it would be nice not to break the shared CODE environment.

This PR adds a script to create a local bucket (`facia-tool-store-local`) and copy the CODE fronts config to it. Once you've uncommented the appropriate config in `conf/local.conf` you can test changes locally without touching the shared S3 config.


waisingyiu (approved):
It looks great to me!

PR #1991, opened 2026-05-26
Facia-tool doesn't require frontend credentials

davidfurey:
Since the connection between facia and facia-press has been done via SNS+SQS rather than just SQS for a long time.


waisingyiu (approved):
👍

PR #1992, opened 2026-05-26
Remove CAPI credentials

davidfurey:
## What's changed?

Use STS for CAPI for local dev. I cannot see any reason why developers should need explicit CAPI IAM access when they can just use the same STS Assume Role session credentials that PROD/CODE does.

This also changes the Janus link to suggest that developers fetch `run-fronts-tool-locally` developer policy access rather than full cms-fronts account access.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ You can find the client for the Fronts tool in [fronts-client](./fronts-client).
 
 1. Install [NVM](https://github.com/creationix/nvm).
 2. Ensure [Docker](https://www.docker.com/products/docker-desktop) is installed
-3. Get credentials from [Janus](https://janus.gutools.co.uk/multi-credentials?&permissionIds=cmsFronts-dev,capi-api-gateway) (`cmsFronts` & `capi`).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ You can find the client for the Fronts tool in [fronts-client](./fronts-client).
 
 1. Install [NVM](https://github.com/creationix/nvm).
 2. Ensure [Docker](https://www.docker.com/products/docker-desktop) is installed
-3. Get credentials from [Janus](https://janus.gutools.co.uk/multi-credentials?&permissionIds=cmsFronts-dev,capi-api-gateway) (`cmsFronts` & `capi`).
+3. Get credentials from [Janus](https://janus.gutools.co.uk/credentials?permissionId=cmsFronts-iam-facia-CODE-RunFaciaToolLocally-nL42h4zEgHhf).
```
waisingyiu, line 15: I suppose we need to have this new credential `cmsFronts-iam-facia-CODE-RunFaciaToolLocally-nL42h4zEgHhf` added to our account before this link works?

waisingyiu (approved):
It looks great to me with just a non-blocking question.

README.md
davidfurey: Yes, I should add it to the editorialToolsDeveloper permission set in Janus. Currently I am the only one with access to this permission

PR #1993, opened 2026-05-27
Show intended audience indicators

davidfurey:
## What's changed?

Shows intended audience indicator on articles in the CAPI Feed and on fronts.

Screenshot 2026-05-27 at 15 30 50

### DIFF
--- a/fronts-client/src/components/card/article/ArticleBody.tsx
+++ b/fronts-client/src/components/card/article/ArticleBody.tsx
@@ -53,6 +53,10 @@ import { Atom, Platform } from '../../../types/Capi';
 import { getActiveAtomProperties } from '../../../util/extractAtom';
 import { isAtom } from '../../../util/atom';
 import { SelfHostedVideoPlayerFormat } from '../../../util/Video';
+import {
+	IntendedAudienceSignifier,
+	IntendedAudienceSignifierProps,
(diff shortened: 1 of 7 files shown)


fronts-client/src/components/card/article/ArticleBody.tsx
```diff
@@ -181,6 +185,10 @@ interface ArticleBodyProps {
 	collectionType?: string;
 	groupIndex?: number;
 	otherCollectionsOnSameFrontThisCardIsOn?: OtherCollectionsOnSameFrontThisCardIsOn;
+	intendedAudience?: {
+		source: IntendedAudienceSignifierProps['source'];
+		target: IntendedAudienceSignifierProps['intendedAudience'];
+	};
```
rebecca-thompson, line 191: this intendedAudience type is declared multiple times. Is it worth declaring it once and referencing it?

rebecca-thompson (approved):
Nothing blocking, as this feature is not being used yet.

I did notice the audience indicator is masked by the hover actions

Image

Image


There has also been conversation about changing the modelling of the source / target to be lowercase to match the tracking tag paths more closely, and to use `au` instead of `AUD`, but that change can go in when the model has been updated in Stand.

davidfurey:
Works in prod:

Screenshot 2026-06-12 at 09 17 59

PR #1996, opened 2026-06-02
Make commonly failing fronts v1 tests more robust

davidfurey:
## What's changed?

Improved logging in tests and added more wait conditions so that (hopefully) tests are less flaky


davidfurey:
Superceeded by https://github.com/guardian/facia-tool/pull/1997

PR #1997, opened 2026-06-09
Resolve mulitple intermittently failing tests

davidfurey:
## What's changed?

The fronts v1 tests fail intermittently. This PR sets out to make the tests more reliable. Almost all of the changes in this PR are to tests. There is one production code change, changing where the isPasteActive observable is added. But this change seems very safe.

## Approach

I set up a loop running the tests and feeding the test output in Claude Opus when the tests failed, and allowing it to add additional logging.

## Results

Running `grunt karma:static` 1000 times:
- on [main](https://github.com/guardian/facia-tool/commit/3d237be2c9bc45532bf80b1afc4e5730bfce4590), tests failed 307 / 1000
- on this branch, tests failed 0 / 1000

## Issues

Claude identified 5 distinct failure signatures appear across 47 local logs. 

| Sig | Error | Test | Spec | Count |
|---|---|---|---|---|
| A | `Timeout waiting for condition` (wait.condition, 5000ms) | `create fronts and collections` | [config.front.spec.js](public/test/spec/config.front.spec.js#L88) | 24 |
| B | `edit action timeout, endpoint /edits` (base-action.js, 2000ms) | `Allows the same alert to be sent multiple times` (11) + `Alerts the user before launch` (6) | [breaking-news.spec.js](public/test/spec/breaking-news.spec.js) | 17 |
| C | `Cannot read properties of undefined (reading '$component')` | `searches articles` | [latest.articles.spec.js](public/test/spec/latest.articles.spec.js#L28) | 4 |
| D | `Expected 'Open Source' to be 'Nothing happened for once'` | `copy to clipboard` | [collections.spec.js](public/test/spec/collections.spec.js#L324) | 1 |
| E | `Expected 1 to be close to 0.6` | `marks an article visited from clipboard as visited…` | [visited.articles.spec.js](public/test/spec/visited.articles.spec.js#L31) | 1 |

### Sig A (Create Fronts and Collections - Timeout)

Originally suspected to be a slow image load, the root cause was actually an order-dependent global-singleton leak. The grid() utility memoized an instance on first construction, reading a base URL that was completely missing from the config.front.spec test model. If this spec ran first, it threw an error, skipped the save step, and timed out. It was fixed by adding the missing baseUrls: { apiBaseUrl, mediaBaseUrl } to the test model's defaults so that grid() initializes deterministically regardless of the Karma test execution order.

### Sig B (Breaking News Tests - Edit Action Timeout)

This failure had two sub-causes. First, the 2000ms timeout timer wrongly started at action construction instead of execution; this was fixed by moving the timer initiation into the execute() step. Second, the true root cause was that an asynchronous JSPM component loaded late on a cold cache, causing column separators to bind before an essential $root.isPasteActive observable existed, preventing a live subscription from forming. This was fixed by explicitly declaring this.isPasteActive = ko.observable(false) in the BaseModel constructor before any templates bind, ensuring a stable, real observable is always present to receive updates.

### Sig C (Searches Articles - $component Undefined)

The root cause was a race condition where the test resolved on the first widget:load event, which was emitted by all widgets after 20ms, often before the autocomplete widget actually rendered. It was fixed by rewriting getAutocomplete to bypass the generic event and instead poll via wait.condition, ensuring the autocomplete element exists and has a Knockout context before trying to read its $component. (A subsequent regression of this exact same bug pattern in the backfill region under Sig A was fixed by converting the backfill component into a lazy getter with a ready() guard).

### Sig D (Copy to Clipboard - Mismatched Assertion)

It was discovered that this was not a rendering race condition, but a state leak across tests. An asynchronous clipboard polling interval (saveInStorage) would occasionally outlive the test teardown and write stale items back to localStorage after it had been cleared. Because the clipboard already contained leaked items at the start of the test, the assertion pulled the oldest leaked headline instead of the newly copied one. This was fixed by clearing localStorage at the start of every Page constructor, guaranteeing a clean slate for each test.

### Sig E (Visited Articles - Opacity Mismatch)

This shared the exact same root cause as Sig D. An orphaned clipboard poller leaked a "snap" link item into localStorage from a previous test. When the test ran, the clipboard mistakenly interacted with this leaked snap item instead of the freshly copied article. Because the snap ID (snap/<timestamp>) didn't match the original ID format of the articles in the latest list, the "visited" state propagation failed to match any elements, leaving the opacity at 1. This was completely resolved by the same fix applied to Sig D—wiping localStorage at the start of every Page constructor to isolate test state.


Georges-GNM (approved):
Got a script to run the tests quite a few times and didn't get any errors - changes seem to make sense too, so looks good to me 👍 

Thanks again for this, always frustrating when a test intermittently fails!

PR #1999, opened 2026-06-17
Add templates for Feast US regional edition

fredex42:
## What's changed?

- Adds a "US Region" edtition for Feast app

## Implementation notes

- Added a new Edition to the tool to represent US Region, with the same templates as the existing regions
- In order to ensure that they publish correctly to the recipes-backend, it's necessary to explicitly add the edition names to the Publishing logic so they get directed to SNS rather than the Editions S3

To test correct publication see the steps at https://github.com/guardian/recipes-backend/pull/212

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/model/editions/templates/feast/FeastUSRegion.scala
+++ b/app/model/editions/templates/feast/FeastUSRegion.scala
@@ -1,0 +1,48 @@
+package model.editions.templates.feast
+
+import model.editions.templates.TemplateHelpers.{collection, front}
+import model.editions.{
+  CapiDateQueryParam,
+  CapiTimeWindowConfigInDays,
+  Daily,
+  Edition,
+  EditionTemplate,
+  FrontTemplate
+}
+
+import java.time.ZoneId
+
+object FeastUSRegion extends FeastAppEdition {
+  override val title: String = "Feast app [US region]"
+  override val edition = Edition.FeastUSRegion.entryName
+  override val locale = Some("en_US")
+  override val notificationUTCOffset: Int = 0
+
+  override val path: String = "us"
+
+  val MainFront: FrontTemplate = front(
+    "All Recipes",
+    collection("Dish of the day")
+  )
+
+  val MeatFreeFront: FrontTemplate = front(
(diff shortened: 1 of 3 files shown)


Divs-B:
We might need to drop "US Only" from the options

Divs-B:
Tiny bit: I see `Us` instead of `US`
image

app/model/editions/templates/feast/FeastUSRegion.scala
```diff
@@ -0,0 +1,48 @@
...
+    fronts = List(
+      MainFront -> Daily(),
+      MeatFreeFront -> Daily()
+    ),
+    timeWindowConfig = CapiTimeWindowConfigInDays(
+      startOffset = 0,
+      endOffset = 0
+    ),
+    capiDateQueryParam = CapiDateQueryParam.Published,
+    zoneId = ZoneId.of("America/New_York"),
```
Divs-B, line 43: Do we need to have selective timezone which is US eastern time zone

Divs-B (approved):
A few comments, I am happy to approve it as it has been tested on Fronts Code and published with the Fronts edition.

fredex42:
> We might need to drop "US Only" from the options

We definitely will. It's not in this PR because they'll want a period of parallel running; once the new front is running smoothly there's a card to removet he old options

app/model/editions/templates/feast/FeastUSRegion.scala
fredex42: Well, I looked this up - apparently `America/New_York` is the canonical TZ name as of the early 90s and the US Eastern timezone symlinks to `America/New_York` 🤷 .  Plus our US operations are based in NY, so it's as good a place as any to choose!

fredex42:
> Tiny bit: I see `Us` instead of `US` image

Yeah that is a bit annoying - something in the CSS I think. Not worth our worry now but might tweak it later.

PR #2000, opened 2026-06-24
Fix collection permissions check

SHession:
## What's changed?

As pointed out by @andrew-nowak, the Fronts tool currently has a bug in the permissions checks for various endpoints which means a user can access and edit collections beyond their permissions. This is due to the config value used by `CollectionPermissions` is set when the controllers are instantiated and before the config has been updated.  

<details><summary>Original message and detailed issue explanation</summary>
<p>

Unfortunately I'm fairly sure I've just discovered that the fronts tool isn't properly checking user's permissions. Despite not having permissions to either edit or launch editorial fronts on CODE, I've still been able to do so. Reproduction is fairly straight forward - remove your permissions (and wait a couple of minutes), then go to https://fronts.code.dev-gutools.co.uk/ and open the editorial priority fronts. (Going directly to https://fronts.code.dev-gutools.co.uk/v2/editorial, or reloading the page while having it open, does do a permissions check and locks me out, but the SPA navigation does not)

I've had a quick dig, and I'm fairly sure the problem is here: https://github.com/guardian/facia-tool/blob/d490441acfc79da933f34d088ce549b367ad5267/app/controllers/FaciaToolV2Controller.scala#L38. The CollectionPermissions class is constructed with the cached state of the config as it was when the instance was first launched, and never updated, but the ConfigAgent initialises its cache to None https://github.com/guardian/facia-tool/blob/d490441acfc79da933f34d088ce549b367ad5267/app/services/ConfigAgent.scala#L21. This means that when the permissions check is done, it falls into this case, which skips checking the user's permission https://github.com/guardian/facia-tool/blob/d490441acfc79da933f34d088ce549b367ad5267/app/permissions/CollectionPermissions.scala#L9.

I think a quick fix would be to change the CollectionPermissions class to take the ConfigAgent directly, and only call get inside the method when the permission check is being performed. This should mean that it always gets the latest state of the cached Config object, though it leaves the permissions checks to fail "open" if for some reason it holds a None at the time the user sends a request (in practice I think this will be unlikely, and if it did occur would probably suggest a failure to talk to S3, which would prevent their queries or updates from going through anyhow! but may still be worth further investigation)

</p>
</details> 

## Implementation notes

As with the existing behaviour, if a user attempts to navigate to locked views directly they be rejected: 

image

However, if they access the root view and navigate to the editorial page they will be able to see the collections albeit with limited controls and visibility: 

image

### Edge Case

As discussed with @abeddow91, there is still an edge case where if config is set to `None` the permissions required will be relaxed. We have added logging to check for this case and will raise a ticket to plug this gap. 

## Testing

To confirm this functionality works as expected, you can set your local / CODE permissions using the CODE permissions tool. Without edit permissions for front, you should not be able to access the page directly. Navigating via the SPA to collections should result in collections not loading with requests returning a 401.

### General
- [x] 🤖 Relevant tests added
- [x] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/controllers/FaciaToolController.scala
+++ b/app/controllers/FaciaToolController.scala
@@ -39,7 +39,7 @@ class FaciaToolController(
     with ModifyCollectionsPermissionsCheck
     with Logging {
 
-  private val collectionPermissions = CollectionPermissions(configAgent.get)
+  private val collectionPermissions = CollectionPermissions(configAgent)
 
   def getConfig = AccessAPIAuthAction.async { request =>
     FaciaToolMetrics.ApiUsageCount.increment()
--- a/app/controllers/FaciaToolV2Controller.scala
+++ b/app/controllers/FaciaToolV2Controller.scala
@@ -35,7 +35,7 @@ class FaciaToolV2Controller(
     with ModifyCollectionsPermissionsCheck
     with Logging {
 
-  private val collectionPermissions = CollectionPermissions(configAgent.get)
+  private val collectionPermissions = CollectionPermissions(configAgent)
 
   def getCollections() =
     AccessAPIAuthAction(parse.json[Seq[CollectionSpec]]).async {
--- a/app/permissions/CollectionPermissions.scala
+++ b/app/permissions/CollectionPermissions.scala
@@ -1,11 +1,12 @@
 package permissions
 
 import com.gu.facia.client.models.ConfigJson
+import services.ConfigAgent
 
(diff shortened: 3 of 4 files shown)


abeddow91 (commented):
Tested on CODE and working as described in the PR. Front access is possible, but  requests for the collections or to launch them will be rejected

abeddow91 (approved):

PR #2011, opened 2026-07-24
First sketch at adding AB headline fields

### LINKED ISSUE #28959: Fronts UI: 'Design AB Test Toggle'
Background:
We want to introduce an A/B test toggle to Fronts cards. This toggle should appear above the existing control headline field.

When enabled, an additional headline field should appear, allowing editors to enter a second headline variant for testing. Both headline fields should have a coloured box around them, to make it clear that the card is entering an AB test state.

When the toggle is turned off, the UI should return to its default state, with only the standard headline field visible and no A/B test styling applied.

Designs in Figma should be used as a guide. Any design changes, edge cases, or interaction amendments should be worked through with Chloe

Requirements
- Add an A/B test toggle above the existing control headline field.
- When the toggle is switched on:
- Show an additional headline field for the test headline.
- Apply visual styling around both headline fields to indicate that an A/B test is being created and active.
- Make it clear which headline is the control and which is the variant.

When the toggle is switched off:
- Hide the additional headline field.
- Remove any A/B test styling.
- Return the card to the existing/default headline editing state.
- Ensure the interaction follows the Figma design as closely as possible.
- Confirm any design changes or unclear interaction states with Chloe before build.

To consider:
Jump state: Consider how the UI behaves when the toggle is enabled or disabled. The layout should not feel jarring or cause the editor to lose context.
Multiple fronts at once: Consider how this looks when an article is appearing on multiple fronts or in multiple editable contexts. We need to avoid the UI becoming visually cluttered or confusing.

Fweddi:
Co-authored by: @emma-imber 

## What's changed?

Part of https://github.com/orgs/guardian/projects/147 – the mission to allow editors to AB test headlines.

This PR is behind a [feature switch](#2014) – to test, enable on /v2/features. 

Fulfils initial user stories, which are as a Fronts Editor, I want to:
🟢 choose a headline I want to AB test
🟢 enter & edit the text for headline A & B

image

image

### Notes
This just works on the client-side at the moment, and isn't wired up to facia-scala-client and S3. You can save the card and should be able to move it around and keep the headline fields and toggle.

Disabling the toggle does _not_ wipe the fields until the card is saved. I think this is fair, in case you misclick the toggle and want to restore the toggle and fields.


fronts-client/src/components/inputs/HeadlineInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+import InputTextArea from './InputTextArea';
+import InputCheckboxToggleInline from './InputCheckboxToggleInline';
+import ConditionalField from './ConditionalField';
+import styled from 'styled-components';
+import pageConfig from '../../util/extractConfigFromPage';
+
+interface HeadlineInputProps {
+	abTestEnabled: boolean;
+	capiHeadline: string;
+	cardId: string;
```
abeddow91, line 13: Is `cardId` just needed here to create a id on the field component?

fronts-client/src/components/inputs/InputCheckboxToggleInline.tsx
```diff
@@ -88,6 +110,9 @@ type Props = {
 	label?: string;
 	id: string;
 	dataTestId?: string;
+	useABTestStyling?: boolean;
+	activeABTest?: boolean;
+	theme: any;
```
abeddow91: I dont think this theme prop is used, it looks like any instances of `theme` used seem to be fromthe theme import on line 7

fronts-client/src/components/inputs/HeadlineInput.tsx
Fweddi: Yes, in line with other Field components of type checkbox – the id is used to yoke together the Checkbox and the Label (see InputCheckboxToggleInline).

fronts-client/src/components/inputs/InputCheckboxToggleInline.tsx
```diff
@@ -5,27 +5,43 @@ import { WrappedFieldMetaProps, WrappedFieldInputProps } from 'redux-form';
...
-	padding-left: 5px;
-	color: ${(props) => props.theme.input.colorLabel};
+const Label = styled(InputLabel)<{
+	useABTestStyling?: boolean;
+	activeABTest?: boolean;
+}>`
+	color: ${(props) =>
+		props.useABTestStyling && props.activeABTest
+			? props.theme.input.abTestColorActive
+			: props.theme.input.colorLabel};
```
abeddow91: We seem to use destructuring elsewhere, could we do that here too to be consistent? Similarly, could `theme` come from the imported theme, which seems in keeping with other theme usages?

```suggestion
	color: ${({useABTestStyling, activeABTest}) =>
		useABTestStyling && activeABTest
			? theme.input.abTestColorActive
			: theme.input.colorLabel};
```

fronts-client/src/constants/theme.ts
```diff
@@ -130,6 +134,8 @@ const input = {
 	placeholderText: base.colors.placeholderDark,
 	radioButtonBackgroundDisabled: '#E6E6E6',
 	checkboxButtonBackgroundDisabled: '#E6E6E6',
+	abTestColorActive: base.colors.abTestActiveColor,
```
abeddow91: Would it be possible to use the same naming convention for both, so either `abTestColorActive` or `abTestActiveColor`?

fronts-client/src/components/inputs/CreateResizeableTextInput.tsx
```diff
@@ -17,6 +17,8 @@ type Props = {
 	>;
 	labelContent?: React.Component<{}>;
 	type?: string;
+	isHeadline?: boolean;
+	cardId?: string;
```
abeddow91: These don't seem to be used by this component - can they be removed?

fronts-client/src/components/inputs/HeadlineInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+	const headlineABTestingFeatureSwitch =
+		pageConfig?.userData?.featureSwitches.find(
+			(feature) => feature.key === 'headline-ab-testing',
+		);
+
+	return (
+		<HeadlineInputContainer
+			abTestEnabled={
+				props.abTestEnabled && headlineABTestingFeatureSwitch?.enabled === true
```
abeddow91, line 62: We make this check 3 times. Could it be lifted up into a single const eg
`	const isABTestEnabled = props.abTestEnabled && headlineABTestingFeatureSwitch?.enabled === true;
`

fronts-client/src/components/inputs/CreateResizeableTextInput.tsx
emma-imber: Good spot!

fronts-client/src/constants/theme.ts
emma-imber: Updated

fronts-client/src/components/inputs/InputCheckboxToggleInline.tsx
emma-imber: Nice spot

abeddow91 (approved):
Tested on CODE. The additional headline inputs fields are added as expected on toggle on and cleared on toggle off once the card is saved. 

Also, edits to the original headline also seem to persist throughout toggling ab test on and off.

In short, all looks good to me! 🥳

fronts-client/src/components/inputs/HeadlineInput.tsx
Fweddi: I think given the feature switch is temporary we're probably okay with this for now?

PR #2017, opened 2026-07-30
Add test status labels to collapsed cards

### LINKED ISSUE #28960: Fronts UI: Add Test Tube Indicator Collapsed card view
Background:
Once an editorial A/B test has been configured, Fronts editors should be able to easily identify which cards are in a test without having to open the card.

The collapsed card should display the current status of the test, allowing editors to distinguish between tests that are configured but not yet live, and tests that are actively running.

In addition, the Front overview should provide a lightweight visual indicator that a container contains one or more active A/B tests.

Requirements

There are two test states to support:
**1. Pre-launch**

When an editor has configured and saved an A/B test, but the test has not yet been launched, the collapsed card should display a status message similar to:

Ready to launch – [headline/variant summary]

(Final copy to be agreed with Jonathan.)

**2. Test Running**

Once the A/B test has been launched, the collapsed card should instead display:

Test in progress – [headline/variant summary]

(Final copy to be agreed with Jonathan.)

The status should update automatically based on the underlying test state.

**Front Overview**

If a container has one or more active A/B tests, display a test tube icon within the Front overview to provide an at-a-glance indication that experimentation is taking place within that container.

- Designs to be followed from Figma but not conrete and may change (work in collab with Chloe)

### LINKED ISSUE #29014: Fronts UI: Add 'Ready to launch' before test has been launched
Image

We want to add a ready to launch message on the card when it has been saved but not launched

emma-imber:
Co-authored by @Fweddi 

## What's changed?
Adds a visual indication of the current status of an AB test on collapsed cards.

If a test is live and still enabled in draft → `Test in progress`
If a test is live but has been switched off in draft → `End test on launch`
If a test has been added in draft but is not yet live →`Ready to launch`

Nothing is shown for cards with no draft or live AB test

https://github.com/user-attachments/assets/305017e2-ea52-44a2-942d-acc60581cc8e


fronts-client/src/components/card/article/ArticleBody.tsx
```diff
@@ -268,6 +306,21 @@ const articleBodyDefault = React.memo(
...
+				return 'Test in progress';
+			}
+			if (hasLiveAbTest && !abTestEnabled) {
+				return 'End test on launch';
+			}
+			if (!hasLiveAbTest && abTestEnabled) {
+				return 'Ready to launch';
+			}
+		};
+
```
abeddow91, line 329: Do we need to handle when both are undefined? perhaps we could early exit

abeddow91 (approved):

fronts-client/src/components/card/article/ArticleBody.tsx
Fweddi: If both parameters are falsy, it will implicitly return undefined – but for legibility let's make this explicit.

PR #2030, opened 2026-08-07
Populate frontsThisTestCanRunOn property when test is created

### LINKED ISSUE #29037: Populate frontsThisTestCanRunOn on test creation
Create a function that takes a collectionId and returns all the front ids that that collection belongs to.

simonbyford:
## What's changed?
Adds a new selector `selectFrontsWithCollection`, which selects any fronts with a given collection ID. This is used to populate the value of the `frontsThisTestCanRunOn` property when a new headline test is created.

Testing locally you can see this value being populated:

Screenshot 2026-08-10 at 12 43 16

### DIFF
--- a/fronts-client/src/selectors/frontsSelectors.ts
+++ b/fronts-client/src/selectors/frontsSelectors.ts
@@ -18,6 +18,7 @@ import {
 	selectFrontId,
 } from 'selectors/shared';
 import { createShallowEqualResultSelector } from 'util/selectorUtils';
+import { selectors as collectionSelectors } from 'bundles/collectionsBundle';
 
 interface FrontConfigMap {
 	[id: string]: FrontConfig;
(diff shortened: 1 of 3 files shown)


fronts-client/src/selectors/frontsSelectors.ts
```diff
@@ -256,6 +270,11 @@ const selectHasUnpublishedChanges = createSelector(
 	getUnpublishedChangesStatus,
 );
 
+const selectFrontsWithCollection = createSelector(
+	[collectionSelectors.selectParentCollectionOfCard, selectFrontsAsArray],
+	getFrontsWithCollection,
+);
```
abeddow91, line 276: createSelector only memoises the returned function, `getFrontsWithCollection` in this case. That would be fine but as `selectParentCollectionOfCard` is not being memoised, I think this could prove to be an expensive selector to use, as 
`selectParentCollectionOfCard` will traverse the entire collections/groups/cards tree on every call of `selectFrontsWithCollection`. 

This is non-blocking for this PR as this selector already exists but I think it should be refactored before this feature is rolled out to production. I've had a look at it [here](https://github.com/guardian/facia-tool/pull/2041) with the help of Copilot. IMO, this refactor makes the `selectParentCollectionOfCard` not only more performant but much more readable.

abeddow91 (approved):

PR #2033, opened 2026-08-07
Set variant headlines to undefined when empty

### LINKED ISSUE #29039: Prevent headlineA field being populated before a test has been created

Fweddi:
This PR:
* Sets variant headline fields to undefined when empty (in the same way we handle the OG headline field)
* Populates headlineA field with the OG headline field value when the toggle is switched on

I opted for this approach over initialising headlineA with the OG headline field, because then you would still need to react to any OG headline field changes and mutate headlineA accordingly.

If you untoggle the test and retoggle it on, you will lose your persisted headlineA field in favour of the current value of the OG headline field. I think this is okay in practice but welcome to discuss, I can't imagine it will be a common user story.

I've kept the logic in selectors/shared.ts (`// if headlineA is not present, populate it with the headline field`) in case the headlineA field is still empty for some reason, but we could now get rid of this logic and rely on it being set correctly at the point the toggle switch is used.

### DIFF
--- a/fronts-client/src/components/form/__tests__/ArticleMetaForm.spec.tsx
+++ b/fronts-client/src/components/form/__tests__/ArticleMetaForm.spec.tsx
@@ -1,0 +1,64 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import { fireEvent, render, cleanup } from 'react-testing-library';
+import { ThemeProvider } from 'styled-components';
+import { theme } from 'constants/theme';
+import configureStore from 'util/configureStore';
+import { state as initialState } from 'fixtures/initialState';
+import ArticleMetaForm from '../ArticleMetaForm';
+
+// TODO: Remove when no longer gated behind feature switch
+jest.mock('util/extractConfigFromPage', () => {
+	const baseConfig = jest.requireActual('fixtures/config').default;
+	return {
+		__esModule: true,
+		default: {
+			...baseConfig,
+			userData: {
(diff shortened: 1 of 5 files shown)


fronts-client/src/components/form/__tests__/ArticleMetaForm.spec.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import { fireEvent, render, cleanup } from 'react-testing-library';
+import { ThemeProvider } from 'styled-components';
+import { theme } from 'constants/theme';
+import configureStore from 'util/configureStore';
+import { state as initialState } from 'fixtures/initialState';
+import ArticleMetaForm from '../ArticleMetaForm';
+
+// TODO: Remove when no longer gated behind feature switch
```
abeddow91, line 10: Could you ticket this up please so we don't forget

fronts-client/src/components/form/__tests__/ArticleMetaForm.spec.tsx
abeddow91, line 1: Thanks for adding this coverage 🙏

abeddow91 (approved):

fronts-client/src/components/form/__tests__/ArticleMetaForm.spec.tsx
Fweddi: Good idea!

PR #2036, opened 2026-08-10
Hide test status icons when headline testing feature switch is off

### LINKED ISSUE #29042: Users who do not have feature switch enabled should not see `test` details on cards

emma-imber:
## What's changed?
Hides test status icons when the headline AB testing feature switch is off. At the moment the test status icons will show to all users regardless of whether the feature switch is on or not. We want to only show these icons when a user has switched on headline AB testing.

## Screenshots

Collection overview:

| Before      | After     |
| ------------- | ------------- |
| Screenshot 2026-08-10 at 16 06 15 | Screenshot 2026-08-10 at 16 12 29 |
| Screenshot 2026-08-10 at 17 22 19 | Screenshot 2026-08-10 at 17 22 39 |

### DIFF
--- a/fronts-client/src/components/FrontsEdit/CollectionOverview.tsx
+++ b/fronts-client/src/components/FrontsEdit/CollectionOverview.tsx
@@ -37,7 +37,7 @@ type FrontCollectionOverviewProps = FrontCollectionOverviewContainerProps & {
 	hasUnpublishedChanges: boolean;
 	hasOpenForms: boolean;
 	liveAndDraftCards: Card[];
-	headlineABTestingIsEnabled: boolean;
+	headlineABTestingIsEnabled?: boolean;
 };
 
 const Container = styled.div<{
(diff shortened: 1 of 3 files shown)


simonbyford (approved):
Makes sense 👍

fronts-client/src/components/FrontsEdit/CollectionOverview.tsx
```diff
@@ -238,6 +243,10 @@ const mapStateToProps = () => {
 				collectionId,
 			) !== -1,
 		liveAndDraftCards: selectLiveAndDraftCards(state, collectionId),
+		headlineABTestingIsEnabled: selectFeatureValue(
+			state,
+			'headline-ab-testing',
+		),
```
Fweddi, line 249: this is different to how we're using the AB testing switch elsewhere (eg in HeadlineInput:

```
const headlineABTestingFeatureSwitch =
		pageConfig?.userData?.featureSwitches.find(
			(feature) => feature.key === 'headline-ab-testing',
		);
```

but this is also a common pattern, temporary, and works, so I think it's okay to be inconsistent – just need to remember to clean both types up on release!

Fweddi (approved):

PR #2041, opened 2026-08-11
Refactor selectParentCollectionOfCard selector

abeddow91:
## What's changed?
This PR memoises `selectParentCollectionOfCard` with a reverse index.

Previously, `selectParentCollectionOfCard` traversed the entire collections/groups/cards tree on every call. Since Reselect invokes input selectors whenever a selector is called, this traversal ran each time `selectFrontsWithCollection` was evaluated, as well as once per card in the `storeMiddleware` batch loop. This meant repeatedly scanning the entire tree, once for every card being processed.

This change introduces `selectCardToCollectionMap`, a memoised selector that builds a cardId -> collectionId lookup, including supporting cards. The map is rebuilt only when the collections, groups, or cards slices change.

`selectParentCollectionOfCard` can now find a card’s parent collection with a single map lookup., avoiding repeated full-tree scans and allowing downstream selector memoisation to be effective.

The existing first-match-wins behaviour is preserved, so the returned collection IDs remain unchanged. Optional-chaining guards have also been added when accessing groups and supporting cards, preventing a latent runtime error caused by dangling group IDs.

## Implementation notes
Initial pass of this refactor was done by Copilot 🤖 and then refined by me 👋

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/bundles/collectionsBundle.ts
+++ b/fronts-client/src/bundles/collectionsBundle.ts
@@ -43,8 +43,6 @@ const getGroupIdsForCollection = (
  * and cards slices so the full tree is only traversed when one of them changes,
  * rather than on every `selectParentCollectionOfCard` call.
  *
- * A card can appear in more than one collection; we keep the first match to
- * preserve the previous selector's behaviour.
  */
 const selectCardToCollectionMap = createSelector(
 	(state: State) => state.collections.data,
--- a/fronts-client/src/bundles/collectionsBundle.ts
+++ b/fronts-client/src/bundles/collectionsBundle.ts
@@ -7,48 +7,71 @@ import { Collection } from 'types/Collection';
 import { State } from 'types/State';
 import { addPersistMetaToAction } from '../util/action';
 import set from 'lodash/fp/set';
+import { createSelector } from 'reselect';
 
 const collectionsEntityName = 'collections';
 
 const { actions, actionNames, reducer, selectors, initialState } =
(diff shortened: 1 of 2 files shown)


fronts-client/src/bundles/collectionsBundle.ts
```diff
@@ -7,48 +7,69 @@ import { Collection } from 'types/Collection';
...
-						return false;
-					});
-				});
-			}),
-		);
-		return collectionId;
-	},
+	selectCardToCollectionMap,
+	selectParentCollectionOfCard: (state: State, cardId: string): string | null =>
+		selectCardToCollectionMap(state)[cardId] ?? null,
```
Fweddi, line 72: I assume the cardId is the UUID here and so there will be only one parent collection (if it was the article ID, and the card was on multiple collections, this method could only return one collection).

Fweddi (approved):
LGTM, added a comment about the UUID. Could maybe do with a test?

fronts-client/src/bundles/collectionsBundle.ts
abeddow91: Yep, its being called with the UUID of the card in all instances

abeddow91:
> LGTM, added a comment about the UUID. Could maybe do with a test?

Good point, I've updated the existing test suite and added some additional ones

Fweddi (approved):

PR #2043, opened 2026-08-12
Don't set `startDate` and `expiryDate` if test has been manually ended

### LINKED ISSUE #29002: Fronts UI: End a test and start another
Adds a new test object to the card when the toggle switch is re-enabled after a test is manually ended (or expired). This test should have a different UUID.

simonbyford:
## What's changed?

When a container is published, `addTestDatesToTrail` sets `startDate`/`expiryDate` for every AB test that didn't include them, including tests that had been manually ended. This gave stopped tests a live-looking date window which is probably harmless but confusing.

This PR fixes that bug and adds a test.

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/app/tools/FaciaApiIO.scala
+++ b/app/tools/FaciaApiIO.scala
@@ -165,9 +165,12 @@ object FaciaApi {
 
   private def addTestDatesToTrail(trail: Trail, now: DateTime): Trail = {
     val updatedTests = trail.tests.map(_.map { test =>
-      val startDate = test.startDate.orElse(Some(now.getMillis))
-      val expiryDate = test.expiryDate.orElse(Some(now.plusDays(1).getMillis))
-      test.copy(startDate = startDate, expiryDate = expiryDate)
+      if (test.hasManuallyEndedOnThisTrail) test
+      else {
+        val startDate = test.startDate.orElse(Some(now.getMillis))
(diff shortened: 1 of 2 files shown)


app/tools/FaciaApiIO.scala
```diff
@@ -165,9 +165,12 @@ object FaciaApi {
 
   private def addTestDatesToTrail(trail: Trail, now: DateTime): Trail = {
     val updatedTests = trail.tests.map(_.map { test =>
-      val startDate = test.startDate.orElse(Some(now.getMillis))
-      val expiryDate = test.expiryDate.orElse(Some(now.plusDays(1).getMillis))
-      test.copy(startDate = startDate, expiryDate = expiryDate)
+      if (test.hasManuallyEndedOnThisTrail) test
```
Fweddi, line 168: I think this makes sense as there is a user story where you enable the test toggle, but then disable it (eg you've made a typo, or changed your mind). This would create a test in draft but we wouldn't want to set its start and expiry dates when the container is launched.

We could also choose to check for if a test has expired here but that would be quite the edge case – expiry date in the past, but no start date, which shouldn't be possible!

Fweddi (approved):

app/tools/FaciaApiIO.scala
simonbyford: Agreed!

PR #2049, opened 2026-08-14
End AB test when a card is moved to a front it can't run on

### LINKED ISSUE #29021: Fronts: Logic to decide when to end a test (when a card is moved off a front)
When a card with an active test is moved to a Front (which is not a valid ID on `frontsThisTestCanRunOn`), we should set `hasManuallyEndedToTrue`.

simonbyford:
## What's changed?

AB tests are configured to run only on the front(s) they were created on (recorded in `frontsThisTestCanRunOn`). Previously, moving a card containing a running test to a different front left the test running there. This PR adds a `mayEndAbTest` action that ends the test on that trail (setting `hasManuallyEndedOnThisTrail = true` and the acting user's name/email) whenever a card lands on a front whose ID isn't in `frontsThisTestCanRunOn`.

`mayEndAbTest` is run alongside the other `mayReset*` card-modification actions in `moveCard`. A new `selectFrontIdsWithCollectionId` selector resolves the destination collection to its host front IDs so we can compare against the test's allowed fronts.

Behaviour covered:

🟢 Moving a card to another front (e.g. US → UK) ends the test.
🟢 Moving a card around within its own front keeps it running.
🟢 Moving a card to the clipboard and re-adding it to an allowed front keeps it running (the clipboard move itself is a no-op).
🟢 Moving a card to the clipboard and not re-adding it ends it (its absence from the front implies the test no longer runs).

Tests with no `frontsThisTestCanRunOn`, already-ended tests, and expired tests are left untouched.

https://github.com/user-attachments/assets/a25b2e04-7cb7-426c-804c-2d21e04efef8

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/components/FrontsEdit/__tests__/mayEndAbTests.spec.tsx
+++ b/fronts-client/src/components/FrontsEdit/__tests__/mayEndAbTests.spec.tsx
@@ -1,0 +1,157 @@
+import type { Card, Test } from 'types/Collection';
+
+import { mayEndAbTest } from '../../../actions/Cards';
+import { baseTo, baseFrom } from './fixtures/groups.fixture';
+
+const FUTURE = Date.now() + 100_000;
+const PAST = Date.now() - 100_000;
+
+const UK_COLLECTION = 'collection-uk';
+const US_COLLECTION = 'collection-us';
+
+const makeTest = (overrides: Partial<Test> = {}): Test => ({
+	testUuid: 'test-1',
+	variantMeta: [],
+	createdByName: 'Jane Doe',
+	createdByEmail: 'jane.doe@guardian.co.uk',
+	expiryDate: FUTURE,
+	frontsThisTestCanRunOn: ['uk'],
+	hasManuallyEndedOnThisTrail: false,
+	...overrides,
+});
+
+const makeCard = (...tests: Test[]): Card => ({
+	id: 'internal-code/page/15334368',
+	frontPublicationDate: 1741879217277,
+	meta: {},
+	uuid: 'card-1',
+	tests: tests.length ? tests : undefined,
+});
+
+// The test was created on the `uk` front, which hosts UK_COLLECTION.
+const state: any = {
+	fronts: {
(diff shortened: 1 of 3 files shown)


fronts-client/src/components/FrontsEdit/__tests__/mayEndAbTests.spec.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+const UK_COLLECTION = 'collection-uk';
+const US_COLLECTION = 'collection-us';
+
+const makeTest = (overrides: Partial<Test> = {}): Test => ({
+	testUuid: 'test-1',
+	variantMeta: [],
+	createdByName: 'Jane Doe',
+	createdByEmail: 'jane.doe@guardian.co.uk',
+	expiryDate: FUTURE,
```
abeddow91, line 17: Nice, v readable

fronts-client/src/selectors/frontsSelectors.ts
```diff
@@ -120,6 +120,12 @@ const getFrontsWithCollection = (
 	return frontsWithCollection.map((front) => front.id);
 };
 
+const selectFrontIdsWithCollectionId = (
+	state: State,
+	collectionId: string | undefined,
+): string[] =>
+	getFrontsWithCollection(collectionId ?? null, selectFrontsAsArray(state));
```
abeddow91: Non-blocking : 
Does this need `?? null`? An undefined `collectionId` should get caught by the falsey check in `getFrontsWithCollection`. 

If `collectionId` can only be string or undefined, perhaps the types should be aligned to a allow `string | undefined | null` rather than coercing the value to null.

abeddow91 (approved):

fronts-client/src/selectors/frontsSelectors.ts
simonbyford: Good point, I've made that change

PR #2051, opened 2026-08-19
Prevent collection launch if test error exists

### LINKED ISSUE #29026: Fronts UI: Add Error State Messaging
🟡 get an error if I try to launch a test with the same headline
🟡 get an error if I try to launch a test with only one headline

abeddow91:
## What's changed?
Prevents a user from launching a collection if there is an incomplete test saved on a card. 

A test is considered incomplete if it meets one of the following criteria:
1. Duplicate - both headline variants are matching
2. Incomplete - at least one headline is blank.

If one of these conditions are met, the launch collection button is disabled and a warning is displayed on hover.

This uses the same underlying component as the open form warning. To facilitate this, this PR refactors the openFormWarning to create a reusable presentational component that can be used with a thin wrapper for both OpenFormsWarning and ABTestHeadlineWarning. I've used this opportunity to modernise the syntax for these components whilst refactoring.

## Screenshot

Screenshot 2026-08-19 at 13 56 08

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

### DIFF
--- a/fronts-client/src/util/abTests.ts
+++ b/fronts-client/src/util/abTests.ts
@@ -14,11 +14,41 @@ export const hasActiveAbTestOnCard = (card: Card | undefined) =>
 export const findActiveOrDraftTest = (card: Card) =>
 	card.tests?.find((test) => isActiveTest(test) || isDraftTest(test));
 
+export type AbTestHeadlineErrorType = 'duplicate' | 'incomplete';
+
+/**
+ * Validates the active headline A/B test (if any) on a card. Returns:
+ *  - 'incomplete' when either variant headline is missing/empty
+ *  - 'duplicate' when both variant headlines are identical
+ *  - null when there is no active test or the test is valid
+ */
+export const getActiveAbTestHeadlineError = (
+	card: Card,
+): AbTestHeadlineErrorType | null => {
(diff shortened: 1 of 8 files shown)


fronts-client/src/util/abTests.ts
```diff
@@ -14,11 +14,42 @@ export const hasActiveAbTestOnCard = (card: Card | undefined) =>
...
+	card: Card,
+): AbTestHeadlineErrorType | null => {
+	const currentTest =
+		card.tests?.find(isActiveTest) || card.tests?.find(isDraftTest);
+	if (!currentTest) {
+		return null;
+	}
+
+	const headlineA = getVariantHeadline(currentTest, 'A')?.trim();
+	const headlineB = getVariantHeadline(currentTest, 'B')?.trim();
```
emma-imber: Not essential, but it might be nice to convert the strings to all lower or upper case? At the moment two identical headlines which only differ by capitalisation would pass the check

emma-imber (approved):

fronts-client/src/util/abTests.ts
abeddow91: Good point. I'll add this in

PR #2057, opened 2026-08-21
Set manuallyEndedOnThisTrailDate when a test is manually ended

### LINKED ISSUE #29099: Add a hasManuallyEnded timestamp field to the model
We believe it will be valuable to know when a test has been manually ended. 

To know this, we will need to add an optional hasManuallyEnded timestamp when a test is manually turned off. 

Tasks

- [ ] Update FSC model to include new timestamp field
- [ ] Bump FSC in facia tool and set timestamp when hasManuallyEnded is set
- [ ] Bump FSC in frontend and ensure this is dropped before the test payload is sent to DCR
- [ ] No changes required in DCR as it does not need to care about this field.

emma-imber:
## What's changed?
Bumps the model to bring in the `manuallyEndedOnThisTrailDate` field, and sets a value for that field when a test is manually ended on a card, and the card is launched.

Screenshot 2026-08-21 at 16 21 58

### DIFF
--- a/app/tools/FaciaApiIO.scala
+++ b/app/tools/FaciaApiIO.scala
@@ -166,11 +166,9 @@ object FaciaApi {
   private def addTestDatesToTrail(trail: Trail, now: DateTime): Trail = {
     val updatedTests = trail.tests.map(_.map { test =>
       if (test.hasManuallyEndedOnThisTrail) {
-        test.manuallyEndedOnThisTrailDate match {
-          case Some(endedDate) => test
-          case _ =>
-            test.copy(manuallyEndedOnThisTrailDate = Some(now.getMillis))
-        }
+        test.copy(manuallyEndedOnThisTrailDate =
+          test.manuallyEndedOnThisTrailDate.orElse(Some(now.getMillis))
+        )
       } else {
         val startDate = test.startDate.orElse(Some(now.getMillis))
         val expiryDate = test.expiryDate.orElse(Some(now.plusDays(1).getMillis))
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -84,7 +84,7 @@ libraryDependencies ++= Seq(
   "com.gu" %% "content-api-client-aws" % "0.7.6",
   "com.gu" %% "content-api-client-default" % capiClientVersion,
   "com.gu" %% "editorial-permissions-client" % "3.0.0",
-  "com.gu" %% "fapi-client-play30" % "37.0.0",
+  "com.gu" %% "fapi-client-play30" % "38.0.0-PREVIEW.eiadd-manually-ended-date.2026-08-21T1150.d5613722",
```
emma-imber: This will need to be bumped to a non-preview version before merging

abeddow91 (approved):

app/tools/FaciaApiIO.scala
```diff
@@ -165,8 +165,13 @@ object FaciaApi {
...
   private def addTestDatesToTrail(trail: Trail, now: DateTime): Trail = {
     val updatedTests = trail.tests.map(_.map { test =>
-      if (test.hasManuallyEndedOnThisTrail) test
-      else {
+      if (test.hasManuallyEndedOnThisTrail) {
+        test.manuallyEndedOnThisTrailDate match {
+          case Some(endedDate) => test
+          case _ =>
+            test.copy(manuallyEndedOnThisTrailDate = Some(now.getMillis))
+        }
```
simonbyford: Non-blocking but I think this would do the same thing:

```suggestion
        test.copy(manuallyEndedOnThisTrailDate =
          test.manuallyEndedOnThisTrailDate.orElse(Some(now.getMillis))
        )
```

simonbyford (approved):

app/tools/FaciaApiIO.scala
emma-imber: Ah that's a little neater, thanks!

PR #2059, opened 2026-08-25
Only allow headline ab testing on the US network front

abeddow91:
## What's changed?
Adds a guard to headline ab testing functionality so that it is only available on the US front. This is for the initial rollout only so that we can limit behaviour to a smaller subset of users. 

Removal of this restriction will be covered by https://github.com/guardian/frontend/issues/29129, once initial testing has finished. 

This should not be merged until we are ready to rollout to production users.

### General
- [ ] 🤖 Relevant tests added
- [ ] ✅ CI checks / tests run locally
- [ ] 🔍 Checked on CODE

### Client
- [ ] 🚫 No obvious console errors on the client (i.e. React dev mode errors)
- [ ] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work)
- [ ] 📷 Screenshots / GIFs of relevant UI changes included

PR #2060, opened 2026-09-01
Lift front restriction so AB headline tests can run on all fronts

### LINKED ISSUE #29129: Fronts tool: Lift restrictions so that all fronts in Fronts Tool can have tests
A restriction has been put in place in DCR and the Fronts tool so that headline testing only appears on the US front during the initial window. 

This ticket is to remove this restriction to enable a full rollout once testing is complete.

Fweddi:
## What's changed?
Effectively reverts #2059. To be rolled out after we have (successfully) tested on the US front.

### DIFF
--- a/fronts-client/src/components/FrontsEdit/CollectionComponents/Collection.tsx
+++ b/fronts-client/src/components/FrontsEdit/CollectionComponents/Collection.tsx
@@ -233,8 +233,6 @@ class Collection extends React.Component<CollectionProps, CollectionState> {
 
 		const groupIds = groups.map((group) => group.uuid);
 
-		const isUSNetworkFront = frontId === 'us';
-
 		return (
 			<>
 				<CollectionDisplay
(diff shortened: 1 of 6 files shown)


emma-imber (commented):
I've tested locally and I am able to add a test on a non US front, but none of the test status indicator icons are appearing. The code in `ArticleBody` and in `CollectionOverview` will also need to be updated so that we can see the card and collection test status indicators on non US fronts

Fweddi:
> I've tested locally and I am able to add a test on a non US front, but none of the test status indicator icons are appearing. The code in `ArticleBody` and in `CollectionOverview` will also need to be updated so that we can see the card and collection test status indicators on non US fronts

Thanks for spotting! It appears Anna wrote some further commits in https://github.com/guardian/facia-tool/pull/2059 that I hadn't picked up – those should now be actioned on.

emma-imber (approved):
Looks good!

PR #2045, opened 2026-08-13
GuCDK migration phase 1: wrap facia CloudFormation with @guardian/cdk

davidfurey:
## What's changed?

**Phase 1 of the CloudFormation → GuCDK migration** for facia-tool: wrap the existing CloudFormation template with `@guardian/cdk` via `CfnInclude`, and take over deployment of the `facia` stack from `editorial-tools-platform`.

- Adds `cdk/` — a GuCDK app that wraps `cloudformation/facia-tool.cfn.yaml` (moved here from the platform repo) with `CfnInclude`. Stack ids `FaciaTool-euwest-1-CODE` / `FaciaTool-euwest-1-PROD`, each with a per-stage `cloudFormationStackName` (`facia-CODE` / `facia-PROD`) matching the live stacks.
- `riff-raff.yaml`: replaces the legacy `ami-cloudformation-parameter` deploy with a `cloud-formation` deploy (`cfn-facia-tool`) of the synthesised templates, still targeting the `facia` stack; `autoscaling` now depends on it.
- CI (`improvedci.yml`): builds + synths the CDK app (node pinned via `cdk/.nvmrc`) and uploads the templates to Riff-Raff.
- `.gitignore`: negate `/*/bin/` for `cdk/bin/` so the CDK entry point is tracked. `.prettierignore`: exclude `cdk/` (the CDK project owns its own formatting).

This is a **tags-only** change — verified with `cdk diff` against the live `facia-CODE` stack: additive `gu:cdk:version` / `gu:repo` / `Stack` / `Stage` / `App` tags and `gu:cdk` metadata only, plus benign representational changes (`Stage` `{Ref}`→literal, ASG `PropagateAtLaunch` `"true"`→`true`, `gu:riffraff:new-asg` `1`→`"1"`) and the known cdk-diff SG-ingress remove/re-add presentation quirk. No resource creations or replacements.

This is a **mixed** template (CloudFront ×2, DNS records, DynamoDB, SNS, cross-account IAM alongside the ELB/ASG compute), so later phases keep the `CfnInclude`; the end-state is `CDK(cfn.yaml) -> cfn.json`.

## Implementation notes

- Front door is CloudFront (origin = the classic ELB), so the eventual traffic cutover (Phase 3) will be a CloudFront origin repoint, not a DNS switch.
- The legacy ASG already carries `gu:riffraff:new-asg` — it will be removed when the new ASG is introduced in Phase 2.
- `plans/cloudformation-to-cdk-migration.md` is the living migration plan.

[...]

PR #2061, opened 2026-09-01
GuCDK migration phase 2: run GuEc2App alongside the legacy ELB

davidfurey:
> [!NOTE] 

## What's changed?

Phase 2 of the [CloudFormation → GuCDK migration](https://github.com/guardian/cdk/blob/main/docs/migration-guide.md): a `GuEc2App` (ALB + ASG) is introduced **in parallel** with the legacy classic ELB, so the two run side by side in the same `facia-CODE` / `facia-PROD` stacks.

**No traffic moves in this PR.** CloudFront still points at the legacy ELB; repointing the origin at the new ALB is phase 3 and is independently revertible.

Follows on from #2045 (phase 1 — wrapping the template in `CfnInclude`), which is deployed to PROD. The living plan is in [`plans/cloudformation-to-cdk-migration.md`](plans/cloudformation-to-cdk-migration.md).

Also in this PR:

- The leftover `gu:riffraff:new-asg` tag is removed from the **legacy** ASG (it was a stale marker from an earlier migration) and added to the **new** ASG, which is what the tag is for.
- `riff-raff.yaml` gains `asgMigrationInProgress: true` so Riff-Raff rotates both ASGs instead of failing on finding two, and switches from `amiParameter` to `amiParametersToTags` so both `AMI` (legacy launch config) and `AMIFaciatool` (new launch template) are resolved at deploy time.

## Implementation notes

**Everything already in the wrapped template is referenced, not retyped.** The VPC, subnets, DB and CAPI security groups, `FrontendRoleToAssume`, `CapiPreviewRole`, `SwitchboardBucket` and `UserDataTablePrefix` all come from `cfnInclude.getParameter(...)`; the DynamoDB table names come from the existing `CrossResources` / `StageMap` mappings via `Fn.findInMap`. Reusing the network parameters also avoids GuCDK's generated `VpcId` / `PrivateSubnets` / `PublicSubnets` parameters clashing with the template's own.

Per-stage values (`domainName`, `instanceType`, min/max instances) live in `bin/cdk.ts` next to the existing per-stage `cloudFormationStackName`, rather than in a stage lookup inside the stack class.

**`GuUserData` isn't usable here**, so the user data is a raw `UserData.forLinux()` replicating the legacy `cfn-init` faithfully: `GuUserData` writes config to `/etc/<app>/` but this app reads `/etc/gu/`, and the private config lives in `facia-private`, not the distribution bucket. Order matters — the user and its config are created *before* the `.deb` is installed, because installing it starts the service.

**IAM is one policy per concern**, named for what it grants, rather than the legacy catch-all `root` policy. Deliberately *not* ported, because `GuInstanceRole` already grants them: ec2/autoscaling `Describe*`, Kinesis log shipping, artifact-bucket `s3:GetObject`, and SSM/SSH. `rds:DescribeDBInstances` *is* ported, since `GuDescribeEC2Policy` doesn't cover it. Note the legacy Parameter Store path (`/facia-tool/cms-fronts/<stage>/*`) differs from GuCDK's convention (`/<stage>/cms-fronts/facia-tool/*`), so both are present.

**Postgres access uses a dedicated security group** rather than the ASG's own `connections`. `Connections` replays its rules onto any group added to it later, so granting DB access via the ASG's connections — which also carries the shared CAPI endpoint group — would have opened the database to everything in that shared group. The CAPI endpoint group is imported with `mutable: false` so no rules are written into it. Instances end up in 3 security groups, comfortably under the limit of 5.

**A brand-new per-stage `GuCertificate`** is created for the domain CloudFront forwards as the `Host` header (`fronts.gutools.co.uk` / `fronts.code.dev-gutools.co.uk`), rather than reusing the legacy ELB's certificate ARN via an escape hatch. Validation records are created automatically, so there's no manual DNS step.

`imageRecipe` is the legacy recipe verbatim (`editorial-tools-jammy-java11`) — no recipe change is needed for this migration.

### Reviewing the diff

`cdk diff` against both live stacks is **purely additive**, apart from the intended `gu:riffraff:new-asg` tag removal from the legacy ASG. Two things in the diff that look odd but are expected:

- The ALB security group gets egress on 9000 to the DB-access and CAPI endpoint groups. That's a side effect of CDK's connections model; the load balancer never uses it.
- `cfn-lint` warns `W9007` "Array property 'Subnets' contains duplicate values" on the new ALB. It's a false positive — the three entries are `Fn::Select` of index 0, 1 and 2.

### Deploying

This creates a new ASG, so it's a **"dangerous" (new-ASG) Riff-Raff deploy**. CODE first, then PROD.

Smoke-test through the new ALB before phase 3, sending the real `Host` header (the ALB DNS name is the `LoadBalancerFaciatoolDnsName` stack output):

```
curl -sk -H 'Host: fronts.code.dev-gutools.co.uk' https://<alb-dns-name>/_healthcheck
```

One thing worth watching during the soak: the new instance security group allows **egress on 443 only** (the GuCDK default), where the legacy one allowed all outbound. Postgres (5432) is covered by the dedicated group and DNS isn't subject to security group rules, but if the app has an outbound dependency on some other port it would show up here.

The synthesized template is ~58KiB, over CloudFormation's 51,200-byte inline limit; Riff-Raff uploads templates that size to S3 automatically.

### General
- [x] 🤖 Relevant tests added — snapshot tests now cover both CODE and PROD
- [x] ✅ CI checks / tests run locally — `npm run lint`, `npm test`, `npm run synth` and `cdk diff` (CODE + PROD) all green
- [ ] 🔍 Checked on CODE

### Client
- [x] 🚫 No obvious console errors on the client (i.e. React dev mode errors) — n/a, infrastructure only
- [x] 🎛️ No regressions with existing user interactions (i.e. all existing buttons, inputs etc. work) — n/a, infrastructure only
- [x] 📷 Screenshots / GIFs of relevant UI changes included — n/a, infrastructure only

### DIFF
--- a/cdk/lib/facia-tool.test.ts
+++ b/cdk/lib/facia-tool.test.ts
@@ -1,12 +1,35 @@
-import { App } from "aws-cdk-lib";
-import { Template } from "aws-cdk-lib/assertions";
-import { FaciaTool } from "./facia-tool";
+import { App } from 'aws-cdk-lib';
+import { Template } from 'aws-cdk-lib/assertions';
+import { FaciaTool } from './facia-tool';
 
-describe("The FaciaTool stack", () => {
-  it("matches the snapshot", () => {
-    const app = new App();
-    const stack = new FaciaTool(app, "FaciaTool", { stack: "cms-fronts", stage: "TEST" });
-    const template = Template.fromStack(stack);
-    expect(template.toJSON()).toMatchSnapshot();
-  });
+describe('The FaciaTool stack', () => {
+	it('matches the CODE snapshot', () => {
+		const app = new App();
+		const stack = new FaciaTool(app, 'FaciaTool-euwest-1-CODE', {
+			stack: 'cms-fronts',
+			stage: 'CODE',
+			env: { region: 'eu-west-1' },
+			domainName: 'fronts.code.dev-gutools.co.uk',
+			instanceType: 't4g.small',
+			minimumInstances: 1,
+			maximumInstances: 2,
+		});
+		const template = Template.fromStack(stack);
+		expect(template.toJSON()).toMatchSnapshot();
+	});
+
+	it('matches the PROD snapshot', () => {
+		const app = new App();
+		const stack = new FaciaTool(app, 'FaciaTool-euwest-1-PROD', {
+			stack: 'cms-fronts',
+			stage: 'PROD',
+			env: { region: 'eu-west-1' },
+			domainName: 'fronts.gutools.co.uk',
+			instanceType: 't4g.medium',
+			minimumInstances: 3,
+			maximumInstances: 6,
+		});
+		const template = Template.fromStack(stack);
+		expect(template.toJSON()).toMatchSnapshot();
+	});
 });
--- a/cdk/lib/facia-tool.ts
+++ b/cdk/lib/facia-tool.ts
@@ -1,15 +1,348 @@
-import { join } from "path";
-import type { GuStackProps } from "@guardian/cdk/lib/constructs/core";
-import { GuStack } from "@guardian/cdk/lib/constructs/core";
-import type { App } from "aws-cdk-lib";
-import { CfnInclude } from "aws-cdk-lib/cloudformation-include";
+import { join } from 'path';
+import { GuEc2App } from '@guardian/cdk';
+import { AccessScope } from '@guardian/cdk/lib/constants';
+import type { GuStackProps } from '@guardian/cdk/lib/constructs/core';
+import {
+	GuDistributionBucketParameter,
+	GuStack,
+} from '@guardian/cdk/lib/constructs/core';
+import { GuSecurityGroup, GuVpc } from '@guardian/cdk/lib/constructs/ec2';
+import { GuAllowPolicy, GuPolicy } from '@guardian/cdk/lib/constructs/iam';
+import type { App } from 'aws-cdk-lib';
+import { Fn, Tags } from 'aws-cdk-lib';
+import type { ISubnet } from 'aws-cdk-lib/aws-ec2';
+import {
+	InstanceType,
+	Port,
+	SecurityGroup,
+	UserData,
+} from 'aws-cdk-lib/aws-ec2';
+import { Effect, PolicyStatement } from 'aws-cdk-lib/aws-iam';
+import { CfnInclude } from 'aws-cdk-lib/cloudformation-include';
+
+const app = 'facia-tool';
+const applicationPort = 9000;
+
+export interface FaciaToolProps extends GuStackProps {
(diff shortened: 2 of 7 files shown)


davidfurey:
Tested on CODE with /etc/hosts pointing to new ALB
- logging in works
- moving an item in a container works

cdk/lib/facia-tool.test.ts
```diff
@@ -1,12 +1,35 @@
...
-    const app = new App();
-    const stack = new FaciaTool(app, "FaciaTool", { stack: "cms-fronts", stage: "TEST" });
-    const template = Template.fromStack(stack);
-    expect(template.toJSON()).toMatchSnapshot();
-  });
+describe('The FaciaTool stack', () => {
+	it('matches the CODE snapshot', () => {
+		const app = new App();
+		const stack = new FaciaTool(app, 'FaciaTool-euwest-1-CODE', {
+			stack: 'cms-fronts',
```
waisingyiu, line 9: Could we export this object from `cdk.ts` so that we don't need to change two places when there are changes to this stack creation param object in future?

cdk/lib/facia-tool.ts
```diff
@@ -1,15 +1,348 @@
...
+					}),
+					new PolicyStatement({
+						effect: Effect.ALLOW,
+						actions: ['s3:ListBucket'],
+						resources: [bucketArn('facia-tool-store')],
+					}),
+				],
+			}),
+
+			new GuAllowPolicy(this, 'PanDomainPolicy', {
```
waisingyiu, line 313: I think we can consider using more specific construct like `GuGetS3ObjectsPolicy` for these policies and set stage-specific resources. 

Maybe it is good to do like-for-like setup in this PR and leave them to a follow-up if it'd useful.

waisingyiu (approved):
It looks great! Happy to approve. Just a few non-blocking comments that can be followed up in a separate PR if they are sensible.

cdk/lib/facia-tool.ts
davidfurey: For flexible-restorer I did this idiomatic improvement in a final cleanup PR after cutover when all of the legacy cloudformation had been removed

cdk/lib/facia-tool.test.ts
davidfurey: Actually, I think I should change this to the pattern in other projects where the test doesn't explicitly cover PROD and CODE but instead an imaginary TEST stage. That way there is no desire to keep the param values in sync. I'll fix in a follow up PR

davidfurey:
Successfully deployed to prod with manual Dangerous deploy in github.
Smoke test adding/removing an article from a container on a training front works

PR #2062, opened 2026-09-01
Fix test status display on clipboard cards

### LINKED ISSUE #29148: Fronts tool: test status is inaccurate when moving to and from clipboard
Active tests become "ready to launch" when moved to the clipboard or from the clipboard back to the same front. 

A test should remain active when it is moved around. This ticket is to check that nothing is changing to the test under the hood when a test is moved to and from the clipboard.

emma-imber:
## What's changed?
This PR handles the status message of tests on cards on the clipboard explicitly, instead of relying on the fallback of `Ready to launch`. `Ready to launch` isn't necessarily inaccurate, but in future if we made changes to this, having the clipboard messaging handled explicitly helps to make sure that this edge case is considered. It also makes what's going on a bit clearer to future developers. 

Currently, when moving a card with an active test to the clipboard, the test status shows `Ready to launch`. This is because the `selectLiveCardForGivenCardId` selector looks for the current collection ID in the live collections to find the live version of the card. The clipboard doesn't exist on the live collections, so no live card ID is returned. This means the code finds no active AB test.

This PR updates that logic to check for tests on the current card if it is on the clipboard, instead of using the `liveCard`, which will be undefined. This then shows the current test status for the clipboard card.

| Before | After |
| ------ | ------ |
| Screenshot 2026-09-08 at 17 40 41 | Screenshot 2026-09-08 at 17 38 14 |

### DIFF
--- a/fronts-client/src/components/OphanBanner.tsx
+++ b/fronts-client/src/components/OphanBanner.tsx
@@ -36,6 +36,7 @@ const OphanBannerLink = styled.a`
 `;
 
 interface OphanBannerProps {
+	isClipboard: boolean;
 	isTestLive: boolean;
 }
 
@@ -47,9 +48,11 @@ const OphanBanner = ({ ...props }: OphanBannerProps) => {
 			</FixedSizeIcon>
 			<OphanBannerText>
 				<OphanBannerTitle>
-					{props.isTestLive
-						? 'Headline test in progress: '
-						: 'Test ready to launch: '}
+					{props.isClipboard
+						? 'Test staged: '
+						: props.isTestLive
+							? 'Headline test in progress: '
+							: 'Test ready to launch: '}
 				</OphanBannerTitle>
 				view results {!props.isTestLive && 'from other cards with this test'} in{' '}
 				<OphanBannerLink
--- a/fronts-client/src/components/card/article/ArticleBody.tsx
+++ b/fronts-client/src/components/card/article/ArticleBody.tsx
@@ -7,7 +7,7 @@ import BasePlaceholder from '../../BasePlaceholder';
 import { getPillarColor } from 'util/getPillarColor';
 import CardMetaContainer from '../CardMetaContainer';
 import CardContent from '../CardContent';
-import { notLiveLabels, liveBlogTones } from 'constants/fronts';
+import { notLiveLabels, liveBlogTones, clipboardId } from 'constants/fronts';
 import TextPlaceholder from 'components/TextPlaceholder';
 import { ThumbnailSmall, ThumbnailCutout } from 'components/image/Thumbnail';
 import CardMetaHeading from '../CardMetaHeading';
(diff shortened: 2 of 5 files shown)


Fweddi:
If you drag a 'Test in progress' card from the clipboard to another collection on the same front, the status display now changes from 'Test in progress' to 'Ready to launch' (though when you open the card, it displays the test as in progress):

https://github.com/user-attachments/assets/ff6b0870-1fda-47cc-864d-353f68d4b694

What do we think is best here? Should the status display 'Test in progress' (to denote it's the same test, even though technically it hasn't been launched to the container), 'Ready to launch' (to denote the test hasn't been launched to the container), or even something like 'Test ready to move' (to denote it's the same test, but hasn't been launched to the container)? Should the status be the same as the status of the card as it is on the clipboard? I think we should address the mismatch of the opened / collapsed states at the very least.

emma-imber:
> If you drag a 'Test in progress' card from the clipboard to another collection on the same front, the status display now changes from 'Test in progress' to 'Ready to launch' (though when you open the card, it displays the test as in progress):
> 
>  Screen.Recording.2026-09-02.at.09.31.54.mov 
> What do we think is best here? Should the status display 'Test in progress' (to denote it's the same test, even though technically it hasn't been launched to the container), 'Ready to launch' (to denote the test hasn't been launched to the container), or even something like 'Test ready to move' (to denote it's the same test, but hasn't been launched to the container)? Should the status be the same as the status of the card as it is on the clipboard? I think we should address the mismatch of the opened / collapsed states at the very least.

I think the approach of adding a new message type for clipboard tests sounds sensible. Will pop this PR back into draft as I make that change and test

emma-imber:
> If you drag a 'Test in progress' card from the clipboard to another collection on the same front, the status display now changes from 'Test in progress' to 'Ready to launch' (though when you open the card, it displays the test as in progress):
> 
>  Screen.Recording.2026-09-02.at.09.31.54.mov 
> What do we think is best here? Should the status display 'Test in progress' (to denote it's the same test, even though technically it hasn't been launched to the container), 'Ready to launch' (to denote the test hasn't been launched to the container), or even something like 'Test ready to move' (to denote it's the same test, but hasn't been launched to the container)? Should the status be the same as the status of the card as it is on the clipboard? I think we should address the mismatch of the opened / collapsed states at the very least.

I have opened a PR to solve this mismatching problem here: https://github.com/guardian/facia-tool/pull/2065

Fweddi:
Nice, I tested locally and TIL, you can actually open a card which is on the clipboard.

If you do, there is a mismatch between the banner and the status again 🥲 (banner says in progress, status says staged):
image

emma-imber:
> Nice, I tested locally and TIL, you can actually open a card which is on the clipboard.
> 
> If you do, there is a mismatch between the banner and the status again 🥲 (banner says in progress, status says staged): image

This is a very good spot, I wasn't aware you could open the card on the clipboard! This should now be fixed 🤞

Screenshot 2026-09-08 at 17 38 23

Fweddi (approved):
Tested locally and LGTM!

PR #2065, opened 2026-09-03
Show different Ophan banner message if test is active but not live

emma-imber:
## What's changed?
Solves an issued flagged by @Fweddi [here](https://github.com/guardian/facia-tool/pull/2062#issuecomment-5506876952).

Currently, if a card with a test is moved to the clipboard, and then added to a different container on the front, there is a mismatch between the status indicator of 'Ready to launch' and the message in the Ophan banner which suggests the test is already live on that card. This is beacuse the Ophan banner is only looking at whether a card has an active test, not whether that test is actually live on that specific card. This PR changes the message in the Ophan banner depending on whether or not the test is launched on that card, to be more specific about the state the card test is currently in. This should help reduce confusion.

### Before

https://github.com/user-attachments/assets/069f595a-be55-4dd9-95f1-1bfb8cc6140f

### After

https://github.com/user-attachments/assets/a0bf5781-24fa-4e96-afbd-b4520ac5c76e

### DIFF
--- a/fronts-client/src/components/OphanBanner.tsx
+++ b/fronts-client/src/components/OphanBanner.tsx
@@ -19,6 +19,14 @@ const OphanBannerContainer = styled.div`
 	font-size: 14px;
 `;
 
+const FixedSizeIcon = styled.span`
+	flex: none;
+`;
+
+const OphanBannerText = styled.p`
+	margin: 0;
+`;
+
 const OphanBannerTitle = styled.span`
 	font-weight: 600;
 `;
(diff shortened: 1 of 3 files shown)


fronts-client/src/components/OphanBanner.tsx
```diff
@@ -27,13 +27,21 @@ const OphanBannerLink = styled.a`
...
+
+const OphanBanner = ({ ...props }: OphanBannerProps) => {
 	return (
 		<OphanBannerContainer role="status">
 			<ConicalFlaskIcon size="xs" fill={theme.abTest.active.icon} />
-			<OphanBannerTitle>Headline Test In Progress :</OphanBannerTitle>
+			<OphanBannerTitle>
+				{props.isTestLive
+					? 'Headline test in progress:'
+					: 'Test ready to launch on this card:'}
```
Fweddi: Wonder if this can be shortened to `Test ready to launch`, given the message after: `View results from other cards with this test`? While explicit, the length of the message makes it a little hard to parse.

Fweddi (approved):
I think this works well, nice job. The banner also shows when moving to another container on the same front which feels nice and consistent.

On narrow fronts the banner looks a little strange: 

Image

The suggestion I made to shorten the label might help with this (we could also add some styling on the icon to stop it shrinking).

fronts-client/src/components/OphanBanner.tsx
emma-imber: Yeah given the context of the second message that's reasonable

emma-imber:
> I think this works well, nice job. The banner also shows when moving to another container on the same front which feels nice and consistent.
> 
> On narrow fronts the banner looks a little strange:
> 
> Image
> The suggestion I made to shorten the label might help with this (we could also add some styling on the icon to stop it shrinking).

Yep good spot - I've updated the styling to make it more consistent across breakpoints. The icon is now a fixed size, and the text is all inline so that it wraps nicely in the space. I did play around a bit with different flex settings but I felt that this looked best. What do you think?

Wide:
Screenshot 2026-09-04 at 15 33 38

Medium:
Screenshot 2026-09-04 at 15 34 04

Narrow:
Screenshot 2026-09-04 at 15 33 52

Fweddi:
> > I think this works well, nice job. The banner also shows when moving to another container on the same front which feels nice and consistent.
> > On narrow fronts the banner looks a little strange:
> > Image
> > The suggestion I made to shorten the label might help with this (we could also add some styling on the icon to stop it shrinking).
> 
> Yep good spot - I've updated the styling to make it more consistent across breakpoints. The icon is now a fixed size, and the text is all inline so that it wraps nicely in the space. I did play around a bit with different flex settings but I felt that this looked best. What do you think?
> 
> Wide: Screenshot 2026-09-04 at 15 33 38
> 
> Medium: Screenshot 2026-09-04 at 15 34 04
> 
> Narrow: Screenshot 2026-09-04 at 15 33 52

I've tested this locally and I think it works great!

PR #2068, opened 2026-09-07
Fixes for devcontainer setup

fredex42:
## What's changed?

I tried to use the devcontainer setup, and found I couldn't get the app running properly in the browser 😢 

This PR fixes the following issues:
- forwards port 9000 from the container
- creates the /etc/gu directory and updates its ownership, post-build, so that the `setup.sh` script completes
- adds useful tooling to the image
- updates the README instructions

> [!IMPORTANT]
> When forwarding ports to the host system, Docker binds the port when the container starts.  Therefore, if you have any other containers (or apps) binding port 9000 when you start the dev container you won't be able to access the Fronts tool when it starts.  The fix is to shut down any other dev containers which might be using port 9000 (i.e., `curl http://localhost:9000` on the host system tells you "Could not connect" immediately) and then start this dev container again

## Testing instructions

- Check out the repo locally
- Go through the updated README instructions

### TL;DR
- Open in VSCode
- In the selector at the top, type `> re` and select `Dev Containers: Reopen in container`
- Watch it build
- When it completes, open a Terminal
- Get cms-fronts permissions
- Run

```bash
./scripts/setup.sh
```

- When this is complete, start up the app by running

```bash
./scripts/dev-start.sh
```

- So long as you have the dev-nginx setup done _on your host machine_, you should now be able to access the whole app through https://fronts.local.dev-gutools.co.uk

## TODO

- ~~work out why dev-setup is not populating /etc/gu with the configuration files~~
- ~~update the README~~
- ~~Why does the frontend not work correctly in browser when running in IntelliJ, but works fine in VSCode with the same configuration~~ Fixed by @Fweddi 😄 

## Implementation notes

The manual updates are always limited to `devenv.yaml`, so we can regenerate it without issue

## Checklist

n/a


.devcontainer/devenv.yaml
```diff
@@ -20,8 +20,8 @@ modules:
   - github-copilot
 
 # Optional: Ports to forward
-# forwardPorts:
-#   - 8080  # same port on host and container
+forwardPorts:
+  - 9000  # same port on host and container
```
Fweddi: do we also want 5173?

fredex42: We don't seem to _need_ it. I assumed that we would, but dev-nginx only talks to 9000 since the frontend is hosted out via the Scala app I guess

Fweddi:
Nice, thanks for this! It would be useful if the setup script also worked within the devcontainer. I think at the moment it fails?

fredex42:
> Nice, thanks for this! It would be useful if the setup script also worked within the devcontainer. I think at the moment it fails?

😄  Working on it!

fredex42:
> Nice, thanks for this! It would be useful if the setup script also worked within the devcontainer. I think at the moment it fails?

✅  It's now tested and working

Fweddi:
I couldn't get this working until I forwarded port 5173: https://github.com/guardian/facia-tool/commit/73918bc0494cdb24acfc9c1783a1e042d44523dc. I'm using IntelliJ.

fredex42:
> I couldn't get this working until I forwarded port 5173: [73918bc](https://github.com/guardian/facia-tool/commit/73918bc0494cdb24acfc9c1783a1e042d44523dc). I'm using IntelliJ.

@Fweddi Confirmed, thanks - merged your branch into this PR

Fweddi (approved):

PR #2074, opened 2026-09-09
Add images to create subnav

Georges-GNM:
## What's changed?

Co-written with Opus 4.8

Makes use of the latest fapi client release to enable adding images to the subnav.

We've agreed with editorial design that we would use a shortcut when it comes to storing images: the s3 uploader tool ([code environment](https://s3-uploader.code.dev-gutools.co.uk/)). The workflow then is for the person uploading images to upload all of them via the tool, where they receive an image url; they will then be able to input those urls in the text fields provided in this interface to add those images to a custom subnav. 

In the future, it would be preferable to not require a user to exit this interface, and instead be able to directly upload the required images. However, due to storage and access requirements, and the fact that uploading images to the Grid doesn't bring enough value compared to the friction of doing so, the s3 uploader ends up being a decent starting point until we want to explore another way to store these images in a way that would allow them to be publicly visible. 

In any case, for now, we therefore just need the user to be able to input an image src url. A distinction is made between the ios vs web/android: current designs expect the latter two to use the same images, but ios will need different assets. And each platform can take up to 8 breakpoints, so we accommodate those in the interface too. 

## Screenshots

image

## Testing

With this [DCR branch](https://github.com/guardian/dotcom-rendering/pull/16664), you can see images being rendered on web/preview. 

### General
- [ ] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client


fronts-client/src/components/Subnav/SubnavImagesSection.tsx
```diff
@@ -0,0 +1,193 @@
...
+	};
+
+	return (
+		<>
+			{platformGroups.map((group) => (
+				<PlatformGroupImages
+					key={group.id}
+					group={group}
+					images={images}
+					findIndex={findIndex}
```
SiAdcock, line 185: ~Is there a reason `findIndex` needs to be passed as a prop, rather than defined in the `<PlatformGroupImages>` component? The function definition can be derived from `images` which is also passed as a prop~

Wait, I see it's also used in `setImageSrc`. Carry on!

SiAdcock (approved):
👍

PR #2064, opened 2026-09-03
GuCDK migration phase 3: serve CloudFront from the new ALB

davidfurey:
> [!NOTE] 
> This PR was opened by Claude Opus 5 using cloudformation-to-gucdk-migration@7fe540d31beb7ee7bc0c7793b4add833ee0275d1 but I have reviewed the output

## What's changed?

Phase 3 of the [CloudFormation → GuCDK migration](https://github.com/guardian/facia-tool/blob/main/plans/cloudformation-to-cdk-migration.md): **switch live traffic from the legacy classic ELB to the ALB** that [#2061](https://github.com/guardian/facia-tool/pull/2061) stood up alongside it.

facia-tool is fronted by a CloudFront distribution whose origin is the ELB, so the traffic switch is simply repointing that origin — there is no DNS change:

```ts
cfnInclude
  .getResource('FaciaCloudfront')
  .addPropertyOverride(
    'DistributionConfig.Origins.0.DomainName',
    ec2App.loadBalancer.loadBalancerDnsName,
  );
```

`cdk diff` against **both** live stacks is that one property, and nothing else:

```
Stack FaciaTool-euwest-1-PROD (facia-PROD)
Resources
[~] AWS::CloudFront::Distribution YamlTemplate/FaciaCloudfront FaciaCloudfront
 └─ [~] DistributionConfig
     └─ [~] .Origins:
         └─ @@ -6,7 +6,7 @@
            [ ] "DomainName": {
            [ ]   "Fn::GetAtt": [
            [-]     "FaciaLoadBalancerNewVPC",
            [+]     "LoadBalancerFaciatool65BDBE55",
            [ ]     "DNSName"
```

## Implementation notes

**Rollback is reverting this PR.** The legacy ELB, ASG and launch config all keep running untouched through Phase 3 — this changes only where CloudFront sends requests. Origin changes propagate in minutes. Phase 4 deletes the legacy compute, and only once CloudWatch shows the old ELBs at 0 requests.

**The distribution stays in the wrapped template.** CDK overrides one property rather than adopting CloudFront, so aliases, cache behaviour, the viewer certificate and the `Guardian::DNS::RecordSet` records are all untouched. Moving the distribution into GuCDK proper is a Phase 5 follow-up, deliberately not part of the cutover.

**Origin id is unchanged** (`facia-tool`), so `TargetOriginId` on the cache behaviour still resolves to the same origin entry.

[...]

PR #2081, opened 2026-09-16
GuCDK migration phase 4: delete the legacy compute resources

davidfurey:
> [!NOTE] 

## What's changed?

Phase 4 of the [CloudFormation → GuCDK migration](https://github.com/guardian/cdk/blob/main/docs/migration-guide.md): **deletes the legacy compute** now that CloudFront serves `fronts[.code.dev]-gutools.co.uk` from the `GuEc2App` ALB (#2064).

Removed from `cloudformation/facia-tool.cfn.yaml`:

| | |
|---|---|
| Compute | `FaciaLoadBalancerNewVPC` (classic ELB), `FaciaAutoscalingGroupNewVPC`, `FaciaLaunchConfigNewVPC` |
| Networking | `AppServerSecurityGroupNewVPC`, `LoadBalancerSecurityGroupNewVPC`, `AppToNewDBEgressNewVPC`, `NewDBToAppIngressNewVPC` |
| IAM | `DistributionRole`, `DistributionInstanceProfile`, and the eight policies that existed only to attach to that role (`PanDomainPolicy`, `PermissionsPolicy`, `CrossAccountPolicy`, `LogServerPolicy`, `SwitchesPolicy`, `CloudwatchPolicy`, `DynamoPressStatus`, `AssumeCapiPreviewRolePolicy`) |
| Config | Parameters `AMI`, `CertificateArn`, `AvailabilityZones`, `ELKKinesisStreamArn`, `ELKKinesisStreamName`, the already-dead `DBSecurityGroupId`; the `StageMap` sizing keys; the `FaciaLoadBalancerDNS` output |

Every one of those IAM policies is already reproduced on the `GuEc2App` instance role (added in #2061), so this is not a loss of permissions.

Also drops the `gu:riffraff:new-asg` tag from the new ASG, and `asgMigrationInProgress` + the legacy `AMI` entry from `riff-raff.yaml`.

## Implementation notes

**The old ELBs really are idle.** In the 24h after the phase 3 deploy, both classic ELBs returned **zero 2XX** — CODE 126×3XX/262×4XX, PROD 238×3XX/862×4XX. That residual is unauthenticated noise hitting the public ELB DNS/IPs directly (pan-domain auth redirects and not-founds), i.e. scanners rather than users, and it goes away with the ELB. Meanwhile the new ALBs carry the real load: PROD 289,288 requests (256,581×2XX) and CODE 15,142 (11,301×2XX) over the same window, matching the legacy PROD ELB's previous ~250–330k/day. No CloudFront distribution in the account still lists a legacy-ELB origin.

**`StorageBucket` needed care** — it was attached to *both* `DistributionRole` and `StorageConsumerRole`, so it survives with only the latter rather than being deleted.

**The CloudFront origin placeholder.** `FaciaCloudfront` can no longer `Fn::GetAtt` the deleted ELB, so the YAML now holds `overridden-by-cdk.invalid` and the existing phase 3 `addPropertyOverride` supplies the real ALB DNS name at synth. The placeholder is verified absent from both synthesized templates, and neither `cdk diff` shows any CloudFront change — it's a synth-time no-op. Migrating the distribution into GuCDK properly is the phase 5 follow-up.

**Kept deliberately:** the CODE-only `RunFaciaToolLocally` developer policy, and all the non-compute resources (both CloudFront distributions, DNS records, DynamoDB, SNS, `StorageConsumerRole`). This is a mixed stack, so the end state stays `CDK(cfn.yaml) -> cfn.json` rather than the template emptying out.

**`cdk diff` against both live stacks is identical and removals-only** — the 17 resources destroyed, the parameters and mapping keys dropped, `StorageBucket.Roles` losing `DistributionRole`, and the ASG losing its migration tag. No replacements, no additions.

### General
- [x] 🤖 Relevant tests added 
- [ ] 🔍 Checked on CODE 

### Client


davidfurey:
Requires a dangerous deploy

davidfurey:
Seems fine in CODE

waisingyiu (approved):
It looks great!

Just a non-blocking question - will we finally migrate everything from the cloudformation yaml? Or there are not much value in doing this?

davidfurey:
> It looks great!
> 
> Just a non-blocking question - will we finally migrate everything from the cloudformation yaml? Or there are not much value in doing this?

Yes, I believe that is phase 5

PR #2083, opened 2026-09-18
GuCDK migration phase 5a: move stateful resources into CDK

davidfurey:
> [!NOTE] 

## What's changed?

Phase 5a of the CloudFormation → GuCDK migration ([plan](https://github.com/guardian/facia-tool/blob/main/plans/cloudformation-to-cdk-migration.md)).

Phases 1–4 moved the *compute* into GuCDK. This moves the **stateful and shared resources** out of the wrapped YAML template and into the CDK stack:

- `FrontsUpdateSNSTopic`, `FeastPublicationTopic` and `FrontsUpdateSNSPolicy`
- `StorageConsumerRole` and its `StorageBucket` policy
- `FrontsUserDataDynamoTable`
- `RunFaciaToolLocally`, the CODE-only developer policy
- all three stack outputs, including both `Export`s

**No resource is created or destroyed.** Every one keeps its original logical ID via `GuStack.overrideLogicalId`, so CloudFormation sees in-place updates.

## Implementation notes

**The `FrontsUpdateSNSTopicARN` export is imported by `cms-fronts-<stage>-eventbridge-to-fanout`** (confirmed with `aws cloudformation list-imports`). CloudFormation refuses to change an export that is in use, so the topics keep their logical IDs and the exported ARNs are unchanged.

**`RunFaciaToolLocally` had to come along.** It references the topics and the table, whose ARNs are not reconstructable once those resources leave the template. Its CFN `Condition: IsCode` becomes a `this.stage === 'CODE'` guard in CDK, so it is simply absent from the PROD template — where `describe-stack-resources` confirms it has never existed.

### `cdk diff` over-reports replacement — worth knowing

`cdk diff` reported **`requires replacement`** on `FrontsUserDataDynamoTable` (`TableName`) and `RunFaciaToolLocally` (`Path`). Both were false alarms: the resolved value is unchanged and only the expression differs (a template literal rather than the YAML's `Fn::Join`/`Fn::Sub`).

The read-only profile cannot `CreateChangeSet`, so `cdk diff` had fallen back to a **template-only** diff, which flags any textual change to a property the resource spec marks "update requires replacement", without resolving intrinsics.

**A real change set settled it** — CloudFormation resolves intrinsics before comparing, and reports `Replacement: False` for both. I had initially worked around the warning by matching the YAML's intrinsics exactly; that is now reverted and the code uses plain template literals.

### Verification

- **A change set against live `facia-CODE` contains no `Remove` actions and no replacements** — every resource is `Modify` / `Replacement: False`, or `Conditional` where a property derives from a `Ref`/`GetAtt` on a parent that is itself `Replacement: False` (the security-group rules and the ASG).
- A script resolved the intrinsics in the live and synthesized templates and compared the effective statements: `RunFaciaToolLocally` (16 statements), `StorageBucket` (2), `StorageConsumerRole` (6) and `FrontsUpdateSNSPolicy` (3) all match exactly, as do the table name and both export names, in both stages.
- The remaining `cdk diff` output is representational: IAM documents moving from `Fn::Sub` to `Fn::Join` with a partition token, the table gaining `DeletionPolicy`/`UpdateReplacePolicy: Retain` (a CDK default, and an improvement), the launch template's `USER_DATA_TABLE` resolving via `Ref: FrontsUserDataDynamoTable`, and the unused `IsCode` condition going away.

### General
- [x] 🤖 Relevant tests added — snapshot tests updated for both stages
- [x] ✅ CI checks / tests run locally — lint, tests and synth green
- [ ] 🔍 Checked on CODE

### Client
- [x] 🚫 No client changes
- [x] 🎛️ No client changes
- [x] 📷 No UI changes


davidfurey:
Changeset shows 3 changes:
1) deletion policy: retain set for dynamodb table (a good thing)
2) IAM Policy change from `["arn:aws:s3:::facia-tool-store"]` to `"arn:aws:s3:::facia-tool-store"` (no op)
3) IAM Role change from version 2008-10-17 to 2012-10-17 (no op)

cdk/lib/facia-tool.ts
jonathonherbert, line 57: `getParameterFromCfnTemplate`? Appreciate it's not introduced in this PR, but as we're passing it into another fn here it's not super clear what the argument does

dominickendrick (approved):
Looks good. Created a changeset on code and observed 3 meaningful changes, and no other erroneous updates

jonathonherbert (approved):

cdk/lib/facia-tool.ts
davidfurey: Reasonable feedback, but I'd rather address this in a follow up PR. For these migration PRs, I am largely thinking of them as intermediary safe deployment points, to get us to a point where we have a stable snapshot test, and then we can focus on refactoring the TS knowing that nothing has changed apart from readability.

PR #2084, opened 2026-09-18
GuCDK migration phase 5b: move CloudFront and DNS into CDK

davidfurey:
> [!WARNING] 
> This PR was opened by Claude Opus 5 using cloudformation-to-gucdk-migration@7fe540d31beb7ee7bc0c7793b4add833ee0275d1 and I have not reviewed the output

## What's changed?

Phase 5b of the CloudFormation → GuCDK migration ([plan](https://github.com/guardian/facia-tool/blob/main/plans/cloudformation-to-cdk-migration.md)).

**Depends on #2083 — merge and deploy that first.**

This moves the last of the YAML into CDK and **deletes `cloudformation/facia-tool.cfn.yaml` and the `CfnInclude`**, so the stack reaches the `CDK -> cfn.json` end-state:

- `FaciaCloudfront` and `StaticCloudfront`
- `DnsRecord` and `StaticCloudFrontDnsRecord` (the two NS1 CNAMEs)
- the remaining 18 template parameters

**No resource is created or destroyed**, and no CloudFront config changes.

## Implementation notes

**The distributions are L1 `CfnDistribution`s, deliberately.** The legacy config uses `ForwardedValues`, which the L2 `Distribution` construct cannot express at all — it requires cache policies. Converting would change caching behaviour, and that must not ride along in a migration PR. If we want cache policies, that is a separate, testable change.

**The placeholder is gone.** Phase 4 left `FaciaCloudfront`'s origin as `overridden-by-cdk.invalid` in the YAML, patched at synth by an `addPropertyOverride`. The origin now takes `ec2App.loadBalancer.loadBalancerDnsName` directly.

**DNS** becomes `GuCname`. `GuDnsRecordSet` uses the construct id as the logical ID, so `DnsRecord` and `StaticCloudFrontDnsRecord` are preserved by naming alone.

**Parameters** are recreated as `CfnParameter`s with the same logical IDs, types and defaults, so Riff-Raff carries over the previous values of the ones without a default (account IDs, VPC, subnets, certificate…). `Stage` is now unreferenced, but it is kept: Riff-Raff supplies a `Stage` parameter when the template declares one, and removing it was not covered by the verified change set.

**`Mappings` are gone.** `CloudFrontAliases`/`StaticCloudFrontAliases` become the `domainName`/`staticDomainName` props, `LowerCaseStage` becomes `this.stage.toLowerCase()`, and `CrossResources.FrontPressedTable` becomes a `frontPressedTable` prop — supplied per-stage from `bin/cdk.ts`, so the stack class describes shape and the entrypoint supplies values.

### Verification

- **A change set against live `facia-CODE`** (covering 5a + 5b + 5c together) contains **no `Remove` actions and no replacements**. `FaciaCloudfront`, `StaticCloudfront`, `DnsRecord` and `StaticCloudFrontDnsRecord` are all `Modify` / `Replacement: False`.
- The template CloudFormation validated is **identical** to what this branch synthesizes, bar CDK's own `CDKMetadata` analytics resource.
- A script resolved the intrinsics in the live and synthesized templates and compared the **fully-resolved properties of every migrated resource**. In both stages they match, as do all 22 parameters and both exports. The one exception is `StorageConsumerRole`'s `AssumeRolePolicyDocument` gaining an explicit `"Version": "2012-10-17"`, which CDK always sets; the statements match and the trust policy uses no policy variables, so it has no behavioural effect.
- The remaining `cdk diff` output is every `Fn::FindInMap`/`Ref` collapsing to the literal it already resolved to, plus the now-unused `Mappings`, `Conditions` and `AWSTemplateFormatVersion` being dropped. `description` is set on the stack so the template `Description` does not churn either.

Nothing outside the plan files referenced `cloudformation/`; CI only uploads `cdk/cdk.out/*.template.json`.

### General
- [x] 🤖 Relevant tests added — snapshot tests updated for both stages
- [x] ✅ CI checks / tests run locally — lint, tests and synth green
- [ ] 🔍 Checked on CODE

### Client
- [x] 🚫 No client changes
- [x] 🎛️ No client changes
- [x] 📷 No UI changes


davidfurey:
I've created a changeset which confirms this makes no changes

(this was before some force-pushes to resolve merge conflicts, but the generated cloudformation is identical so no re-test was required)

waisingyiu (approved):
👍

PR #2086, opened 2026-09-18
Packages backend pt1

fredex42:
## What's changed?

Implements the backend data models and database evolutions for both Recipe Packages (Feast's "Guardian Selections") and future Story Packages work.

The functionality is not yet exposed, so testing is done via the test suites provided in the PR

## Implementation notes

The `EditionsDB` base classes have been renamed FaciaDB, since Editions is now quite a misleading name

### General
- [x] 🤖 Relevant tests added
- [ ] 🔍 Checked on CODE

### Client

n/a, there are no client changes


conf/evolutions/default/14.sql
```diff
@@ -0,0 +1,42 @@
+# --- !Ups
+
+CREATE TABLE packages (
+						  id            TEXT    PRIMARY KEY DEFAULT uuid_generate_v4()::TEXT,
+
+						  name          TEXT    NOT NULL,
+						  is_hidden     BOOLEAN NOT NULL,
+						  package_type  TEXT,
```
davidfurey, line 8: A comment like line 25 would be helpful here

conf/evolutions/default/14.sql
```diff
@@ -0,0 +1,42 @@
+# --- !Ups
+
+CREATE TABLE packages (
+						  id            TEXT    PRIMARY KEY DEFAULT uuid_generate_v4()::TEXT,
+
+						  name          TEXT    NOT NULL,
+						  is_hidden     BOOLEAN NOT NULL,
```
davidfurey, line 7: I haven't looked at the convention in this project, but I'd probably call this `hidden` rather than `is_hidden`, but no objection if you disagree

conf/evolutions/default/14.sql
```diff
@@ -0,0 +1,42 @@
...
+CREATE TABLE package_cards (
+							   package_id    TEXT   REFERENCES packages(id) ON DELETE CASCADE NOT NULL,
+
+							   card_type	   TEXT 	  NOT NULL, -- 'article','recipe','chef', etc.
+							   page_code      TEXT        NOT NULL,
+							   index          INT         NOT NULL,
+
+							   metadata       JSONB,
+
+							   added_on       TIMESTAMPTZ NOT NULL,
```
davidfurey, line 31: Minor, but why does this not have a default but line 11 does?

conf/evolutions/default/14.sql
```diff
@@ -0,0 +1,42 @@
...
+							   added_by       TEXT        NOT NULL,
+							   added_email    TEXT        NOT NULL,
+							   PRIMARY KEY (package_id, page_code)
```
davidfurey, line 34: Should card_type be part of this?

conf/evolutions/default/14.sql
```diff
@@ -0,0 +1,42 @@
...
+);
+
+CREATE INDEX idxPackageCards ON package_cards (package_id);
+
+# --- !Downs
+
+DROP TABLE package_cards;
+DROP TABLE packages CASCADE;
```
davidfurey, line 42: Why is the cascade necessary here but not on line 41?

app/model/packages/client/CreatePackageRequest.scala
```diff
@@ -0,0 +1,25 @@
...
+import org.postgresql.util.PGobject
+import play.api.libs.json.{Json, OFormat}
+
+case class CreatePackageRequest(
+    id: String,
+    name: String,
+    isHidden: Boolean,
+    packageType: Package.PackageType.Value,
+    metadata: Option[PackageMetadata],
+    createdOn: Long,
```
davidfurey, line 13: millis or seconds?

app/model/packages/Package.scala
```diff
@@ -0,0 +1,85 @@
...
+    createdOn: Option[OffsetDateTime],
+    createdBy: Option[String],
+    createdEmail: Option[String],
+    updatedOn: Option[OffsetDateTime],
+    updatedBy: Option[String],
+    updatedEmail: Option[String]
+) extends MetadataHelpers {
+
+  def metadataPG: Option[PGobject] = metadata flatMap { meta =>
+    Try { Json.toJson(meta) }.toOption.map(toPGobject)
```
davidfurey, line 32: Can `Json.toJson` throw?

Also, why is this different to https://github.com/guardian/facia-tool/pull/2086/changes#diff-0731ca7ed69600177f12b50e69d9dd976a45b2bf1c53daeee766fbace7c9e71bR19 ?

davidfurey: Or, rather, why are both not `metadata.map(Json.toJson).map(toPGobject)`?

app/model/packages/Package.scala
```diff
@@ -0,0 +1,85 @@
...
+  private def getMetadata(
+      rs: WrappedResultSet,
+      id: String,
+      packageType: PackageType.Value
+  ): Option[PackageMetadata] = packageType match {
+    case PackageType.Feast =>
+      try {
+        rs.stringOpt("metadata")
+          .map(Json.parse)
+          .map(_.as[FeastPackageMetadata])
```
davidfurey, line 54: I think there is a `.validate` method you can use instead which avoids a try-catch

app/model/packages/PackageCard.scala
```diff
@@ -0,0 +1,134 @@
+package model.packages
+
+import model.editions.EditionsChefMetadata
+import model.editions.EditionsFeastCollectionMetadata
+import model.packages.PackageCardType
+import model.packages.PackageCardType.Recipe
+import play.api.libs.json._
+import scalikejdbc.WrappedResultSet
+
+import java.time.Instant
```
davidfurey, line 10: Is there a reason for `Instant` in some places and `OffsetDateTime` in others?

app/model/packages/PackageCard.scala
```diff
@@ -0,0 +1,134 @@
...
+        PackageChefCard.format.writes(c)
+      case c: PackageSubcollectionCard =>
+        PackageSubcollectionCard.format.writes(c)
+    }
+  }
+
+  def fromRowOpt(rs: WrappedResultSet): Option[PackageCard] = {
+    for {
+      id <- rs.stringOpt("id")
+      cardTypeStr <- rs.stringOpt("card_type")
```
davidfurey, line 101: These aren't optional in the db, why imply they are here?

app/model/packages/PackageCardType.scala
```diff
@@ -0,0 +1,42 @@
...
+  case object Subcollection extends PackageCardType {
+    override def toString = "subcollection"
+  }
+
+  def fromString(value: String): Option[PackageCardType] = value match {
+    case "recipe"        => Some(Recipe)
+    case "chef"          => Some(Chef)
+    case "subcollection" => Some(Subcollection)
+    case _               => None
+  }
```
davidfurey, line 28: Up to you, but:
```suggestion
  def fromString(value: String): Option[PackageCardType] = scala.PartialFunction.condOpt(value) collect {
    case "recipe"        => Recipe
    case "chef"          => Chef
    case "subcollection" => Subcollection
  }
```

app/model/packages/PackageCard.scala
```diff
@@ -0,0 +1,134 @@
...
+      cardType <- PackageCardType.fromString(cardTypeStr)
+      addedOn <- rs.zonedDateTimeOpt("added_on").map(_.toInstant)
```
davidfurey, line 103: This isn't optional in the db, why imply it is here?

app/model/packages/PackageCard.scala
```diff
@@ -0,0 +1,134 @@
...
+    } yield cardType match {
+      case Recipe =>
+        val recipeId = rs.string("page_code")
+        PackageRecipeCard(
+          id = recipeId,
+          addedOn = addedOn
+        )
+      case PackageCardType.Chef =>
+        val metadata = rs
+          .stringOpt("feast_metadata")
```
davidfurey, line 113: db field has changed name to `metadata`

app/model/packages/PackageCard.scala
```diff
@@ -0,0 +1,134 @@
...
+          .map(Json.parse)
+          .map(_.as[EditionsChefMetadata])
```
davidfurey, line 115: `.validate` is safer

app/model/packages/PackageCard.scala
```diff
@@ -0,0 +1,134 @@
...
+        PackageChefCard(
+          id = chefId,
+          metadata = metadata,
+          addedOn = addedOn
+        )
+      case PackageCardType.Subcollection =>
+        val metadata = rs
+          .stringOpt("feast_metadata")
+          .map(Json.parse)
+          .map(_.as[EditionsFeastCollectionMetadata])
```
davidfurey, line 126: As above

app/model/packages/PackageCard.scala
```diff
@@ -0,0 +1,134 @@
...
+      case c: PackageRecipeCard =>
+        PackageRecipeCard.format.writes(c)
+      case c: PackageChefCard =>
+        PackageChefCard.format.writes(c)
+      case c: PackageSubcollectionCard =>
+        PackageSubcollectionCard.format.writes(c)
+    }
+  }
+
+  def fromRowOpt(rs: WrappedResultSet): Option[PackageCard] = {
```
davidfurey, line 98: What is the overlap between this and https://github.com/guardian/facia-tool/pull/2086/changes#diff-415847ac38e7e6fa045c57b2cfd9e54a5f397db59180e3b9a39702e65565569f

app/model/packages/PackageCardRow.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+object PackageCardRow {
+  implicit val format: OFormat[PackageCardRow] = Json.format[PackageCardRow]
+
+  def fromRow(rs: WrappedResultSet): PackageCardRow =
+    PackageCardRow(
+      packageId = rs.string("package_id"),
+      cardType = PackageCardType
```
davidfurey, line 35: Why does this throw but PackageCard.fromRowOpt doesn't?

app/model/packages/PackageCardRow.scala
```diff
@@ -0,0 +1,49 @@
...
+        .fromString(rs.string("card_type"))
+        .getOrElse(
+          throw new IllegalArgumentException(
+            s"Invalid card type: ${rs.string("card_type")}"
+          )
+        ),
+      pageCode = rs.string("page_code"),
+      index = rs.int("index"),
+      metadata = rs.stringOpt("metadata").map(Json.parse),
```
davidfurey, line 44: Can `Json.parse` throw?

app/model/packages/PackageCardType.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  def apply(value: String): Option[PackageCardType] = fromString(value)
+
+  implicit val reads: Reads[PackageCardType] = Reads { json =>
+    json.asOpt[String].flatMap(fromString) match {
+      case Some(value) => play.api.libs.json.JsSuccess(value)
```
davidfurey, line 34: Import play.api.libs.json.JsSuccess and play.api.libs.json.JsError

app/model/packages/PackageCardType.scala
```diff
@@ -0,0 +1,42 @@
...
+      case None        => play.api.libs.json.JsError("Invalid PackageCardType")
+    }
+  }
+
+  implicit val writes: Writes[PackageCardType] = Writes { cardType =>
+    play.api.libs.json.JsString(cardType.toString)
```
davidfurey, line 40: And import this

app/model/packages/PackageMetadata.scala
```diff
@@ -0,0 +1,43 @@
...
+
+object FeastPackageMetadata {
+  implicit val format: OFormat[FeastPackageMetadata] =
+    Json.format[FeastPackageMetadata]
+}
+
+case class WebPackageMetadata(
+    // just examples at the moment
+    headline: Option[String],
+    customKicker: Option[String]
```
davidfurey, line 23: Probably better not to speculate?

app/model/packages/PackageMetadata.scala
```diff
@@ -0,0 +1,43 @@
...
+      case f: FeastPackageMetadata => FeastPackageMetadata.format.writes(f)
+      case w: WebPackageMetadata   => WebPackageMetadata.format.writes(w)
+    }
+
+    override def reads(json: JsValue): JsResult[PackageMetadata] = {
+      FeastPackageMetadata.format.reads(json) orElse WebPackageMetadata.format
+        .reads(json)
+    }
+  }
+}
```
davidfurey, line 43: I think we should avoid having a PackageMetadata json formatter since the discriminator between the feast and web package is either in Package or in the DB

app/model/packages/Package.scala
```diff
@@ -0,0 +1,85 @@
...
+  * of Cards. The difference to a Collection is that a Collection must belong to
+  * a Front; a Package is a top-level object that exists independently of any
+  * Front, Issue or Edition
+  */
+final case class Package(
+    id: String,
+    name: String,
+    isHidden: Boolean,
+    packageType: PackageType.Value,
+    metadata: Option[PackageMetadata],
```
davidfurey, line 22: To avoid issues knowing which kind of metadata the package has when serialising, I think Package should be a sealed trait with two implementations

app/services/editions/db/PackageQueries.scala
```diff
@@ -0,0 +1,430 @@
...
+  ): Seq[Package] =
+    DB readOnly { implicit session =>
+      {
+        val idList = packageIds.getOrElse(Seq.empty).map(_.toString)
+
+        val maybeIdCondition =
+          if (idList.nonEmpty) Some(sqls.in(sqls"id", idList)) else None
+
+        val maybeDateCondition = lastModified.map { modifiedSince =>
+          if (strictTimestamp) {
```
davidfurey, line 47: So strict timestamp means "on this day" and not-strict means "before this exact time"? This is unintuitive to me

app/services/editions/db/PackageQueries.scala
```diff
@@ -0,0 +1,430 @@
...
+      sql"""INSERT INTO package_cards (package_id, card_type, page_code, index, metadata, added_on, added_by, added_email) VALUES (${packageId.toString}, ${card.cardType.toString}, ${card.pageCode}, ${card.index}, ${card.metadataPG}, ${card.addedOn}, ${card.addedBy}, ${card.addedEmail})""".update
+        .apply()
+  }
+
+  /** Removes the card with the given pageCode from the given package. Since the
+    * primary key is (packageId, pageCode) this is sufficient to uniquely
+    * identify a package
+    * @param packageId
+    *   package ID to update
+    * @param pageCode
```
davidfurey, line 139: Is it possible for a chef and an article to have the same pageCode?

app/services/editions/db/PackageQueries.scala
```diff
@@ -0,0 +1,430 @@
...
+  def updatePackageMeta(
+      packageId: UUID,
+      newMeta: PackageMetadata,
+      userName: String,
+      userEmail: String
+  ) = {
+    val metaPg = toPGobject(PackageMetadata.format.writes(newMeta))
+    val fieldName = newMeta match {
+      // sqls is needed to inject the column names by value into the sql rather than as parameterised value
+      case _: FeastPackageMetadata => sqls"feast_metadata"
```
davidfurey, line 173: I think this has changed?

app/model/packages/client/CreatePackageRequest.scala
```diff
@@ -0,0 +1,25 @@
+package model.packages.client
+
+import model.packages.{MetadataHelpers, PackageMetadata, Package}
+import org.postgresql.util.PGobject
+import play.api.libs.json.{Json, OFormat}
+
+case class CreatePackageRequest(
+    id: String,
```
davidfurey, line 8: Should this be a UUID?

app/services/editions/db/PackageQueries.scala
```diff
@@ -0,0 +1,430 @@
...
+    *   set of cards to belong to the package. Any cards not in this list will
+    *   be removed.
+    * @return
+    *   count of affected packages
+    */
+  def updatePackage(packageMeta: Package, packageContent: Seq[PackageCardRow]) =
+    DB localTx { implicit session =>
+      // FOR UPDATE locks the selected rows for the duration of this transaction, allowing us to safely update without a race condition
+      val existingContent = fetchPackageContentSql(where =
+        sqls"WHERE package_id=${packageMeta.id} FOR UPDATE"
```
davidfurey, line 202: I'm not sure if you are consistent in upper/lower case for this id

app/services/editions/db/PackageQueries.scala
```diff
@@ -0,0 +1,430 @@
...
+        is_hidden,
+        metadata,
+        created_on,
+        created_by,
+        created_email,
+        updated_on,
+        updated_by,
+        updated_email
+	) VALUES (
+ 	   ${metadata.id.toLowerCase},
```
davidfurey, line 271: I'm not sure if you query with `.toLowerCase`

app/services/editions/db/PackageQueries.scala
```diff
@@ -0,0 +1,430 @@
...
+ 			"""
+    sql
+      .map(rs => {
+        val metadata = rs.stringOpt("metadata").map(Json.parse)
+        PackageCardRow(
+          packageId = rs.string("package_id"),
+          cardType = PackageCardType
+            .fromString(rs.string("card_type"))
+            .getOrElse(
+              throw new IllegalArgumentException(
```
davidfurey, line 388: Should this throw if just one row in the response is corrupt?

app/services/editions/db/PackageQueries.scala
```diff
@@ -0,0 +1,430 @@
...
+    override def toString = "updated_on"
+    override def toSql = sqls"updated_on"
+  }
+  case object Title extends OrderingField {
+    override def toString = "name"
+    override def toSql = sqls"name"
+  }
+
+  object OrderingField {
+    def fromString(str: String): Option[OrderingField] = str match {
```
davidfurey, line 423: condOpt?

davidfurey (commented):
I haven't reviewed the tests.

conf/evolutions/default/14.sql
fredex42: packages has dependents in package_cards, but package_cards does not have any dependents

app/model/packages/client/CreatePackageRequest.scala
fredex42: all millis - will add a note

app/model/packages/PackageCardType.scala
fredex42: it does work, true, but I prefer to be more explicit - it's easier to read for anyone who is less experienced in Scala

app/services/editions/db/PackageQueries.scala
fredex42: no it isn't - they should always be distinct AFAIK

app/services/editions/db/PackageQueries.scala
fredex42: good spot thanks

app/services/editions/db/PackageQueries.scala
fredex42: again i prefer explicit, it's not more typing and it's a little clearer to the non-initiated

conf/evolutions/default/14.sql
fredex42: I _would_ agree; but the convention elsewhere is for `is_hidden` so i followed that

app/model/packages/PackageCard.scala
fredex42: you're right; I tend to just assume that any input data could be incorrect and handle it as such

app/model/packages/PackageCard.scala
fredex42: PackageCardRow is a class which represents the database data directly; PackageCard is a sealed trait which presents the data ergonomically depending on the exact type of object returned.  It felt easier to handle the data that way but I could combine them if you want

app/model/packages/PackageCardRow.scala
fredex42: I don't believe it can in this case, since `metadata` is typed as a JSONB so the database itself should ensure that the string coming back in `rs` is valid JSON

app/services/editions/db/PackageQueries.scala
fredex42: it's now properly typed to UUID, so not an issue

app/services/editions/db/PackageQueries.scala
fredex42: I _did_ cover it in the docstring 😇  Changed the name to `thisDayOnly` ?

app/model/packages/Package.scala
fredex42: Hmmmmm, I do see your point; but that would be _even more_ boilerplate.  It seems more efficient to only branch the one bit that is different - platform-specific metadata - and keep the rest the same.  There should be no issue with knowing the right kind of metadata, since the code just switches on which implementation of PackageMetadata is present, transparently in the formatter.

app/model/packages/PackageMetadata.scala
fredex42: As below - I see your point but that would require even _more_ boilerplate, and there is quite a bit of that already

app/model/packages/PackageCard.scala
fredex42: PackageCard should not have had a `fromRowOpt` method for this reason, and I have therefore removed it

conf/evolutions/default/14.sql
fredex42: technically - yes, you're right. Practically it shouldn't make a difference, but I'll put it in to be accurate

conf/evolutions/default/14.sql
davidfurey: Fair enough

conf/evolutions/default/14.sql
davidfurey: But you deleted package_cards before packages? In any case, have you tested the "Downs"?

app/services/editions/db/PackageQueries.scala
davidfurey: Clearer to the non-initiated, but `Some` noise for the initiated. Fine.

app/model/packages/PackageMetadata.scala
```diff
@@ -0,0 +1,75 @@
...
+              JsError(s"Unexpected field(s): ${extraKeys.mkString(", ")}")
+            } else {
+              Json.reads[FeastPackageMetadata].reads(obj)
+            }
+          case _ => JsError("Expected a JSON object")
+        }
+    }
+}
+
+case class WebPackageMetadata(
```
davidfurey: I'm not sure web is the correct name here, since this covers web and apps.

app/model/packages/Package.scala
davidfurey: Transparently but not necessarily correctly. At the moment an empty object deserialises as `FeastPackageMetadata` when it if packageType is Web, it should deserialise as `WebPackageMetadata`

app/model/packages/PackageMetadata.scala
```diff
@@ -0,0 +1,75 @@
...
+        Set("theme", "bodyText", "targetedRegions", "excludedRegions")
+      override def writes(o: FeastPackageMetadata): JsObject =
+        Json.writes[FeastPackageMetadata].writes(o)
+
+      override def reads(json: JsValue): JsResult[FeastPackageMetadata] =
+        json match {
+          case obj: JsObject =>
+            val extraKeys = obj.keys.diff(allowedKeys)
+            if (extraKeys.nonEmpty) {
+              JsError(s"Unexpected field(s): ${extraKeys.mkString(", ")}")
```
davidfurey: Is this desirable? It makes schema change rollouts more difficult to manage

conf/evolutions/default/14.sql
fredex42: Yes, packages depends on package_cards but package_cards does not depend on anything (so cascade makes no difference on it).  I have tested the downs locally and they have been running ok

fredex42:
> At the moment an empty object deserialises as FeastPackageMetadata when it if packageType is Web, it should deserialise as WebPackageMetadata

@davidfurey hmmm you're right that's an issue

conf/evolutions/default/14.sql
davidfurey: I remain confused as to how packages can depend on something that you've deleted, but am reassured that it runs locally.

app/model/packages/MetadataHelpers.scala
```diff
@@ -0,0 +1,49 @@
...
+  private def formatJsResult[T](data: JsResult[T]): Either[String, T] = {
+    data.asEither.left.map(errorList => {
+      val errors = errorList.map({ case (path, errors) =>
+        s"$path has ${errors.length} errors: ${errors.map(_.toString).mkString(",")}"
+      })
+      errors.mkString("\n")
+    })
+  }
+
+//  protected def getPackageMetadata(
```
davidfurey: This can be deleted

davidfurey (approved):

